PR #24, opened 2015-09-24
Fix/bids1.0.0

chrisgorgo:
This PR makes the 1.0.0-rc1 examples validate.

### DIFF
--- a/validators/bids.js
+++ b/validators/bids.js
@@ -52,9 +52,9 @@ var BIDS = {
 
                 if (
                     path[0].indexOf('.nii.gz') > -1 &&
-                    (path[1].indexOf('anatomy') > -1 ||
-                     path[1].indexOf('functional') > -1 ||
-                     path[1].indexOf('diffusion') > -1) &&
+                    (path[1].indexOf('anat') > -1 ||
+                     path[1].indexOf('func') > -1 ||
+                     path[1].indexOf('dwi') > -1) &&
                     (
                         (path[2] && path[2].indexOf('ses') > -1) ||
                         (path[2] && path[2].indexOf('sub') > -1)
@@ -135,7 +135,7 @@ var BIDS = {
                         reason: 'NifTi files should be compressed using gzip.',
                         severity: 'error'
                     }
-                    
+
                     self.errors.push({file: file, errors: [newError]});
                 }
                 cb();
@@ -172,7 +172,7 @@ var BIDS = {
             } else {
                 cb();
             }
-        
+
         }, function () {
             callback(self.errors, self.warnings);
         });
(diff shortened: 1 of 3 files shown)


validators/bids.js
```diff
@@ -121,61 +125,50 @@ var BIDS = {
         var self = this;
 
         // validate individual files
-        async.forEachOf(fileList, function (file, key, cb) {
```
chrisgorgo, line 124: I started replacing async with classic for loops for speed and clarity, but then I realised that this might mess up the way it behaves in the browser. Did you opt for using async loops to make them non blocking?

constellates, line 124: I used async so we can use asynchronous operations inside the loop without the potential for race condition issues. Generally anything remotely cpu intensive should be performed asynchronously in node and the browser so you don't block the event loop in node or lock up the UI in the browser. The contents of the loop are already asynchronous, but if we weren't using async loops the loop would end as soon as all the asynchronous operations started but potentially before those operations end. 

In this particular loop the places where we'll most likely see issues is where files are being read and where jshint is running. So the final callback could be called missing errors and warning that were found but didn't call back quick enough. This issue is exacerbated by the browser but could definitely happen in node as well.

chrisgorgo, line 124: Thanks for the clarification! I understand the general principle, but I
struggle to see which functions inside this loop are asynchronous. Could
you point them out? Sorry for the silly question.

constellates, line 124: Basically any function that takes a callback function as an argument is asynchronous. So in the TSV and JSON sections you'll see the function signatures

```
utils.readFile(file, function (contents) {})

TSV(contents, function (errs, warns) {})

JSON(contents, isSidecar, function (errs) {})
```

So far it looks like these are all the async functions being called from this loop.

chrisgorgo, line 124: Correct me if I'm wrong, but is TSV() truly asynchronous? It takes a callback as an argument, but it does not run any IO operations nor does it use process.nextTick() or setTimeout().

constellates, line 124: Yes you're right. There's nothing truly async inside of the TSV check. You should be safe to use standard loops in there.

chrisgorgo, line 124: Is there a reason to return variables through a callback function passed as an argument in non async functions?

constellates, line 124: There's no reason the TSV check must return results through a callback. Although the benefits I see in doing so are;  we have structure in place that will support asynchronous operations if in the future we need to use them within the TSV check and we maintain a reasonably similar structure across validators. 

The latter provides arguably better code readability by providing a common structure for similar operations. Common structures are also easier to expand upon as the project grows because we can more easily spot when we're unnecessarily repeating ourselves, write helper functions that can easily work with all of the validators, and use the validators dynamically.

And lastly it can also make things easier for developers using the validator by providing a common API. Our main use case is to use the validator on a whole BIDS dataset but we also expose the ability to use the validators independently. For example...

```
var validate = require('bids-validator');

// you can use
validate.TSV();
validate.JSON();
```

All that being said the validator is still small enough at this point that we can change conventions if there's a need to do so.

chrisgorgo:
I'm closing it, because I started mixing bug fixes with refactoring too much.

PR #28, opened 2015-10-12
fix CLI not reporting a single error

chrisgorgo:
I fixed a bug with JSOn parsing and did some code cleanup.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -7,7 +7,7 @@ module.exports = function (dir) {
 	    validate.BIDS(dir, function (errors, warnings) {
 	    	if (errors === 'Invalid') {
 	    		console.log(colors.red("This does not appear to be a BIDS dataset. For more info go to http://bids.neuroimaging.io/"));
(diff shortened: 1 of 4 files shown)


constellates:
Thanks Chris. This looks great. I'm going to go ahead and merge this but I did notice one issue while testing that we might want to address in the future. A non gzipped nifti file will throw an error and a warning that may be confusing when shown together. Let me know what you think.

```
/sub-01/anat/sub-01_inplaneT2.nii
NifTi files should be compressed using gzip.
Evidence: sub-01_inplaneT2.nii
Severity: error

/sub-01/anat/sub-01_inplaneT2.nii
This file is not part of the BIDS specification, make sure it isn't included in the dataset by accident. Data derivatives (processed data) should be placed in /derivatives folder.
Evidence: sub-01_inplaneT2.nii
Severity: warning
```

chrisgorgo:
Good point - I'll send another PR fixing this.

PR #29, opened 2015-10-12
Browser fix

constellates:
Did some minor cleanup and fixed issues that prevented the validator from working w/ the browser application.

### DIFF
--- a/validators/bids.js
+++ b/validators/bids.js
@@ -52,25 +52,27 @@ var BIDS = {
     quickTest: function (fileList, callback) {
         var couldBeBIDS = false;
         for (var key in fileList) {
-            var file = fileList[key];
-            var path = utils.relativePath(file);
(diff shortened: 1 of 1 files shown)


chrisgorgo:
Looks good to me. Is there some easy way to add browser functionality to the tests?

constellates:
I'm not sure the best way to automate testing the browser functionality. The main challenge is testing things that relate to browser file objects. We can mock the objects but the FileReader won't actually be able to read them. I'll have to keep thinking on that. 

There are headless browsers like phantom.js for running tests in that could be run with CI but we may have trouble finding one that supports mock/automated folder selects.

chrisgorgo:
This might be worth looking into:
https://circleci.com/integrations/saucelabs I think it's free for open
source projects.

PR #30, opened 2015-10-14
Improved metadata validation

chrisgorgo:
The major improvement in this PR is checking if each BOLD nifti file has required and recommended metadata defined.

### DIFF
--- a/bin/bids-validator
+++ b/bin/bids-validator
@@ -3,12 +3,14 @@
 process.title = 'bids-validator';
 
 var argv = require('yargs')
-    .usage('Usage: $0 <dataset_directory>')
+    .usage('Usage: $0 <dataset_directory> [options]')
     .demand(1, 1)
+    .boolean('ignoreWarnings')
+    .describe('ignoreWarnings', 'disregard non-critical issues')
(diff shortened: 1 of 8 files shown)


chrisgorgo:
One thing I would like someone to double check is my use of lists and dictionaries inside async loops. On a few occasions I update externally declared objects inside those loops - it seems to work, but I'm not sure if it works only accidentally (is .push() atomic?).

constellates:
There's no reason you can't update an externally declared object inside a loop so long as it's accessible from that loops scope. And for grabbing items from inside loops you pretty much have to externally declare objects like you're doing. Javascript is also generally single-threaded so you shouldn't need to worry about true atomicity in the same way you might it other ecosystems. The main thing to worry about with asynchronicity is whether or not events are happening in the order they need to. The "async" library has a lot of tools for ensuring that. 

Everything looks pretty good to me and I just tested it in the browser. The only thing I notice is I'm getting a lot more warnings now (but I think that's expected). I'm going to go ahead and merge.

There are a couple things I'd like to do to clean up the code organization. I'll try to get an update and a pull request up pretty soon for you to take a look.

Issue #34, opened 2015-10-16
Check if repetition time defined in the nifti header is the same as the one in corresponding JSON file.


chrisgorgo:

constellates:
I'm testing out reading the headers with this library https://github.com/scijs/nifti-js . So far I've only tested two files (ds001/sub-01/anat/sub-01_inplaneT2.nii.gz and ds001/sub-01/anat/sub-01_T1w.nii.gz). Here's an example of what that library is extracting. I don't see repetition time? Do you know if those particular files don't contain it or if that's a limitation of the library?

```
{
    "dimension": 3,
    "type": "int16",
    "encoding": "raw",
    "endian": "little",
    "sizes": [
        128,
        128,
        33
    ],
    "spaceUnits": [
        "mm",
        "mm",
        "mm"
    ],
    "spacings": [
        1.5625,
        1.5625,
        4
    ],
    "spaceDimension": 3,
    "buffer": {},
    "data": {...}
}
```

It also appears that this library doesn't support partial file reads (i.e. reading and parsing the first 1000 bytes). So I haven't been able to test any of the larger functional scans due to memory limitations.

Do you know of any other javascript libraries that parse nifti binaries? And if not do you know where to find the specification for how the binary data is encoded? I found a link to this NIH side that appears to be down. http://nifti.nimh.nih.gov/

constellates:
Adding this for reference. It looks to be a good resource for understanding the nifti format http://brainder.org/2012/09/23/the-nifti-file-format/

chrisgorgo:
Yes - Anderson did a great job writing this up. To answer your questions:
- RepetitionTime only makes sense for _dwi and _bold files (they are 4 dimensional)
- I don't know of any other npm installable nifti reading library. I agree we should try to read only the header. If not to complicated we should enhance the nifti-js library by this feature or publish our own nifti reading library. There are some other JS based project that have nifti reading support built in that we can use for inspiration: https://github.com/rii-mango/Papaya and https://github.com/xtk/X

constellates:
I've had some luck this morning getting nifti-js to only read through the header values. However, now I might need some help understanding how to get the repetition time from the header. Anderson lists out all of the header values and their location/size on that write up. Repetition time isn't mentioned in the header list. Is this a value that I need to interpret from other values or does it actually exist in the header somewhere I'm not seeing?

constellates:
[...]

PR #35, opened 2015-10-26
Nifti header

constellates:
Added optional functionality for parsing Nifti header data and optionally validating against it by comparing related json metatdata. This should address all issues in #34. 

Note: Whenever the nifti-js library gets updated on npm we should update our package.json to point there instead of the nifti-js github repo.

@chrisfilo Take a look and let me know if you have any question/concerns before I merge.


chrisgorgo:
Excellent work. One suggestion I would make is to add as small nifti file and add a header reading test on it. There are some small files in this dataset: https://googledrive.com/host/0B2JWN60ZLkgkMEw4bW5VUUpSdFU/ds003_downsampled.tar (which we use for testing other software).

Issue #33, opened 2015-10-16
Check if there is a corresponding _events.tsv file for each _bold.nii.gz file


chrisgorgo:
One needs to account for hierarchies (done events file for the whole dataset, one events file for each subject, one events file for each session, one events file for each run). If file does not exist a warning should be thrown - unless word "rest" (any caps) is present in the task name.

constellates:
With the exception of the "rest" keyword does the hierarchy logic work the same with these events files as it does with the JSON sidecars. I'm planning on reusing the logic that determines if those exist. And can you give me an example of how the word "rest" might appear in a taskname.

chrisgorgo:
1) yes the hierarchy logic is the same as for json sidecars files
2) you should look for "rest" in the label part of "_task-<label>_"

Let me know if this helps.

PR #36, opened 2015-10-26
Events

constellates:
Added warnings for missing event files. Should take care of #33.

### DIFF
--- a/tests/nii.spec.js
+++ b/tests/nii.spec.js
@@ -3,14 +3,39 @@ var validate = require('../index');
 
 describe('NIFTI', function(){
 
-	var path = '';
+	var path = '/sub-15/func/sub-15_task-mixedeventrelatedprobe_run-01_bold.nii.gz';
 	var header = {error: 'Unable to read ' + path};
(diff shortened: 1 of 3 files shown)


tests/nii.spec.js
```diff
@@ -3,15 +3,41 @@ var validate = require('../index');
...
 	});
 
+	it('should warn user about misisng events file', function() {
+		validate.NIFTI(header, path, jsonContentsDict, events, function (errors, warnings) {
+			assert(warnings.length = 1);
+		});
+	});
+
+	it('should ignore missing events files for rest scans', function() {
+		jsonContentsDict['/sub-15/func/sub-15_task-mixedeventrelatedprobe-rest_run-01_bold.json'] = jsonContentsDict['/sub-15/func/sub-15_task-mixedeventrelatedprobe_run-01_bold.json'];
```
chrisgorgo: this is not a valid BIDS path - task names cannot include hyphens

chrisgorgo:
otherwise looks good

Issue #52, opened 2015-11-20
Error reading some .nii.gz files


chrisgorgo:
I'm getting errors reading this file https://drive.google.com/folderview?id=0B2JWN60ZLkgkcndhTWpLYmswRDA&usp=sharing&tid=0B2JWN60ZLkgkN0dMTVQ1QU1IUEk (actually all nifti files from this dataset - ds109):

On commandline:

```
orange:bids-validator filo$ node nameOfFile.js /Volumes/Samsung_T1/bids_examples/symlinked/ds109/sub-01/anat/sub-01_T1w.nii.gz
{ error: 'Unable to read /Volumes/Samsung_T1/bids_examples/symlinked/ds109/sub-01/anat/sub-01_T1w.nii.gz' }
```

With the online version when I try to validate full version of ds109 I get an JS error:

```
r   @   app.min.js:24606
i.onloadend @   app.min.js:30419
```

(which is also concerning - why would I get a different error on command line and online?)

BTW the file is not corrupted:

```
orange:bids-validator filo$ fslinfo /Volumes/Samsung_T1/bids_examples/symlinked/ds109/sub-01/anat/sub-01_T1w.nii.gz
data_type      INT16
dim1           144
dim2           192
dim3           192
dim4           1
datatype       4
pixdim1        1.200000
pixdim2        1.197917
pixdim3        1.197917
pixdim4        2.200000
cal_max        0.0000
cal_min        0.0000
file_type      NIFTI-1+
```

constellates:
Did you send me the wrong file? If I download the linked file it's a zero byte file?

constellates:
The online error also looks like the error from #50 that was caused by not handling errors while reading empty files. If you're still seeing that online you may want to try clearing your cache. And if that fixes it we probably want to implement cache busting during every change.

chrisgorgo:
Here's a correct link
https://docs.google.com/uc?id=0B77zr9yIiKOTQ1c0X1RuY0h3TTg&export=download
this file fails as well
https://docs.google.com/uc?id=0B77zr9yIiKOTQVZ0WW5oZDFVSEk&export=download

constellates:
Thanks. It looks like some files are not making it through gzip decompression. I'm looking into it.

constellates:
[...]

PR #53, opened 2015-11-23
Unzip fail

constellates:
Potential fix for #52.


constellates:
@chrisfilo  This is working for me now in node v5.0.0 and greater. It does not yet support the intentionally unzipped .nii files. I'll tackle that in a separate pull request.

chrisgorgo:
Works with 5.1 as well. Great job!

PR #58, opened 2015-12-02
Removed and ignored test data that is downloaded by the testing scripts

constellates:
Small fix to prevent dynamically downloaded data from getting saved in git.


chrisgorgo:
my bad!

PR #64, opened 2015-12-23
Updated to BIDS 1.0.0-rc3

chrisgorgo:


constellates:
Looks good to me except I think you used the wrong issue code on the slice encoding check. They're both 34 currently. I'm responding from a phone so I'm not able to test it, but everything looks fine to me if you want to merge after checking the issue code.

PR #77, opened 2016-01-12
check validity of intendedFor fields

### LINKED ISSUE #73: Check if "IntendedFor" field matches an existing file

chrisgorgo:
Closes #73

### DIFF
--- a/validators/nii.js
+++ b/validators/nii.js
@@ -10,7 +10,7 @@ var Issue = utils.Issue;
  * it finds while validating against the BIDS
  * specification.
  */
-module.exports = function NIFTI (header, file, jsonContentsDict, bContentsDict, events, callback) {
+module.exports = function NIFTI (header, file, jsonContentsDict, bContentsDict, fileList, events, callback) {
     var path = file.relativePath;
     var issues = [];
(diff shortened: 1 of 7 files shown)


validators/nii.js
```diff
@@ -200,6 +200,30 @@ module.exports = function NIFTI (header, file, jsonContentsDict, bContentsDict,
         }
     }
 
+    if (path.includes("_phasediff.nii") || path.includes("_phase1.nii") ||
+        path.includes("_phase2.nii") || path.includes("_fieldmap.nii") || path.includes("_epi.nii")){
+        if (mergedDictionary.hasOwnProperty('IntendedFor')) {
+            var intendedForFile = "/" + path.split("/")[1] + "/" + mergedDictionary['IntendedFor'];
```
constellates, line 206: I'm guessing this is correct but want to double check. Is it guaranteed that the intended for file is always located at path.split("/")[1]? Could it be nested deeper?

constellates, line 206: Other than that question. This looks good.

chrisgorgo, line 206: Yes this is how it's said in the spec.

PR #80, opened 2016-01-14
Remove duplicate words

mih:


constellates:
Closing because this is the same as #81. Let me know if this is incorrect.

mih:
Not the same, was a different file. Same strings though.

constellates:
Ok. That makes sense. Thanks again.

PR #81, opened 2016-01-14
Remove more duplicate words

mih:
Sorry, for two PRs -- used online editor.


constellates:
Good catch. Thanks for the contribution.

Issue #56, opened 2015-11-29
Report warnings for inconsistent or missing sessions


chrisgorgo:
A common problem with experimental data are missing scans for one or two subjects. We should give warning in such situations. I propose the following procedure:
- For session label session_id
  - For every subject calculate a set of file names for session session_id, remove subject identifiers from those file names.
  - Calculate a union of those sets across all subjects.
  - Report those files from the union that are not preset for all subjects. List which subjects are missing particular file in the warning message

PR #97, opened 2016-02-11
Add new validator for missing session files

rwblair:
In reference to #56


tests/bids.spec.js
```diff
@@ -37,7 +38,17 @@ var suite = describe('BIDS example datasets ', function() {
...
                 validate.BIDS("tests/data/BIDS-examples-" + test_version + "/" + path + "/", options, function (errors, warnings) {
                     assert.deepEqual(errors, []);
-                    //assert.deepEqual(warnings, []);
+                    if (missing_session_files.indexOf(path) === -1) {
+                        //pass
+                    } else {
+                        session_flag = false;
+                        for (var warning in warnings) {
+                            if (warnings[warning]['code'] === '39') {
+                                session_flag = true;
```
chrisgorgo: and break?

utils/issues.js
```diff
@@ -149,6 +149,14 @@ module.exports = {
 	37: {
 		severity: 'error',
 		reason: "'IntendedFor' field needs to point to an existing file."
-	}
+	},
+    38: {
+        severity: 'error',
+        reason: "Directory structure has depth greater than 5."
+    },
```
chrisgorgo: You never use this error code...

utils/issues.js
```diff
@@ -149,6 +149,14 @@ module.exports = {
...
+    39: {
+        severity: 'warning',
+        reason: "Subject session may be missing file"
```
chrisgorgo: either "missing a file" or "missing some files"

tests/bids.spec.js
```diff
@@ -37,7 +38,17 @@ var suite = describe('BIDS example datasets ', function() {
 		    	var options = {ignoreNiftiHeaders: true};
                 validate.BIDS("tests/data/BIDS-examples-" + test_version + "/" + path + "/", options, function (errors, warnings) {
                     assert.deepEqual(errors, []);
-                    //assert.deepEqual(warnings, []);
+                    if (missing_session_files.indexOf(path) === -1) {
```
chrisgorgo, line 48: why not one if statement with "!= -1" and no "else"?

utils/issues.js
rwblair: Had a warning for the directory depth quick test in the works but then the quicktest check for it was removed. l'll remove it.

tests/bids.spec.js
chrisgorgo, line 48: Actually the tests should also fail if code 39 is thrown in a dataset that should have all files for all subjects/sessions.

constellates:
I wonder if we should limit this check to the handful of file extensions that could be session files. I'm getting false positives triggered by files that are already caught by the check to see if a given file is part of the spec.

```
1: This file is not part of the BIDS specification, make sure it isn't included in the dataset by accident. Data derivatives (processed data) should be placed in /derivatives folder. (code: 1)
    /.DS_Store
        Evidence: .DS_Store
    /sub-01/anat/.DS_Store
        Evidence: .DS_Store
    /sub-01/func/.DS_Store
        Evidence: .DS_Store
    /sub-02/.DS_Store
        Evidence: .DS_Store

2: Subject session may be missing a file. (code: 38)
    /sub-16/.DS_Store
        Evidence: Subject: sub-01; Missing file: /.DS_Store
    /sub-16/.DS_Store
        Evidence: Subject: sub-02; Missing file: /anat/.DS_Store
    /sub-16/.DS_Store
        Evidence: Subject: sub-02; Missing file: /func/.DS_Store
    /sub-16/.DS_Store
        Evidence: Subject: sub-03; Missing file: /anat/.DS_Store
    /sub-16/.DS_Store
        Evidence: Subject: sub-03; Missing file: /func/.DS_Store
    /sub-16/.DS_Store
        Evidence: Subject: sub-03; Missing file: /.DS_Store
    /sub-16/.DS_Store
        Evidence: Subject: sub-04; Missing file: /anat/.DS_Store
    /sub-16/.DS_Store
        Evidence: Subject: sub-04; Missing file: /func/.DS_Store
    /sub-16/.DS_Store
        Evidence: Subject: sub-04; Missing file: /.DS_Store
    /sub-16/.DS_Store
        Evidence: Subject: sub-05; Missing file: /anat/.DS_Store
    ... and 35 more files having this issue (Use --verbose to see them all).
```

I also think we might want to clarify the error message to explain that this was thrown because "not all sessions contain corresponding files. Each session should contain the same number of files with the same naming scheme." Or something to that effect. I don't have a great understanding of this particular spec requirement so I'm not sure where the emphasis should be placed.

chrisgorgo:
Good catch! I think we should just exclude files that are not BIDS compatible from this test. This should be easy to achieve if you use this function: https://github.com/INCF/bids-validator/blob/master/utils/type.js#L16

rwblair:
isBIDS function did the trick, fixed the non standard files being reported and updated session warning to:

"Not all sessions contain the same corresponding files. Each session should contain the same number of files with the same naming scheme."

constellates:
Looks good to me. @chrisfilo do you have any more questions, changes or concerns before I merge this in?

PR #99, opened 2016-02-12
updated node version on CI

constellates:
Updated node in CI config to latest stable so newer gulp sass can be used on gh-pages.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,6 +1,6 @@
 machine:
   node:
-    version: 5.6.0
+    version: 4.3.0
 
 deployment:
   production:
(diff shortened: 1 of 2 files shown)


chrisgorgo:
What are the pros and cons of going to 5.6 instead of 4.3 LTS?

constellates:
I guess it depends what we want to target support for, but LTS is probably what we should use. It's going to be slightly more stable, but more importantly more deployable based on various people's and organizations' policies.

The benefit of going newer is mainly that we've run into issues in the past with compatibility with v5 and newer and this would test against those.

chrisgorgo:
Let's use 4.3 on circleCI. I got 5.6 on my machine and use it for testing. Unfortunately CircleCI does not support testing agains multiple node.js versions.

constellates:
Sounds good.

Issue #92, opened 2016-02-10
Hangs with dangling symlink


mih:
If the validator hits a symlink that points straight to Valhalla it hangs forever and eventually exists with an exception -- without producing any other output. I guess it should simply report a dangling symlink as such. Those can be present of one uses git-annex for data management.

rwblair:
Error I'm seeing with a dangling symlink:

fs.js:892
  return binding.stat(pathModule._makeLong(path));
                 ^

Error: ENOENT: no such file or directory, stat 'tests/data/BIDS-examples-1.0.0-rc3u4/ds001//sub-01/anat/dangling.sym'
    at Error (native)
    at Object.fs.statSync (fs.js:892:18)
    at getFiles (/Users/rblair/projects/bids-validator/utils/files.js:95:16)
    at getFiles (/Users/rblair/projects/bids-validator/utils/files.js:96:13)
    at getFiles (/Users/rblair/projects/bids-validator/utils/files.js:96:13)
    at Object.readDir (/Users/rblair/projects/bids-validator/utils/files.js:73:17)
    at Object.BIDS.start [as BIDS](/Users/rblair/projects/bids-validator/validators/bids.js:30:21)
    at module.exports (/Users/rblair/projects/bids-validator/cli.js:8:15)
    at Object.<anonymous> (/Users/rblair/projects/bids-validator/bin/bids-validator:20:18)
    at Module._compile (module.js:413:34)

Different function may need to be used to tell if a file is a symlink, fs.stat' fs.Stats object wont contain an isSymbolicLink attribute:
https://nodejs.org/dist/latest-v4.x/docs/api/fs.html#fs_class_fs_stats
https://nodejs.org/dist/latest-v4.x/docs/api/fs.html#fs_fs_lstat_path_callback

constellates:
Thanks @rwblair . That looks like the root of this issue. I have some time to take a look at this now, so let me know if you've already started writing a fix so we're not doubling up.

rwblair:
Have not started on it yet so you should be good.

constellates:
Added new PR #156

PR #100, opened 2016-02-16
Symlinks

constellates:
Fix for #92. I added better error handling for orphaned symlinks and otherwise unreadable files. Node has a variety of tools for evaluating files and symlinks, but I have not found the same for the browser. So node will actually check if a file is a symlink and inaccessible. The browser just checks if a file is readable and returns a more generic issue.

Also fixed a couple missing variable declarations that were causing errors in the browser.


chrisgorgo:
LGTM - @hanke could you test it?

constellates:
I merged in the most recent master. I'll comment again when I'm done re-testing the fix.

constellates:
This has been tested and is still working. I also fixed a couple missing variable declarations that got added in when I merged the recent master. Node is lenient enough to assume we want them defined but they break the browser.

constellates:
Up to date with current master.

utils/files.js
```diff
@@ -279,4 +294,28 @@ function relativePath (file) {
...
 
+/**
+ * Test File
+ *
+ * Takes a file and callback and tests if it's viable for
+ * reading. Callsback with an error and stats if it isn't
+ * or null and stats if it is.
+ */
+function testFile (file, callback) {
+    fs.lstat(file.path, function (statErr, stats) {
```
chrisgorgo, line 305: do those calls work in the browser?

utils/files.js
```diff
@@ -279,4 +294,28 @@ function relativePath (file) {
...
+        fs.access(file.path, function (accessErr) {
+            if (!accessErr) {
+                callback(null, stats);
+                return;
+            } else {
+                if (stats.isSymbolicLink()) {
+                    callback(new Issue({code: 43, file: file}), stats);
```
chrisgorgo, line 312: This discards all symbolic links - those that point to real files as well as dangling ones.

chrisgorgo:
I jumped the gun and merged it just to discover it has a critical bug: it raises and error for any symbolic links not only dangling ones. I had to revert the merge here: https://github.com/INCF/bids-validator/commit/58156e52d2a1d1c20caae9294a55fd7070b1cfa7. Sorry for the mess.

utils/files.js
constellates, line 305: Those calls don't work in the browser, but that function is only ever called in node. We first check if `fs` exists. I'll update this to fix the issue you found below and incorporate master and resubmit.

PR #102, opened 2016-03-14
Test for consistent dimensionality and resolution in nifti file headers

rwblair:
Here is the output from 7t_trt doing these checks:
https://gist.github.com/rwblair/74605fc0965f1a442d16

In my branch headerCompare I added a function to not raise a warning if an header field for a failed differs from the most common header values by less than 1

Here is the more concise output with it:
https://gist.github.com/rwblair/1e242fff10e4dfbdad64

If this is preferable I can open a PR on that branch.


chrisgorgo:
There is something off with those reports. I get different results when using nibabel:
selection_011

You should also use the first value of dim and pixdim to figure out how many diemnsions arethere (see http://brainder.org/2012/09/23/the-nifti-file-format/ for more info).

We should also ignore differences below 0.00001

rwblair:
Updated to pull the length of the dim array from its first element, and ignore differences of less than .00001. Output of a run on 7t_trt:

https://gist.github.com/rwblair/6a3e9133b054142e9a00

chrisgorgo:
Thanks for the update:
- I don't think pixdim[0] is useful to report here, you should report pixdim[4] when appropriate, but take into account the units stored in the xyzt_units
- different runs should have the same scanning parameters - please combine them when calculating the most common parameter set (but split across sessions and `acq`)
- as I showed above: /sub-19/ses-1/func/sub-19_ses-1_task-rest_acq-fullbrain_run-1_bold.nii.gz has pixdim 1.5x1.5x1.5 which is consistent with the mode. Your logs show it's 3x3x3 so there is something off... Maybe you can validate your files locally with `nib-ls`?

rwblair:
Wrapped pixdim values up with their units for comparison. Evidence messages are now more specific depending on the field being checked:

screen shot 2016-03-22 at 4 13 56 pm

utils/issues.js
```diff
@@ -154,6 +154,10 @@ module.exports = {
         severity: 'warning',
         reason: "Not all sessions contain the same corresponding files. Each session should contain the same number of files with the same naming scheme."
 
-    } 
+    }, 
+    39: {
+        severity: 'warning',
+        reason: "Not all subjects/sessions/runs have the same scanning parameter"
```
chrisgorgo: parameters

validators/headerFields.js
```diff
@@ -0,0 +1,165 @@
...
+        } else if (field === 'pixdim') {
+            field_value = [];
+            var pix_dim = header[field].slice(0,4);
+            var units = header['xyzt_units'].slice(0,4);
+            for (var i = 0; i < 4; i++) {
+                field_value.push('' + pix_dim[i] + units[i]); 
+            }
+            field_value = field_value.toString();
+        } else {
+            console.log("Checks against header field: " + field + " are currently unsupported.")
```
constellates, line 54: Do we need this log here or was this leftover from testing?

rwblair, line 54: It should never be hit, was thinking of it as a courtesy if anyone tried to call the function with something other than dim or pixdim. If called with something else it will return with no issues, field_value will be undefined which can be used as a key for an object:

```
> x = []
[]
> x[undefined] = 5
5
```

constellates, line 54: That makes sense. Thanks for the explanation.

chrisgorgo:
Is this ready for review/merge?

rwblair:
Sorry I missed this, all the datasets I've tested on my end have been good. I guess the version number should be bumped for this. I'll wait until #117 gets merged, and then do another commit for the incremented version number.

chrisgorgo:
This should be fixed now - could you rebase/merge master?

Issue #104, opened 2016-03-21
Error validating very large dataset (web)


oesteban:
The bids validator hangs when validating a tree of ~2224 images (1112 subjects):
![bids_validator_error](https://cloud.githubusercontent.com/assets/598470/13930044/738f185e-ef5a-11e5-91c0-5e166c9f1ee8.png)

constellates:
Hey, is it possible for you to share this dataset or even a file tree description? If it's from a set of test data that Chris has shared I may already have a copy. Otherwise I can do some testing by creating some large datasets to try to reproduce this.

Also do you know if this fails with the command line version?

oesteban:
Command line:

```
/home/oesteban/.nvm/v5.9.0/lib/node_modules/bids-validator/validators/nii.js:74
        var repetitionUnit = header.xyzt_units[3];
                                              ^

TypeError: Cannot read property '3' of undefined
    at NIFTI (/home/oesteban/.nvm/v5.9.0/lib/node_modules/bids-validator/validators/nii.js:74:47)
    at /home/oesteban/.nvm/v5.9.0/lib/node_modules/bids-validator/validators/bids.js:178:29
    at Gunzip.<anonymous> (/home/oesteban/.nvm/v5.9.0/lib/node_modules/bids-validator/utils/files.js:126:21)
    at emitOne (events.js:90:13)
    at Gunzip.emit (events.js:182:7)
    at readableAddChunk (_stream_readable.js:153:18)
    at Gunzip.Readable.push (_stream_readable.js:111:10)
    at Gunzip.Transform.push (_stream_transform.js:128:32)
    at Zlib.callback (zlib.js:572:14)
```

oesteban:
For the dataset, I can email a link to download it.

constellates:
Thanks @oesteban the command line error should be enough info to replicate this. It looks like we're not seeing if xyzt units are present in the header before trying to act on them.

oesteban:
Ok, if it helps, it is very common that the field is empty. Maybe issuing a warning is fine.

constellates:
Ok there's a PR open now. It turns out we already had an issue defined for repetition units not being defined is either seconds, milliseconds or microseconds. So cases in which it is undefined will now fall under the same issue which is currently set as an error.

Let me know if you're able to test if the above branch works with your dataset and loop Chris in if you think we should switch this case from an error to a warning.

constellates:
[...]

PR #105, opened 2016-03-21
Undefined Repetition Unit

constellates:
Small change to account for undefined repetition units as per issue #104 

This case will now return issue code 11 "Repetition time was not defined in seconds, milliseconds or microseconds in the scan's header."


chrisgorgo:
+1!

Issue #107, opened 2016-03-22
app.min.js:32362 Uncaught TypeError: Cannot read property 'endsWith' of undefined


chrisgorgo:
Web version of the validator throws:

```
app.min.js:32362 Uncaught TypeError: Cannot read property 'endsWith' of undefined
```

when validating 7t_trt example. 

Command line version gives a proper set of warnings.

constellates:
Fixed and working on gh-pages site.

PR #125, opened 2016-04-05
Issue 107

constellates:
Fixes issue #107.

I'll try to keep an eye out for other merges and update the version on this as necessary.


constellates:
Rebased to current master.

Issue #128, opened 2016-04-06
Header Fields check doesn't account for missing units


constellates:
The headerFields.js validator relies on xyzt_units and breaks the validation if they're not defined.

https://github.com/INCF/bids-validator/blob/master/validators/headerFields.js#L48

Here's how we handle a similar issue in the main nifti validator.

https://github.com/INCF/bids-validator/blob/master/validators/nii.js#L74

We could use a similar ternary, but I don't know enough about the headerFields.js to know what else may be affected by setting units to null or an empty array.

@rwblair can you take a look or give me a breakdown of how this check is using the units. 

You can see the issue by testing ds108.

rwblair:
I'll update headerFields.js to check to make sure the header values are there, and to skip dimensionality and resolution checks depending on what header values aren't present.

PR #131, opened 2016-04-08
Check for empty or blank header fields in headerFields.js for #128

rwblair:
In reference to #128 
Created three new warnings for each of the headers being tested. This causes a bit of overlap for error code 11 and the new warning code 41:

```

    1: Repetition time was not defined in seconds, milliseconds or microseconds in the scan's header. (code: 11)
        /sub-01/func/sub-01_task-Emotionregulation_run-01_bold.nii.gz

    6: Nifti file's header field for unit information for x, y, z, and t dimensions empty or too short (code: 41)
        /sub-01/func/sub-01_task-Emotionregulation_run-01_bold.nii.gz
```

@chrisfilo Is this redundancy in warning messages ok?

### DIFF
--- a/utils/issues.js
+++ b/utils/issues.js
@@ -160,5 +160,17 @@ module.exports = {
         severity: 'warning',
         reason: "Not all subjects/sessions/runs have the same scanning parameters"
     }
+    40: {
+        severity: 'warning',
+        reason: "Nifti file's header field for dimension information blank or too short."
+    },
+    41: {
+        severity: 'warning',
+        reason: "Nifti file's header field for unit information for x, y, z, and t dimensions empty or too short"
+    },
+    42: {
+        severity: 'warning',
+        reason: "Nifti file's header field for pixel dimension information empty or too short."
+    }
 
 };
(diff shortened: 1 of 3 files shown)


constellates:
Thanks @rwblair this fixes the x,y,z,t units issue. I'm now seeing a couple small issues that are preventing this from working in the browser. There's a couple variable declaration issues that node will ignore, but the browser won't handle.

This line should end in a comma so headers is defined.
https://github.com/INCF/bids-validator/blob/master/validators/bids.js#L97

And these three variables need the var keyword in order to be defined.
https://github.com/INCF/bids-validator/blob/master/validators/headerFields.js#L102
https://github.com/INCF/bids-validator/blob/master/validators/headerFields.js#L144
https://github.com/INCF/bids-validator/blob/master/validators/headerFields.js#L145

I believe that should do it. I can retest the browser version locally after those changes. I've added a linter in a separate PR that will hopefully make these types of things obvious sooner.

rwblair:
Very silly things for me to have missed...

I'll fix these and test the browser build locally.

rwblair:
Tested with the reduced ds108, 7t_trt, ds172 and ds200 and did not see any javascript console errors.

constellates:
Looks good to me. Thanks.

PR #157, opened 2016-05-10
Issue limiting

constellates:
Moved primary file limiting from CLI into the main `BIDS` validator file. The CLI should work exactly as it did before the changes. After this merge the web demo will only display 10 files per issue and we'll be able to add a message about the remaining files that aren't being displayed including a count.

note: Both currently open PRs have the same version number. I'll bump either as needed.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -8,7 +8,6 @@ var bytes     = require('bytes');
 var fs        = require('fs');
 
 module.exports = function (dir, options) {
-    options.filesPerIssueMax = 10;
     if (fs.existsSync(dir)) {
         validate.BIDS(dir, options, function (errors, warnings, summary) {
             if (errors === 'Invalid') {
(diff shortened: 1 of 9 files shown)


chrisgorgo:
Seems fine, but I have one concern: if a user of the web version would like to see more than 10 files how could they achieve this?

constellates:
So currently they couldn't, but we have a couple options and the web-validator and app don't necessarily have to act identically. Also 10 files is an arbitrary number at this point sow we can adjust as needed or make it configurable.

#### Web App (CRN app)

The original intent of this PR was to give us a strategy to limit how much validation data we're storing alongside datasets so they can load faster. The plan I've had in mind is the following.
1. Limit issues to 10 files per issue, but track the file count beyond that.
2. Display the limited issues and when there are more than 10 issues also display something  like `and 39 other files` at the end of the file list.
3. When a users modifies their dataset to fix some of the issues the validator will be re-run and the list of issues will be updated. So if they fixed 10 files when 40 had the same issue they will now see the next 10 affected files along with the note `and 20 other files`.

#### Web Validator (gh-pages)

For the standalone web validator there is less of an issue with showing all the issues because we're not storing and retrieving them. However, I have seen issues where rendering becomes a problem when there are thousands of reported files. For this we can do one of the following.
1. Set the verbose option to `true` and it will continue to work like it does now.
2. Provide a toggle on the interface to set the verbose option before selecting a dataset.
3. Always run the verbose mode but limit how many files we render in the app and provide a `click to view more files` option.

chrisgorgo:
For the gh-pages I like option 3 the most. As for the app it might be a bit annoying having to iterate over 10 file batches when fixing errors, but we can fix that later.

constellates:
Sounds good. I'll start working on an update to gh-pages for options 3.

PR #161, opened 2016-05-16
Validator gh pages file limit

constellates:
gh-pages changes as discussed in PR #157 

These changes should work the same with or without #157 

Limits issue rendering to 10 files per issue with a "View x more files" link (example below).

screen shot 2016-05-16 at 11 29 00 am


constellates:
This is noted as failing because there are no tests in the gh-pages branch.

Issue #160, opened 2016-05-16
Files listed twice in EchoTime{1,2} error for phasediff images


wtriplett:
See below - each file is listed twice.

```
    2: You have to define 'EchoTime1' and/or 'EchoTime2' for this file. (code: 15)
        /sub-100003/fmap/sub-100003_phasediff.nii.gz
        /sub-100003/fmap/sub-100003_phasediff.nii.gz
        /sub-100042/fmap/sub-100042_phasediff.nii.gz
        /sub-100042/fmap/sub-100042_phasediff.nii.gz
        /sub-100057/fmap/sub-100057_phasediff.nii.gz
        /sub-100057/fmap/sub-100057_phasediff.nii.gz
        /sub-100059/fmap/sub-100059_phasediff.nii.gz
        /sub-100059/fmap/sub-100059_phasediff.nii.gz
        /sub-100063/fmap/sub-100063_phasediff.nii.gz
        /sub-100063/fmap/sub-100063_phasediff.nii.gz
        /sub-100068/fmap/sub-100068_phasediff.nii.gz
        ... and 101 more files having this issue (Use --verbose to see them all).
```

constellates:
Looks like the issue is here.

https://github.com/INCF/bids-validator/blob/master/validators/nii.js#L150-L164

There are two separate checks for "EchoTime1" and "EchoTime2" that are sharing an issue code. So if a file is missing both it gets added twice.

I should be able to create a fix/PR later today.

constellates:
@chrisfilo I'll make sure we're not adding these files twice, but there's also some discrepancy between the error message and what we're actually checking?

The error message says you have to define "EchoTime1 **and/or** EchoTime2", but the current logic works as if you must define "EchoTime1 **and** EchoTime2" for a phasediff.

Which is currently correct, the message or the logic?

chrisgorgo:
The logic is correct - the message needs to be corrected. Thanks for
catching this!

PR #162, opened 2016-05-16
Issue 160

constellates:
Fix for issue #160

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bids-validator",
-  "version": "0.18.4",
+  "version": "0.18.5",
   "description": "",
   "main": "index.js",
   "repository": {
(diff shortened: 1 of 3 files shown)


chrisgorgo:
This will still give an effect of the same file being listed twice in the non-verbose view. Maybe the best option would be to split it into two errors with separate codes? @wtriplett

constellates:
What's a case where this would still list the same file twice within that issue? Previously if both EchoTime1 and EchoTime2 were missing it would check those individually and add the file twice. I updated the check so that if either or both EchoTime1 and EchoTime2 are missing it will add the issue/file.  

I've tested it and switched back and forth to master to verify the issue and it seems to be working as expect, but I've only tested a couple limited cases.

chrisgorgo:
Sorry - you are right! I misread the code!

Issue #163, opened 2016-05-16
nifti_file is used before it's defined in headerChecks


constellates:
A variable named `nifti_file` is first used here https://github.com/INCF/bids-validator/blob/master/validators/headerFields.js#L73, but not defined until here https://github.com/INCF/bids-validator/blob/master/validators/headerFields.js#L158.

If the header 'dim' or 'pixdim' checks are triggered the following error will be thrown. 

```
/Users/zack/Projects/crn/validator/validators/headerFields.js:82
                issues[nifti_file.relativePath] = new utils.Issue({
                                 ^

TypeError: Cannot read property 'relativePath' of undefined
    at headerField (/Users/zack/Projects/crn/validator/validators/headerFields.js:82:34)
    at headerFields (/Users/zack/Projects/crn/validator/validators/headerFields.js:21:22)
    at /Users/zack/Projects/crn/validator/validators/bids.js:266:50
    at /Users/zack/Projects/crn/validator/node_modules/async/lib/async.js:52:16
    at done (/Users/zack/Projects/crn/validator/node_modules/async/lib/async.js:246:17)
    at /Users/zack/Projects/crn/validator/node_modules/async/lib/async.js:44:16
    at /Users/zack/Projects/crn/validator/validators/bids.js:259:33
    at NIFTI (/Users/zack/Projects/crn/validator/validators/nii.js:217:5)
    at /Users/zack/Projects/crn/validator/validators/bids.js:257:29
    at Gunzip.<anonymous> (/Users/zack/Projects/crn/validator/utils/files.js:135:21)
```

constellates:
I'm not super familiar with `headerFields.js` but it looks like a variable name may have changed during an update and a few instances of the variable were missed. I just pushed a branch to correct what I believe the naming issue is. @rwblair and/or @chrisfilo please take a look and see if there are any issues I may have missed otherwise feel free to merge it in.

I've locally tested the command line and web version with a dataset that trips issue 41 (invalid xyzt units).

Here's the branch.
https://github.com/INCF/bids-validator/tree/issue-163

rwblair:
Changes look good tested them locally without issue. References to nifti_file appears to have been copied directly from the headerField function.

PR #165, opened 2016-05-20
Updated relative path determination based on available data

rwblair:


rwblair:
@chrisfilo want to give you a chance to look at @constellates for #163 before merging.

Issue #166, opened 2016-05-25
Exit code of command-line tool


glatard:
It would be nice if the command-line tool returned a non-zero exit code when the dataset is not valid.

constellates:
This makes sense to me. I added a quick PR to add basic exit codes for errors. @glatard let me know if this looks like it will work for you.

chrisgorgo:
Thanks for reporting Tristan!

glatard:
Looks great, thanks @constellates.

PR #167, opened 2016-05-25
Exit code

constellates:
Added basic exit codes for #166 
- Any valid dataset (with our without warnings) will continue to exit with 0
- Any dataset with errors will exit with 1
- Passing an invalid dataset path will exit with 2

I'm open to changing the specific codes if there's a reason, but the thinking behind these two is. 1 is most often used as a general catch all for errors and 2 is listed (http://www-numi.fnal.gov/offline_software/srt_public_context/WebDocs/Errors/unix_system_errors.html) as "no file or directory".

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -24,7 +24,7 @@ module.exports = function (dir, options) {
         });
     } else {
         console.log(colors.red(dir + " does not exist"));
-        process.exit(1);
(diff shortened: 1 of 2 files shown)


chrisgorgo:
LGTM

Issue #175, opened 2016-07-07
Raise an error if a wrong .bvec/.bval separator is being used


arokem:
This is for hand-crafted single-subject data-set from HCP. I have verified independently that the bvals/bvecs do have the right number of entries, but I might be missing something. How is correspondence (file-to-file) determined? In this case, the bval/bvec files are in the `dwi` folder.

chrisgorgo:
Could you share the bvec/back files? What separator do they use?

arokem:
Here are the bvals/bvecs. Let me know if you need the DWI, and I can share in a less public fashion

https://drive.google.com/open?id=0BxmU4IuCMsYYVjVYNWtkX25SVGc

arokem:
To answer your question: they use two spaces as the separator.

chrisgorgo:
Double spaces are the problem - if you use single spaces (as described in the standard) instead everything should be fine. The error messages should reflect that.

We should test if:
1. multiple spaces are used as delimiters
2. tabs are used as delimiterd

and throw an error that the wrong separator is used.

arokem:
Gotcha. I guess the HCP data that comes down from S3 is just non-compliant.
Oh well.

constellates:
Determining the delimiter can be somewhat tricky, especially when bval files only contain a single row. Would it suffice to assume a single space delimiter and throw an error if any of the values are not numbers. Using this strategy a double spaced file would show a bunch of empty values that would be caught and we could also catch other delimiters like tabs or commas.

Then the error message would then indicate the delimiter and value requirments.

> Error
> .bvec files must be single space delimited and contain only floating point values.
>    /dwi.bvec

PR #177, opened 2016-07-11
Issue 175

constellates:
Addresses #175 
- Added check for inconsistent row lengths
- Added checks for correct value types after splitting on expecting single space separator.
- The latter will also catch invalid separator usage.
- Updated affected tests.


chrisgorgo:
@arokem - could you have a look?

arokem:
LGTM

Issue #82, opened 2016-01-14
Required JSON field 'TaskName' is not tested for presence


mih:
see title

chrisgorgo:
There is one remaining issue after applying #192. Validator complains about lack of TaskName for sbref files where it is not required.

```
    1: You have to define 'TaskName' for this file. (code: 50)
            /sub-100307/func/sub-100307_task-EMOTION_acq-LR_sbref.nii.gz
```

constellates:
Does the above PR cover this or are there any additional scans with "task-" keys that don't require a sidecar "TaskName" definition?

PR #192, opened 2016-09-08
Issue 82

constellates:
- Test for presence of 'TaskName' on task scans for issue #82

### DIFF
--- a/validators/nii.js
+++ b/validators/nii.js
@@ -84,7 +84,7 @@ module.exports = function NIFTI (header, file, jsonContentsDict, bContentsDict, 
                 issues.push(new Issue({
                     file: file,
                     code: 50,
-                    reason: "You have to define 'TotalReadoutTime' for this file. " + sidecarMessage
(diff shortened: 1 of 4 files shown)


validators/nii.js
```diff
@@ -77,6 +77,19 @@ module.exports = function NIFTI (header, file, jsonContentsDict, bContentsDict,
...
 
     if (!mergedDictionary.invalid) {
+
+        // task scan checks
+        if (path.includes('_task-')) {
+            if (!mergedDictionary.hasOwnProperty('TaskName')) {
+                issues.push(new Issue({
+                    file: file,
+                    code: 50,
+                    reason: "You have to define 'TotalReadoutTime' for this file. " + sidecarMessage
```
chrisgorgo: Copy paste error?

constellates: Yep. Good catch. I'll update it.

constellates: Updated.

PR #214, opened 2016-10-10
add support for multiple intendedFor files

jokedurnez:
**intendedFor** key in fmap json file only accepted 1 entry.  With this patch, the validator loops over all entries.

### DIFF
--- a/validators/nii.js
+++ b/validators/nii.js
@@ -204,7 +204,7 @@ module.exports = function NIFTI (header, file, jsonContentsDict, bContentsDict, 
         if (path.includes("_phasediff.nii") || path.includes("_phase1.nii") ||
             path.includes("_phase2.nii") || path.includes("_fieldmap.nii") || path.includes("_epi.nii")){
             if (mergedDictionary.hasOwnProperty('IntendedFor')) {
-              var intendedFor = typeof mergedDictionary['IntendedFor'] == "string" ? {1:mergedDictionary['IntendedFor']} : mergedDictionary['IntendedFor'];
+              var intendedFor = typeof mergedDictionary['IntendedFor'] == "string" ? [mergedDictionary['IntendedFor']] : mergedDictionary['IntendedFor'];
 
               for(var key = 0; key<intendedFor.length; key++){
                 var intendedForFile = "/" + path.split("/")[1] + "/" + intendedFor[key];
--- a/validators/nii.js
+++ b/validators/nii.js
@@ -204,12 +204,7 @@ module.exports = function NIFTI (header, file, jsonContentsDict, bContentsDict, 
         if (path.includes("_phasediff.nii") || path.includes("_phase1.nii") ||
             path.includes("_phase2.nii") || path.includes("_fieldmap.nii") || path.includes("_epi.nii")){
             if (mergedDictionary.hasOwnProperty('IntendedFor')) {
-
-              if (typeof mergedDictionary['IntendedFor'] == "string"){
-                var intendedFor = {1:mergedDictionary['IntendedFor']};
-              } else {
-                var intendedFor = mergedDictionary['IntendedFor'];
-              }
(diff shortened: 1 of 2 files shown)


validators/nii.js
```diff
@@ -204,7 +204,10 @@ module.exports = function NIFTI (header, file, jsonContentsDict, bContentsDict,
         if (path.includes("_phasediff.nii") || path.includes("_phase1.nii") ||
             path.includes("_phase2.nii") || path.includes("_fieldmap.nii") || path.includes("_epi.nii")){
             if (mergedDictionary.hasOwnProperty('IntendedFor')) {
-                var intendedForFile = "/" + path.split("/")[1] + "/" + mergedDictionary['IntendedFor'];
+              var intendedFor = mergedDictionary['IntendedFor'];
+              for(var key in intendedFor){
+                var onTheList = false;
```
chrisgorgo: this variable is declared twice

validators/nii.js
```diff
@@ -204,7 +204,10 @@ module.exports = function NIFTI (header, file, jsonContentsDict, bContentsDict,
         if (path.includes("_phasediff.nii") || path.includes("_phase1.nii") ||
             path.includes("_phase2.nii") || path.includes("_fieldmap.nii") || path.includes("_epi.nii")){
             if (mergedDictionary.hasOwnProperty('IntendedFor')) {
-                var intendedForFile = "/" + path.split("/")[1] + "/" + mergedDictionary['IntendedFor'];
+              var intendedFor = mergedDictionary['IntendedFor'];
+              for(var key in intendedFor){
```
chrisgorgo: We need to support two cases: when IntendedFor is a single path and when it's a list of paths. That's why the tests are failing.

chrisgorgo (changes requested):

validators/nii.js
constellates: Also please switch the loop from a "for in" to a more standard loop like `for (var key = 0; key < intendedFor.length; key++) {}`  

"for in" loops over arrays are prone to causing issues in javascript.

validators/nii.js
```diff
@@ -204,9 +204,14 @@ module.exports = function NIFTI (header, file, jsonContentsDict, bContentsDict,
         if (path.includes("_phasediff.nii") || path.includes("_phase1.nii") ||
             path.includes("_phase2.nii") || path.includes("_fieldmap.nii") || path.includes("_epi.nii")){
             if (mergedDictionary.hasOwnProperty('IntendedFor')) {
-              var intendedFor = mergedDictionary['IntendedFor'];
-              for(var key in intendedFor){
-                var onTheList = false;
+
```
constellates: It looks like tests are still failing due to declaring "intendedFor" in two spots. I haven't tested this but this should work. 

```
var intendedFor;
if (typeof mergedDictionary['IntendedFor'] == "string"){
    intendedFor = [mergedDictionary['IntendedFor']];
} else {
    intendedFor = mergedDictionary['IntendedFor'];
}
```

I also changed the first conditional clause to create an array instead of an object to match how the spec declares multiple intendedFor file paths. 

Also you can run `npm test` and `npm run lint` if you want to run the test processes locally, but committing and having CI run them is also fine.

jokedurnez: oh, I've done it now with a conditional operator, but did't add the array.  Will test locally next time !

constellates:
I'm struggling to find a dataset with fieldmaps on my computer to test the expected functionality but the changes look good to me know. The only thing thats missing is the version number should be incremented in the /package.json file. A version change is required for the validator to get updated in the npm registry.

You can update the version here https://github.com/jokedurnez/bids-validator/blob/master/package.json#L3

For this case bumping the smallest digit should be fine.. so 0.18.18 can be upped to 0.18.19

jokedurnez:
Well it now works on the data I'm working with.

Changed the version, so it should be fine.  Yay, first javascript PR :-)

Thanks for guiding me through it !

constellates:
Awesome. Thanks for the PR!

chrisgorgo (approved):

Issue #201, opened 2016-09-26
Add support for phenotypic data split into multiple files (together with metadata


chrisgorgo:
See https://docs.google.com/document/d/1HFUkAEE-pB-angVcYe6pf_-fVf4sCpOHKesUvfb8Grc/edit#heading=h.pi5iigxxt8vy

constellates:
Hey @chrisfilo I don't have a clear grasp of what needs to be done here. So here are a few questions for you.

From the spec:

> If the dataset includes multiple sets of participant level measurements (for example responses from multiple questionnaires) they can be split into individual files separate from participants.tsv. Those measurements should be kept in phenotype/ folder and end with the .tsv extension.

Are there limitations on where the phenotype folder can be placed in the dataset? Or can it be placed anywhere and lower level directories will inherit this data?

Are there any limitations about how these phenotype files are named (aside from the fact they have to have .tsv extensions and can optionally have a .json version for metadata)?

It looks like the tsv phenotype files must contain participant_id column. Aside from that is there more validation you would like to do on them at this time?

chrisgorgo:
On Tue, Oct 11, 2016 at 1:22 PM, constellates notifications@github.com
wrote:

> Hey @chrisfilo https://github.com/chrisfilo I don't have a clear grasp
> of what needs to be done here. So here are a few questions for you.
> 
> From the spec:
> 
> If the dataset includes multiple sets of participant level measurements
> (for example responses from multiple questionnaires) they can be split into
> individual files separate from participants.tsv. Those measurements should
> be kept in phenotype/ folder and end with the .tsv extension.
> 
> Are there limitations on where the phenotype folder can be place in the
> dataset? Or can it be placed anywhere and lower level directories will
> inherit this data?
> 
> yes it's only in the root
> 
> Are there any limitations about how these phenotype files are named (aside
> from the fact they have to have .tsv extensions and can optionally have a
> .json version for metadata)?
> 
> no there are not limitations
> 
> It looks like the tsv phenotype files must contain participant_id column.
> Aside from that is there more validation you would like to do on them at
> this time?
> 
> nope - this should be sufficient

BTW here's an example dataset https://openfmri.org/dataset/ds000030/

constellates:
Update merged.

PR #216, opened 2016-10-11
Phenotypic data

constellates:
Allow .tsv and .json phenotype data.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bids-validator",
-  "version": "0.18.20",
+  "version": "0.18.21",
   "description": "",
   "main": "index.js",
   "repository": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bids-validator",
-  "version": "0.18.22",
+  "version": "0.18.23",
   "description": "",
   "main": "index.js",
   "repository": {
--- a/tests/type.spec.js
+++ b/tests/type.spec.js
@@ -196,3 +196,15 @@ var suiteDWI = describe('utils.type.isDWI', function(){
         require('assert').ok(true);
     });
 });
+
+describe('utils.type.isPhenotypic', function () {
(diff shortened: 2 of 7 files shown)


constellates:
for issue #201

chrisgorgo:
Could we add the following checks to this feature:
- participant_id column is present in each phenotypic tsv file
- values from participant_id columns match to "sub-*" folders

Basically we should apply the same tests to phenotypic files as we apply to participants.tsv

constellates:
With regards to matching participant_id columns to "sub-*" folders, does every .tsv in the phenotype directory need to have a row for every subject or are there any cases where portions of subjects would be split across multiple .tsv files within the phenotype directory?

chrisgorgo:
I think we should allow subsets here. It is likely that certain subjects
did not get any measurements from a particular tool and then they will not
be included in the corresponding .tsv file.

We should however catch a situation in which a a subject label that does
not occur in the sub-\* folder happens.

constellates:
This has been updated to test phenotype tsvs against the same basic checks as participants.tsv. The one exception is that participants.tsv files must list every subject in the dataset but phenotype files may list only a subset. An error will be thrown if they list any subjects not found in the dataset.

Definitely check the "reason" changes in the list at utils/issues.js to make sure they sound correct. I added one new issue and modified another to let it apply to phenotype and participants files.

constellates:
Also here are two screen shots of the new issue. It looks fine if there's a reasonable amount of subject differences but becomes a mess if there are a ton. Let me know if you'd like any changes to this output.

screen shot 2016-10-17 at 4 31 36 pm

screen shot 2016-10-17 at 4 29 09 pm

chrisgorgo:
Nice! Is this an error you forced or are you getting this on vanilla ds000030 dataset?

constellates:
It's a forced error. I only downloaded the metadata and a subset of the subjects from openfmri.

chrisgorgo:
Great. LGTM. Merge when ready.

Issue #227, opened 2016-10-21
Uncaught TypeError: Cannot read property 'match' of undefined


chrisgorgo:
From https://groups.google.com/d/msg/bids-discussion/wv4_nJfT2C8/TEfWJKMcBAAJ

> OK, this occurs in line 34560 of app.min.js
> HTH
> 
> ```
>  /**
>  * if elements of the two arrays differ by less than one we won't raise a
>  * warning about them. There are a large number of floating point rounding
>  * errors that cause resolutions to be slightly different. Returns true if
>  * the two headers are signifigantly different
>  */
> function headerFieldCompare(header1, header2) {
>     var hdr1 = header1.split(',');
>     var hdr2 = header2.split(',');
>     for (var i = 0; i < hdr1.length; i++) {
>         var hdr1_val = Number(hdr1[i].match(/-?\d*\.?\d*/));
>         var hdr2_val  = Number(hdr2[i].match(/-?\d*\.?\d*/)); // HERE IS THE ERROR Uncaught TypeError: Cannot read property 'match' of undefined
>         // Matching alphas with * will return '' on headers without units
>         var hdr1_unit = hdr1[i].match(/[A-Za-z]*$/)[0];
>         var hdr2_unit = hdr2[i].match(/[A-Za-z]*$/)[0];
>         if (Math.abs(hdr1_val - hdr2_val) > .00001) {
>             return true;
>         }
>         if (hdr1_unit !== hdr2_unit) {
>             return true;
>         }
>     }
>     return false;
> }
> module.exports = headerFields;
> ```

chrisgorgo:
Here's the file that was causing the problem 
[sub-05_ses-07_task-ClipsVal09_acq-ap_bold.nii.gz](https://github.com/INCF/bids-validator/files/548292/sub-05_ses-07_task-ClipsVal09_acq-ap_bold.nii.gz)

We should handle such cases with a more informative error.

PR #228, opened 2016-10-24
Compare header field length before individual values

constellates:
Fix for issue #227. I was unable to directly duplicate the bug with the file attached to the issue. Because the issue happens while comparing header values it relies on the context of multiple scan headers.

However it looks pretty straight forward what happened and I was able to mock some data locally to force this issue. The issue is caused when comparing values that are comma separated lists (dim and pixdim). This validation wasn't accounting for lists of different lengths and only checked the individual values.

So I updated this to consider lists of different lengths to be different dimensions. So 

`[ '3.13mm', '3.13mm', '4.00mm', '1.00s' ]`

is not the same as 

`[ '3.13mm', '3.13mm', '4.00mm']`

I believe the existing issue (below) still explains this well but let me know if you think we need any changes.

```
39: {
    key: 'INCONSISTENT_PARAMETERS',
    severity: 'warning',
    reason: "Not all subjects/sessions/runs have the same scanning parameters."
}
```

Verbose mode will also show

`The most common resolution is: 3.13mm x 3.13mm x 4.00mm x 1.00s, This file has the resolution: 3.13mm x 3.13mm x 4.00m.`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bids-validator",
-  "version": "0.19.1",
+  "version": "0.19.2",
   "description": "",
   "main": "index.js",
   "repository": {
--- a/utils/issues/list.js
+++ b/utils/issues/list.js
@@ -266,5 +266,10 @@ module.exports = {
         key: 'NO_T1W',
         severity: 'ignore',
         reason: "Dataset does not contain any T1w scans."
+    },
+    54: {
+        key: 'BOLD_NOT_4D',
+        severity: 'error',
+        reason: 'Bold scans must be 4 dimensional.'
     }
 };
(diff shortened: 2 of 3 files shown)


chrisgorgo:
This does solve this particular issue (and should be merged), but I am wondering if we can go even further. In issue #227 the problem was that an anatomical scan was accidentally confused with a bold scan. It would be useful if we raised an error if a _bold file is not 4D to avoid such situations in the future.

constellates:
This has been updated to ensure the dim header value of bold scans are 4 dimensional. I tested it with the scan linked in issue #227 . The file name looks different in the results because I had to rename it to avoid throwing naming issues before header checks. Here's the example output.

```
2: Bold scans must be 4 dimensional. (code: 54 - BOLD_NOT_4D)
    /sub-01/func/sub-01_task-balloonanalogrisktask_run-01_bold.nii.gz
        Evidence: header field "dim" = 3,128,128,93
```

Note: The first value in a dim value is the number of dimensions followed by the actual dimension values.

chrisgorgo:
Thanks!

Issue #230, opened 2016-10-25
cannot read property length


chrisgorgo:
From a user:

> The BIDS dataset is within ~/Documents/DE_420_01/, in ./anat, ./dwi, etc. 
> However, when I check the files using the bids-validator web interface, all the files check out as BIDS format. I installed bids-validator and ran bids-validator ~/Documents/DE_420_01/ as well as ~/Documents/DE_420_01/anat, but I get this error:
> 
> /usr/lib/node_modules/bids-validator/cli.js:25
>             } else if (errors.length >= 1 || warnings.length >= 1) {
>                              ^
> 
> TypeError: Cannot read property 'length' of undefined
>     at /usr/lib/node_modules/bids-validator/cli.js:25:30
>     at /usr/lib/node_modules/bids-validator/validators/bids.js:43:29
>     at Object.BIDS.quickTest (/usr/lib/node_modules/bids-validator/validators/bids.js:87:9)
>     at /usr/lib/node_modules/bids-validator/validators/bids.js:39:26
>     at Object.readDir (/usr/lib/node_modules/bids-validator/utils/files.js:89:9)
>     at /usr/lib/node_modules/bids-validator/validators/bids.js:38:29
>     at /usr/lib/node_modules/bids-validator/utils/options.js:20:17
>     at /usr/lib/node_modules/bids-validator/utils/options.js:63:25
>     at Object.module.exports.parse (/usr/lib/node_modules/bids-validator/utils/json.js:29:17)
>     at /usr/lib/node_modules/bids-validator/utils/options.js:53:22

yeunkim:
version information:

$ bids-validator -v
0.19.1
$ node -v
v4.6.1

I downloaded/installed node.js via:
curl -sL https://deb.nodesource.com/setup_4.x | sudo -E bash -
sudo apt-get install -y nodejs

then bids-validator:
npm install -g bids-validator

Thanks!

chrisgorgo:
The dataset causing this error is not BIDS compatible. Nonetheless the previous version of the validator (before the config feature was merged) worked correctly giving: "This does not appear to be a BIDS dataset. For more info go to http://bids.neuroimaging.io/"

constellates:
Thanks for the extra details I was having a hard time reproducing this error until I saw the "This does not appear to be a BIDS dataset" error.

PR #231  should address this issue.

chrisgorgo:
This has been fixed in 0.19.2 - thanks for reporting @yeunkim

PR #231, opened 2016-10-25
Issue 230

constellates:
Fix for issue #230 

In the configuration update I needed to change the issue output and missed a change for handling datasets that fail the quick test. This PR should address that. I've pushed a similar change to gh-pages. I'm also adding a small change to the circle.yml that I hope will address build failures we get whenever we have to update the gh-pages code.

I was hoping to add some unit tests to catch issues like this during parsing the output in the cli.js file, but we'll need to do a little refactoring of the cli first because it relies on a lot of private functions and inputing an actual dataset in a way that makes it difficult to mock data. So I'll open a new issue to address that.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -30,6 +30,7 @@ deployment:
       - rm -rf node_modules/
       - rm -rf tests/
       - git checkout gh-pages
+      - git pull
       - npm install
       - gulp build
       - git status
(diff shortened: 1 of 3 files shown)


chrisgorgo:
Thanks for a quick fix!

Issue #233, opened 2016-10-31
anat/*_defacemask.nii.gz should are part of the spec


chrisgorgo:
should be easy to fix . see here: https://github.com/INCF/bids-validator/blob/master/utils/type.js#L8

constellates:
Is there a specific deface mask file path that's causing issues? It looks like it's already listed as an anat suffix and the current validator is allowing me to have a _defacemask.nii.gz in an anat directory.

constellates:
Or is this literally for a file named "anat/*_defacemask.nii.gz"?

chrisgorgo:
Sorry - it was my bad reporting. It's actually a func issue:

```
        /sub-01/ses-localizer/func/sub-01_ses-localizer_task-movielocalizer_run-1_defacemask.nii.gz
            Evidence: sub-01_ses-localizer_task-movielocalizer_run-1_defacemask.nii.gz
```

So it's this line: https://github.com/INCF/bids-validator/blob/master/utils/type.js#L159

constellates:
No worries. That makes more sense.

constellates:
I also added support for a func/*_defacemask.json because anat will currently allow that but I wasn't sure if it's actually necessary.

PR #234, opened 2016-10-31
Func deface mask

constellates:
For issue #233


chrisgorgo:
This was solved by #239.

PR #237, opened 2016-11-03
Added Evidence when validator complains about participants_id mismatch

suyashdb:


chrisgorgo:
It seems you just did this again - mixed up commits from two features/bugfixes in one pull request.

PR #238, opened 2016-11-03
Added Evidence for participants_id mismatch

suyashdb:


validators/bids.js
```diff
@@ -294,6 +294,7 @@ var BIDS = {
                 if (participants && !utils.array.equals(summary.subjects, participants.list, true)) {
                     self.issues.push(new Issue({
                         code: 49,
+                        evidence: summary.subjects.toString() +" is not \n" + participants.list.toString(),
```
chrisgorgo: Could you also print which list was derived from the participants.tsv file and which one was derived from the folder structure?

suyashdb: Yes, Pushed commits to print those.

validators/bids.js
```diff
@@ -294,6 +294,7 @@ var BIDS = {
                 if (participants && !utils.array.equals(summary.subjects, participants.list, true)) {
                     self.issues.push(new Issue({
                         code: 49,
+                        evidence: "Subjects(derived from BIDS folder structure):" + summary.subjects.toString() +" are not in \n participants.tsv file:" + participants.list.toString(),
```
chrisgorgo: So an example output would be:

> Subjects(derived from BIDS folder structure):[01 02 03 04] are not in 
> participants.tsv file:[01 02 03]

There are a couple of white spaces missing here, but the bigger issue is that this message is still confusing (because subjects 01 02 03 **are** in participants.tsv).

Saying something like:

> Inconsistent lists of subjects detected:
> participants.tsv: [01 02 03 04]
> folder structure: [01 02 03]

seems easier to understand.

chrisgorgo: Also, let's use the word "participants" rather than "subjects".

Issue #250, opened 2016-11-25
why PARTICIPANT_ID_MISMATCH  ?


yarikoptic:
getting
```
2: Subjects found in this dataset did not match the participant_ids found in the participants.tsv file. (code: 49 - PARTICIPANT_ID_MISMATCH)
	/participants.tsv
```
although I believe they are all kosher:

```shell

$ awk '/sub-/{print $1;}' participants.tsv | while read f; do ls -ld $f; done
drwxr-xr-x 1 mvdoc dbic 30 Nov  6 17:22 sub-sid000050                            
drwxr-xr-x 1 mvdoc dbic 24 Nov  6 18:19 sub-sid000056                            
drwxr-xr-x 1 mvdoc dbic 30 Nov  6 19:14 sub-sid000057                            
drwxr-xr-x 1 mvdoc dbic 24 Nov  6 20:07 sub-sid000059                            
drwxr-xr-x 1 mvdoc dbic 30 Nov  6 20:50 sub-sid000060                            
drwxr-xr-x 1 mvdoc dbic 24 Nov  6 21:50 sub-sid000061                            
drwxr-xr-x 1 mvdoc dbic 24 Nov  6 22:48 sub-sid000062                            
drwxr-xr-x 1 mvdoc dbic 24 Nov  6 23:36 sub-sid000063                            
drwxr-xr-x 1 mvdoc dbic 30 Nov  7 00:33 sub-sid000064                            
drwxr-xr-x 1 mvdoc dbic 24 Nov  7 01:35 sub-sid000065                            
drwxr-xr-x 1 mvdoc dbic 24 Nov  7 02:26 sub-sid000066                            
drwxr-xr-x 1 mvdoc dbic 24 Nov  7 03:18 sub-sid000067                            
drwxr-xr-x 1 mvdoc dbic 30 Nov  7 04:15 sub-sid000068                            
drwxr-xr-x 1 mvdoc dbic 30 Nov  7 05:18 sub-sid000069                            
drwxr-xr-x 1 mvdoc dbic 24 Nov  7 06:19 sub-sid000074                            
drwxr-xr-x 1 mvdoc dbic 24 Nov  7 07:09 sub-sid000075                       
drwxr-xr-x 1 mvdoc mvdoc 24 Nov 23 12:24 sub-sid000078                      
drwxr-xr-x 1 mvdoc mvdoc 24 Nov 23 13:24 sub-sid000079                      
[...]

PR #251, opened 2016-11-28
Ensure file is a BIDS file before parsing the subject value

constellates:
Potential fix for #250. See more info in issue.

### DIFF
--- a/utils/type.js
+++ b/utils/type.js
@@ -203,11 +203,11 @@ module.exports = {
     getPathValues: function (path) {
         var values = {}, match;
 
-        // capture subjects
+        // capture subject
         match = (/^\/sub-(\w+)/).exec(path);
         values.sub = match && match[1] ? match[1] : null;
 
-        // capture sessions
+        // capture session
         match = (/^\/sub-\w+\/ses-(\w+)/).exec(path);
         values.ses = match && match[1] ? match[1] : null;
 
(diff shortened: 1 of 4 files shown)


chrisgorgo:
I would recommend refactoring the subject and session "detectors" to use strict regular expressions looking  only for the appropriately named files (top level for subjects, level below the top for sessions, no extra suffixes or prefixes). It would be more robust.

constellates:
That makes sense. I've added updates to use this strategy. I added a `getPathValues` method to the type utils. It will currently return and sub- and ses- values found in the appropriate directory names, but we could expand this to grab other path values. It doesn't check that a path is valid, only that the values are in the correct spots, but the isBIDS check is run before this when generating summary values.

Take a look at the regexs to see if they're reasonable. I'm not super well versed in regex.

utils/type.js
```diff
@@ -190,6 +190,28 @@ module.exports = {
...
+     * Takes a file path and returns and values
+     * found for the following path keys.
+     * sub-
+     * ses-
+     */
+    getPathValues: function (path) {
+        var values = {}, match;
+
+        // capture subject
+        match = (/^\/sub-(\w+)/).exec(path);
```
chrisgorgo: For consistency with other regexps in the validator could we replace `/w` with `[a-zA-Z0-9]+`

PR #253, opened 2016-11-29
Sort participant lists

### LINKED ISSUE #252: Ordering when comparing subject lists
Lists should be ordered when comparison is performed and feedback is presented to the user
image

chrisgorgo:
Fixes #252

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bids-validator",
-  "version": "0.19.9",
+  "version": "0.20.0",
   "description": "",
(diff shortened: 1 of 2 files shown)


constellates:
It might also be good to do a little more formatting for the output evidence. Instead of 

    participantsFromFile.toString();

You can use

    participantsFromFile.join(', ');

The added space will help with making sure those lists wrap well. You can also change the separator to anything you want.



And it looks likes CI is failing because you need to ensure participants is defined before acting on it.

PR #257, opened 2017-02-16
issue-256

suyashdb:
added warning for participants with age 89 or 89+

### DIFF
--- a/utils/issues/list.js
+++ b/utils/issues/list.js
@@ -278,8 +278,8 @@ module.exports = {
         reason: 'JSON file is not formatted according the the schema.'
     },
     56: {
-        key: 'Participants age 89 or 89+',
+        key: 'Participants age 89 or higher',
         severity: 'warning',
-        reason: 'Participants age is 89 or 89+. Tag them as 89+'
+        reason: 'As per section 164.514(C) of "The De-dentification Standard" under HIPPA guidelines, participants with age 89 or higher should be tagged as 89+. More information can be found at https://www.hhs.gov/hipaa/for-professionals/privacy/special-topics/de-identification/#standard'
     }
 };
--- a/utils/issues/list.js
+++ b/utils/issues/list.js
@@ -276,5 +276,10 @@ module.exports = {
         key: 'JSON_SCHEMA_VALIDATION_ERROR',
         severity: 'error',
         reason: 'JSON file is not formatted according the the schema.'
+    },
+    56: {
+        key: 'Participants age 89 or 89+',
+        severity: 'warning',
+        reason: 'Participants age is 89 or 89+. Tag them as 89+'
     }
 };
--- a/validators/tsv.js
+++ b/validators/tsv.js
@@ -143,15 +143,15 @@ module.exports = function TSV (file, contents, fileList, callback) {
 
     }
   // check partcipants.tsv for age 89+
-    var headers = rows[0].split('\t');
+
     if (file.name === 'participants.tsv'){
+      var header = rows[0].replace(/(\r\n|\n|\r)/gm,"").split('\t'); //remove EOL character if any
+      var ageIdColumn = header.indexOf("age");
       for (var i = 0; i < rows.length; i++) {
         var row = rows[i];
-        var row = row.slice(0,-1)   // remove /r after every row
-        var values = row.split('\t');
-        var ageIdColumn = headers.indexOf("age\r");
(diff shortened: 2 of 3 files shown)


utils/issues/list.js
```diff
@@ -276,5 +276,10 @@ module.exports = {
         key: 'JSON_SCHEMA_VALIDATION_ERROR',
         severity: 'error',
         reason: 'JSON file is not formatted according the the schema.'
+    },
+    56: {
+        key: 'Participants age 89 or 89+',
```
chrisgorgo: "89 or higher"

utils/issues/list.js
```diff
@@ -276,5 +276,10 @@ module.exports = {
...
+        severity: 'warning',
+        reason: 'Participants age is 89 or 89+. Tag them as 89+'
```
chrisgorgo: Give more explanation about why this should be changed (privacy concerns). Link to some more information for example: https://www.hhs.gov/hipaa/for-professionals/privacy/special-topics/de-identification/

validators/tsv.js
```diff
@@ -142,6 +142,28 @@ module.exports = function TSV (file, contents, fileList, callback) {
...
     }
+  // check partcipants.tsv for age 89+
+    var headers = rows[0].split('\t');
+    var ageIndex = headers.indexOf('age');
+    if (file.name === 'participants.tsv'){
+      for (var i = 0; i < rows.length; i++) {
+        var row = rows[i];
+        var row = row.slice(0,-1)
+        var values = row.split('\t');
+        var ageIdColumn = headers.indexOf("age\r");
```
chrisgorgo: Isn't this the same as `ageIndex` as defined in line 147?

validators/tsv.js
```diff
@@ -142,6 +142,28 @@ module.exports = function TSV (file, contents, fileList, callback) {
...
+        age = values[ageIdColumn]
+        if (age == 89 || age > 89) {
```
chrisgorgo: Isn't his the same as `age >= 89`?

validators/tsv.js
```diff
@@ -142,6 +142,28 @@ module.exports = function TSV (file, contents, fileList, callback) {
...
+            age_89_plus = true;
```
chrisgorgo: This variable is not used anywhere.

validators/tsv.js
chrisgorgo: Als don't you need to cast this to float to do this comparison?

validators/tsv.js
```diff
@@ -142,6 +142,28 @@ module.exports = function TSV (file, contents, fileList, callback) {
         }
 
     }
+  // check partcipants.tsv for age 89+
+    var headers = rows[0].split('\t');
+    var ageIndex = headers.indexOf('age');
+    if (file.name === 'participants.tsv'){
+      for (var i = 0; i < rows.length; i++) {
+        var row = rows[i];
+        var row = row.slice(0,-1)
```
chrisgorgo: Why is this necessary?

chrisgorgo (changes requested):
Please add a test.

validators/tsv.js
suyashdb: as every row in tsv file ends with /r, sliced last two characters of every rows. 
For eg:  sub-01    group-A    20/r

chrisgorgo:
Are you sure the end of the line characters are included? If so we need
take care of files created with UNIX systems (new line characters) and
windows (new line and carriage return characters). It might be better to
deal with this when you cast from string to a number.

suyashdb:
It was the case for the test dataset(ds212) participant file, i have been
using. Let me check other tsv files too.

suyashdb:
When i checked other tsv files, i observed that not all tsv files had EOL characters. Hence, first tried to deal with EOL characters while casting from string to Numbers. However, in case of header row, because in some files there was age\r or age\n and in others just age, i had to trim these EOL char while reading rows itself. 

For this reason, i went with the above logic (as per latest commit). 
Please let me know if you foresee any limitation in using this logic.

```
var header = rows[0].replace(/(\r\n|\n|\r)/gm,"").split('\t'); //remove EOL character if any
var ageIdColumn = header.indexOf("age");
```

I will add test in another commit.

chrisgorgo:
This might be helpful https://www.w3schools.com/jsref/jsref_trim_string.asp

Issue #261, opened 2017-03-03
Big dataset problem


ntraut:
When the database is too big (~800 subjects), I get this error:

> /pasteur/homes/ntraut/local/lib/node_modules/bids-validator/validators/tsv.js:14
>     var rows = contents.split('\n');
>                        ^
> 
> TypeError: Cannot read property 'split' of undefined
>     at TSV (/pasteur/homes/ntraut/local/lib/node_modules/bids-validator/validators/tsv.js:14:24)
>     at /pasteur/homes/ntraut/local/lib/node_modules/bids-validator/validators/bids.js:162:21
>     at ReadFileContext.callback (/pasteur/homes/ntraut/local/lib/node_modules/bids-validator/utils/files.js:48:17)
>     at FSReqWrap.readFileAfterOpen [as oncomplete] (fs.js:303:13)

This error does not happen in all system environments and is caused by abnormal empty callback returns from `fs.readFile`. When I reduce the size of the dataset, `bids-validator` works properly.

I updated all the dependencies to their latest version and now I get this error with large datasets:

> /pasteur/homes/ntraut/src/bids-validator/utils/files.js:239
>         var pathParts = relPath.split('/');
> ^
> 
> RangeError: Maximum call stack size exceeded
>     at String.split (native)
>     at Object.relativePath (/pasteur/homes/ntraut/src/bids-validator/utils/files.js:239:33)
>     at /pasteur/homes/ntraut/src/bids-validator/validators/bids.js:120:36
> at replenish (/pasteur/homes/ntraut/src/bids-validator/node_modules/async/dist/async.js:881:17)
>     terateeCallback (/pasteur/homes/ntraut/src/bids-validator/node_modules/async/dist/async.js:866:17)
> at /pasteur/homes/ntraut/src/bids-validator/node_modules/async/dist/async.js:843:16
> at /pasteur/homes/ntraut/src/bids-validator/validators/bids.js:149:17
> at replenish (/pasteur/homes/ntraut/src/bids-validator/node_modules/async/dist/async.js:881:17)
>     terateeCallback (/pasteur/homes/ntraut/src/bids-validator/node_modules/async/dist/async.js:866:17)
> at /pasteur/homes/ntraut/src/bids-validator/node_modules/async/dist/async.js:843:16
> at /pasteur/homes/ntraut/src/bids-validator/validators/bids.js:149:17
[...]

PR #262, opened 2017-03-03
fix to avoid to much concurrent operations

ntraut:
For correcting issue #261: Big dataset problem.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,24 +22,24 @@
   },
   "homepage": "https://github.com/INCF/bids-validator",
   "dependencies": {
-    "async": "^1.2.1",
+    "async": "^2.1.5",
     "bytes": "^2.3.0",
     "cliff": "^0.1.10",
     "colors": "^1.1.0",
     "jshint": "^2.8.0",
     "minimatch": "3.0.3",
     "nifti-js": "^1.0.0",
-    "pako": "^0.2.8",
-    "pluralize": "^1.2.1",
-    "yargs": "^3.25.0",
+    "pako": "^1.0.4",
+    "pluralize": "^3.1.0",
+    "yargs": "^6.6.0",
     "ajv": "^4.9.0"
   },
   "devDependencies": {
     "adm-zip": "",
     "chai": "",
-    "eslint": "^2.7.0",
-    "mocha": "^2.2.5",
-    "publish": "^0.5.0",
-    "sync-request": "2.2.0"
+    "eslint": "^3.16.1",
+    "mocha": "^3.2.0",
+    "publish": "^0.6.0",
+    "sync-request": "4.0.1"
   }
 }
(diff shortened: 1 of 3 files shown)


constellates:
Nice. This looks good to me. Two minor things.

Any of the process.nextTick calls that take a generic async callback don't need the extra anonymous function.

For example this

    process.nextTick(function() { cb() });

could be written as this

    process.nextTick(cb);


And lastly, for the current npm publishing work flow to work you'll need to up the version number in the package.json file.

Thanks for the PR.

ntraut:
Thanks!
I submitted a new commit taking into account your suggestions.

constellates:
Oh.. I actually meant the version of the bids validator needs to be changed. So in package.json the `version` property (line 3) needs to be changed.

	{
		"name": "bids-validator",
		"version": "0.20.0",
		"description": "",
		"main": "index.js",
		"repository": {
			"type": "git",
			"url": "https://github.com/INCF/bids-validator.git"
		},
		...
	}

Probably a "minor" version bump.

	{
		"name": "bids-validator",
		"version": "0.21.0",
		"description": "",
		"main": "index.js",
		"repository": {
			"type": "git",
			"url": "https://github.com/INCF/bids-validator.git"
		},
		...
	}

The dependency version updates are probably still a good thing though and it looks like tests are still passing with them.

ntraut:
OK

chrisgorgo:
Is this ready to merge?

ntraut:
For me yes: it is working, and it solved the issues that I encountered with a big number of files.

constellates (dismissed):
I just realized one big issue with this approach is `process.nextTick` won't exist in the browser. It looks like there are some polyfills, but they don't seem to be widely used so I'm unsure right now what other problems they might pose.

Here's one example https://www.npmjs.com/package/browser-next-tick

@ntraut have you tried this with keeping the async limit, but not using process.nextTick()?

ntraut:
As you want but I think this issue is outdated because I just tested the browser version and it worked.
You can check on your side, modify my pull request (I think it can possibly work just with the async limit, anyway it is better with it), or reject it.

constellates:
@ntraut Thanks for testing this in the browser. I just dug into the issue a little deeper and it looks like browserify (a part of our web build flow) automatically shims `process.nextTick()` during the build process.

So @chrisfilo I'm good with these changes now with the caveat that this means the validator will continue to work in our applications and in node, but anyone else using it in the browser will need to make sure they have a shim for `process.nextTick()` in place. 

We could also build a shim into the validator.

chrisgorgo:
Thanks for looking into this.

Wouldn't everyone use browserify before trying to use it in the browser? By saying our applications do you also include "http://incf.github.io/bids-validator"?

constellates:
Yea by "our applications" I mean the gh-pages and the crn_app repo. There are a number of ways to bundle a library like this for the web, but yes most will likely include shims for various node specifics by default. I just checked and webpack also includes a shim.

Issue #269, opened 2017-03-14
IntendedFor missing file error not specific to number of missing files


jbwexler:
If one of the files listed in the IntendedFor field does not exist, the error raised states that all of the files listed don't exist. Example:

/sub-010029/ses-02/fmap/sub-010029_ses-02_acq-GEfmap_run-02_phasediff.nii.gz
			'IntendedFor' property of this fieldmap ('ses-02/func/sub-010029_ses-02_task-rest_acq-AP_run-02_bold.nii.gz,ses-02/func/sub-010029_ses-02_task-rest_acq-PA_run-02_bold.nii.gz') does not point to an existing file. Please mind that this value should not include subject level directory ('/sub-010029/').

In this example, 'ses-02/func/sub-010029_ses-02_task-rest_acq-AP_run-02_bold.nii.gz' actually does exists and 'ses-02/func/sub-010029_ses-02_task-rest_acq-PA_run-02_bold.nii.gz' does not. However, the error states that both don't exist.

suyashdb:
@jbwexler I tried recreating the error with latest version on bids but failed.. however was able to generate error using bids-validator installed on TACC.
Can you install latest version and check if missing IntendedFor file error is generated..?   
Looks like newer version is not generating the error wheee intended for files are missing.

PR #276, opened 2017-04-20
269

suyashdb:
specific file missing error when IntendedFor file din fieldmap json is checked. Also added test for the same.

### DIFF
--- a/tests/nii.spec.js
+++ b/tests/nii.spec.js
@@ -97,20 +97,13 @@ describe('NIFTI', function(){
 
       var jsonContentsDict = {
           '/sub-09/ses-test/dwi/sub-09_ses-test_run-01_fieldmap.json': {
-              EchoTime: 1,
-              PhaseEncodingDirection: 3,
-              EffectiveEchoSpacing: 5,
-              SliceTiming: 3,
-              SliceEncodingDirection: 4,
-              RepetitionTime: 1,
-              TotalReadoutTime: 3,
               TaskName: 'Mixed Event Related Probe',
               IntendedFor: ['sub-15_task-mixedeventrelatedprobe_run-01_bold.nii.gz','sub-15_task-mixedeventrelatedprobe_run-02_bold.nii.gz'
             ]
           }
         }
-      validate.NIFTI(null, file, jsonContentsDict, {}, [], events, function (issues) {
-          assert(issues.length = 1 && issues[0].code == 37);
+      validate.NIFTI(null, file, jsonContentsDict, {}, [], [], function (issues) {
+          assert(issues.length = 2 && issues[0].code == 17  && issues[1].code == 37);
       });
     })
 
--- a/tests/nii.spec.js
+++ b/tests/nii.spec.js
@@ -105,9 +105,8 @@ describe('NIFTI', function(){
               RepetitionTime: 1,
               TotalReadoutTime: 3,
               TaskName: 'Mixed Event Related Probe',
-              IntendedFor: {'sub-15_task-mixedeventrelatedprobe_run-01_bold.nii.gz',
-              'sub-15_task-mixedeventrelatedprobe_run-02_bold.nii.gz'
-            }
+              IntendedFor: ['sub-15_task-mixedeventrelatedprobe_run-01_bold.nii.gz','sub-15_task-mixedeventrelatedprobe_run-02_bold.nii.gz'
+            ]
           }
         }
       validate.NIFTI(null, file, jsonContentsDict, {}, [], events, function (issues) {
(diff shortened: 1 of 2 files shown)


suyashdb:
Closes #269.

suyashdb:
Hey @ntraut, 
with `process.nextTick(cb);`, validators skipped checking for IntendedFor field completely(or didn't give any errors) even when files listed in `IntendedFor` in the json didn't exist. 

I have made following  two changes. Wanted you to take a look:

i) Changed `async.eachOfLimit()` to 'async.forEachOf()' . [Here.](https://github.com/INCF/bids-validator/pull/276/files#diff-60709260fc11f2ede4c5ea6724588606L217)
As per @chrisfilo suggestion i tried both eachOfLimit and forEachOf. They both work fine but only with explicit call back `cb()`. Can you share more on switching to using `async.eachOfLimit()` method?   

ii) `process.nextTick(cb)` to `cb()`

ntraut:
Hello, here are the reasons why I made those changes:

i) `async.eachOfLimit()`
It prevents Node from opening too many iterations in parallel. The risk is that too many files are opened at the same time and the system limitations are exceeded.

ii) `process.nextTick(cb)`
With a direct call to `cb()`, each call add itself to the stack and if there are too many iterations the call stack limit can be exceeded.

Normally the way of functioning of `async.eachOfLimit()` is that each time an iteration is over it free the place for the next ones, so I do not see why it would prevent some iterations or code to be executed...

ntraut:
It is possible that `async.eachOfLimit()` also solves the problem that `process.nextTick(cb)` is supposed to settle, but `async.eachOfLimit()` is mandatory for data sets with a large number of files.

suyashdb:
Thank you @ntraut! I have made changes to use `async.eachOfLimit()` and `cb()`. 
Closes #269

chrisgorgo:
One remaining thing is to add a test with a correct IntendedFor information.

suyashdb:
@ntraut @chrisfilo 
My bad. Previously i tried  this PR edits only on smaller dataset hence confirmed that its working but apparently it doesn't. 

1) When i use `async.eachOfLimit()` and `cb()`, it does throw error where intended files are missing but doesn't work for _large datasets_. 

2) Keeping `async.eachOfLimit()` as it is and changing `cb()` to `process.nextTick(cb)` doesn't throw error about IntendedFor missing files at all..

``` 

/home1/04275/suyashdb/node_modules/async/dist/async.js:825
        return i < len ? { value: obj[key], key: key } : null;
                         ^

RangeError: Maximum call stack size exceeded
    at next (/home1/04275/suyashdb/node_modules/async/dist/async.js:825:26)
    at replenish (/home1/04275/suyashdb/node_modules/async/dist/async.js:872:28)
    at iterateeCallback (/home1/04275/suyashdb/node_modules/async/dist/async.js:866:17)
    at /home1/04275/suyashdb/node_modules/async/dist/async.js:843:16
    at NIFTI.async.eachOfLimit.issues.push.Issue.file (/home1/04275/suyashdb/bids-validator/validators/nii.js:217:21)
    at replenish (/home1/04275/suyashdb/node_modules/async/dist/async.js:881:17)
    at iterateeCallback (/home1/04275/suyashdb/node_modules/async/dist/async.js:866:17)
    at /home1/04275/suyashdb/node_modules/async/dist/async.js:843:16
    at NIFTI.async.eachOfLimit.issues.push.Issue.file (/home1/04275/suyashdb/bids-validator/validators/nii.js:217:21)
    at replenish (/home1/04275/suyashdb/node_modules/async/dist/async.js:881:17)
 ```

ntraut:
@suyashdb 
I'm sorry. With `process.nextTick`, do you get the bug even with small datasets?
Can you provide an example dataset with which you get the bug?
If I can help...

suyashdb:
@ntraut  

can you clarify what you mean by bug given we are talking about two issues here:
1) validator not throwing error regrading missing intendedfor files 
2) validator not able to handle large dataset..

Looks like we don't have large published dataset with fieldmap images that can help you as a test dataset. I am working on one but its still unpublished.

ntraut:
If I understand correctly the bug 1. is when `process.nextTick` is used and happens with all sizes of dataset. I just wanted to be sure and was asking if there were an avaible dataset with which I could reproduce this bug.

suyashdb:
@ntraut - yes your understanding is correct. Can you try using ds000223. For first 5 subjects they have fmap data and hence `IntendedFor` in jsons. Try deleting the files in IntendedFor field. 
With current master(which uses `process.nextTick`), validator shouldn't raise any error that IntendedFor file is missing.

ntraut:
OK, I saw the problem: if the file linked by the `IntendedFor` field is missing, the test is done and an issue is raised even with `process.nextTick`. But because it is raised in a `async` loop, the `issues` variable of the function `NIFTI` is already returned at this time.

I guess the reason it works with hard `cb()` calls is that with them the code is not really executed asynchronously.

Two solutions would be possible:
1) give up `async.eachOf` and come back to an old `for` loop.
2) rethink the architecture of the code for a proper asynchronous management.

PR #296, opened 2017-08-19
Strip leading / from the paths reported

### LINKED ISSUE #294: [begging] please strip leading '/' from the paths reported
I think I have whined before but because it is a constant annoyance I guess I might keep coming.

Most frequent pattern for use of the validator is to run it, quickly cut paste the path into clipboard and paste into the command to view or edit a file. Linux terminals are super good at selecting contiguous portions of text, such as paths, so usually it is a quick double click on a path to select it all.  BUT with bids-validator it would not work since it uses leading `/` in the paths as they are from some kind of root.

Could
`/sub-sid000021/ses-actions1/sub-sid000021_ses-actions1_scans.tsv`
become a 
`sub-sid000021/ses-actions1/sub-sid000021_ses-actions1_scans.tsv`
or
`./sub-sid000021/ses-actions1/sub-sid000021_ses-actions1_scans.tsv`
the world would become such a better place!

validator 0.21.3

suyashdb:
closes #294


yarikoptic (commented):
But why not just lstrip the relative path and save us two characters? What am I missing?

yarikoptic:
I meant to lstrip off /

suyashdb:
@yarikoptic wouldn't adding ./ serve the same purpose. 
Any particular reason/pro's which will be served well with stripped '/' against adding './'..?

yarikoptic:
Just not needed... Takes away valuable 2 characters. Do not remember if on Windows there is .
But I could live with either one. Thanks!

chrisgorgo:
@NellH would this change break anything in the openneuro side?

nellh:
@chrisfilo This should be fine to change without breaking anything in openneuro.

PR #306, opened 2017-09-08
Fix out of array bounds issue with illegal character checks

nellh:
This does not work in the browser because it descends into the Array.prototype methods after the list elements if they are iterable. After iterating over the expected items in the array, it sets err_regex to a function and that function does not have the expected regexp.exec property.

```javascript
for (var err in illegalchar_regex_list) {
  var err_regex = illegalchar_regex_list[err][0];
  var err_code = illegalchar_regex_list[err][1];
  var err_evidence = illegalchar_regex_list[err][2];
  if (err_regex.exec(completename)){
    ...
  }
}
```

The test added should catch this in the future for bad use of for ... in within fullTest.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,5 +5,6 @@ tests/data/BIDS-examples-1.0.0-rc2
 tests/data/BIDS-examples-1.0.0-rc3u2
 tests/data/BIDS-examples-1.0.0-rc3u3
 tests/data/BIDS-examples-1.0.0-rc3u4
+tests/data/BIDS-examples-1.0.0-rc3u5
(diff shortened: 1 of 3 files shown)


chrisgorgo:
Linter seems to be unhappy:

```
/home/ubuntu/bids-validator/tests/bids.spec.js
  114:76  error  Missing semicolon       semi
  116:9   error  'after' is not defined  no-undef
```

nellh:
@chrisfilo Sorry about that, updated with a linting fix.

chrisgorgo:
New version is up on npm.

PR #308, opened 2017-09-11
Fix various spelling issues in error codes

nellh:
I noticed an issue with the new one so I ran the strings through a spell checker.


chrisgorgo:
Thanks!

PR #312, opened 2017-09-14
Enable Firefox support for Github pages demo

nellh:
This enables Firefox support for the gh pages demo since it is working in Firefox 49 and later.


chrisgorgo:
This is pretty cool! We should consider enabling Firefox support on OpenNeuro in the future.

nellh:
@chrisfilo Yeah, OpenNeuro also pretty much works on Firefox now but it does need a styling pass to catch some CSS issues that hide information before we can enable it there.

PR #318, opened 2017-09-28
Do not check files in derivatives for illegal characters in sub,ses, task,acq name

### LINKED ISSUE #317: "code: 58 - TASK_NAME_CONTAIN_ILLEGAL_CHARACTER" should not check in derivatives/
Given that derivatives filenames are currently freeform, they should not be required to meet this standard. Perhaps this could a warning if it's in derivatives (or sourcedata, etc.)

suyashdb:
Closes #317


validators/bids.js
```diff
@@ -173,18 +173,19 @@ BIDS = {
 
         for (var f in fileList) {
             var completename = fileList[f].relativePath;
+            if(!completename.startsWith('/derivatives')) {
```
chrisgorgo: We should include `/code` and `/sourcedata` here as well.

chrisgorgo (changes requested):

suyashdb (commented):
I have added /code and /sourcedata to the check

chrisgorgo (approved):

PR #319, opened 2017-09-28
Dataset description Funding converted to List

### LINKED ISSUE #314: validator fails when funding is a list rather than a single string
with the following in dataset_description.json:

   "Funding": [
        "TODO",
        "GRANT #1",
        "GRANT #2"
    ],

I get:

	1: JSON file is not formatted according the the schema. (code: 55 - JSON_SCHEMA_VALIDATION_ERROR)
		./dataset_description.json
			Evidence: .Funding should be string

when it's changed to a string rather than a list, the validation is fine.

validation should be changed to accept either a string or a list of strings.

suyashdb:
closes #314


validators/bids.js
```diff
@@ -173,18 +173,19 @@ BIDS = {
 
         for (var f in fileList) {
             var completename = fileList[f].relativePath;
+            if(!completename.startsWith('/derivatives')) {
```
chrisgorgo: we should probably also exclude /code and /sourcedata

validators/schemas/dataset_description.json
```diff
@@ -12,7 +12,10 @@
             "type": "string"
         },
         "Funding": {
-            "type": "string"
+            "items": {
+                "type": "string"
+            },
+            "type": "array"
```
chrisgorgo: I wonder if for backward compatibility we should not allow both string and array of strings

chrisgorgo:
I think you accidentally added `.idea` folder once again. Might be worth adding it to .gitignore.

validators/schemas/dataset_description.json
suyashdb: May be i didn't understand the above concern. Do you mean to say for backward  compatibility we **_should allow_** both string and array of strings. Above changes in this PR compels one to have _List_ _only_ ..

validators/bids.js
suyashdb: @chrisfilo  True.. will add those two directory too.

validators/schemas/dataset_description.json
chrisgorgo: Apologies for misspeaking - "we should".

suyashdb: Np. Will incorporate that!

suyashdb:
@chrisfilo I have added **Funding** and **HowToAcknowldge** to be either list or list of string. 
I have seen **HowToAcknowldge** also in both formats. 
Let me know if you prefer **HowToAcknowldge** to be string only.

validators/schemas/dataset_description.json
```diff
@@ -12,10 +12,13 @@
...
         },
         "Funding": {
-            "type": "string"
+            "items": {
+                "type": "string"
+            },
+            "type": ["array", "string"]
         },
         "HowToAcknowledge": {
-            "type": "string"
```
chrisgorgo, line 18: doesn't this one also require specification that items need to be strings?

chrisgorgo:
Could you merge master or rebase?

suyashdb:
Ready for review.

validators/schemas/dataset_description.json
chrisgorgo: Any thoughts on my comment?

validators/schemas/dataset_description.json
```diff
@@ -12,10 +12,13 @@
...
+            "type": ["string", "array"]
```
suyashdb: yes, for "HowToAcknowledge" as well i think giving option of string and array of strings would make it backward compatible.

chrisgorgo: I'm talking about adding
```
"items": {
                "type": "string"
},
```
to HowToAcknowledge the same way you did for Funding.

suyashdb: Oh i see. I have pushed commit with those changes.

PR #339, opened 2017-10-15
Check for presence of compulsory datasetdescription.json file

suyashdb:


chrisgorgo:
It seems that you accidentally included a bunch of nifti files. Please start over from a clean, up to date version of the master branch (to avoid polluting history with adding and seeing large files).

suyashdb:
Oh Sorry. removed those now.

chrisgorgo:
How big were those files? We want to avoid the .git folder to grow unnecessarily (see https://stackoverflow.com/questions/2100907/how-to-remove-delete-a-large-file-from-commit-history-in-git-repository).

suyashdb:
I removed those nifti's from .git/history but now it is not allowing me push as git remote branch has one more commit which is about deleting these files.
in my local branch, when i deleted history of nifti files, it automatically deleted the commit with deletion of these files, making head of remote branch and local branch un even.
I am closing this branch and will create a new branch for same PR.

PR #346, opened 2017-10-18
Optimize traversal of directory

suyashdb:
During directory traversal, `['derivatives', 'sourcedata']` is removed from dir list.


suyashdb:
closes issue #346.

suyashdb:
Ready for review.

utils/files.js
```diff
@@ -77,7 +77,8 @@ function readFile (file, callback) {
  */
 function readDir (dir, callback) {
     if (fs) {
-        var files = getFiles(dir);
+        var doNotTraverseDirList = ['derivatives', 'sourcedata']; //list of directories to be excluded from
```
chrisgorgo: We should add "code" on this list.

utils/files.js
```diff
@@ -94,9 +95,11 @@ function readDir (dir, callback) {
     }
 }
 
-function getFiles (dir, files_){
+function getFiles (dir, files_, toNotTraverseDirList_){
     files_ = files_ || [];
+    toNotTraverseDirList_ = toNotTraverseDirList_ || [];
     var files = fs.readdirSync(dir);
+    files = files.filter(function(x){return toNotTraverseDirList_.indexOf(x)<0;}); // this will filter directories which are not to be traversed
```
chrisgorgo: This only catches top level directories right?

utils/files.js
chrisgorgo: and "stimuli"

chrisgorgo (changes requested):
We should also remove the other checks in here: https://github.com/INCF/bids-validator/blob/c82f0064ad95c8fee5a3f42776d590444acbcf48/validators/bids.js#L185 and here https://github.com/INCF/bids-validator/blob/c82f0064ad95c8fee5a3f42776d590444acbcf48/validators/bids.js#L219

utils/files.js
suyashdb: Yes this will catch only top level directories.

chrisgorgo:
You missed https://github.com/INCF/bids-validator/blob/c82f0064ad95c8fee5a3f42776d590444acbcf48/validators/bids.js#L219

suyashdb:
Ready for review.

utils/files.js
```diff
@@ -77,7 +77,8 @@ function readFile (file, callback) {
  */
 function readDir (dir, callback) {
     if (fs) {
-        var files = getFiles(dir);
+        var doNotTraverseDirList = ['derivatives', 'sourcedata', 'code', 'stimuli']; //list of directories to be excluded from
```
chrisgorgo: ok - it also seems that we should add '.git' here as well. Sorry for not catching this earlier.

utils/type.js
```diff
@@ -61,7 +61,7 @@ module.exports = {
      * Check if file is appropriate associated data.
      */
     isAssociatedData: function (path) {
-        var associatedData = new RegExp('^\\/(?:code|derivatives|sourcedata|stimuli|[.]git)\\/(?:.*)$');
```
chrisgorgo: I would reverse those changes and remove all unnecessary calls to this function (for example the one here: https://github.com/INCF/bids-validator/blob/c82f0064ad95c8fee5a3f42776d590444acbcf48/validators/bids.js#L219) instead.

chrisgorgo:
Also ignore all `.` files and folders (and their subfolders) at any folder level. See #348

suyashdb:
ready for review. .

chrisgorgo:
Have you tested it in the browser?

utils/type.js
chrisgorgo: I still see calls to `isAssociatedData` function. Furthermore as stated above I don't think we should or need to modify this function.

suyashdb:
I tried for about 2-3 hours but couldn't get browserify work properly.

chrisgorgo:
What about my other comment? https://github.com/INCF/bids-validator/pull/346#discussion_r149762423

utils/type.js
suyashdb: Oh sorry i didn't explain why i kept call to this function.. 

`isAssociatedData` is being called at many different checks other than just checking `.git` directory. 
Removing call to this function borke many other tests, hence i kept call to isAssociatedData.

chrisgorgo: Could you elaborate a bit on why we would still need the call I listed above (https://github.com/INCF/bids-validator/blob/c82f0064ad95c8fee5a3f42776d590444acbcf48/validators/bids.js#L219)? What other places do you think it is necessary to use that function and why?

chrisgorgo:
Please open an issue about the problems with the instructions for building the web version locally.

chrisgorgo:
After https://github.com/INCF/bids-validator/pull/367 will be merged it should be easier to test the web version.

suyashdb:
i checked web version of validator for this PR. It works fine.

chrisgorgo:
I just tested it and it does not seem to ignore files in `/derivatives`.
image

chrisgorgo:
However the command line version does seem to work as expected 
```
filo@MSI MINGW64 /d/drive/workspace/bids-validator ((347d2f6...))
$ bin/bids-validator /d/data/ds000051_R2.0.2/
        1: You should define 'SliceTiming' for this file. If you don't provide this information slice time correction will not be possible. (code: 13 - SLICE_TIMING_NOT_DEFINED)
                ./sub-01/func/sub-01_task-abstractconcretejudgment_run-01_bold.nii.gz
                ./sub-01/func/sub-01_task-abstractconcretejudgment_run-02_bold.nii.gz
                ./sub-01/func/sub-01_task-abstractconcretejudgment_run-03_bold.nii.gz
                ./sub-01/func/sub-01_task-abstractconcretejudgment_run-04_bold.nii.gz
                ./sub-01/func/sub-01_task-abstractconcretejudgment_run-05_bold.nii.gz
                ./sub-01/func/sub-01_task-abstractconcretejudgment_run-06_bold.nii.gz
                ./sub-01/func/sub-01_task-abstractconcretejudgment_run-07_bold.nii.gz
                ./sub-01/func/sub-01_task-abstractconcretejudgment_run-08_bold.nii.gz
                ./sub-02/func/sub-02_task-abstractconcretejudgment_run-01_bold.nii.gz
                ./sub-02/func/sub-02_task-abstractconcretejudgment_run-02_bold.nii.gz
                ... and 92 more files having this issue (Use --verbose to see them all).

        2: Not all subjects contain the same files. Each subject should contain the same number of files with the same naming unless some files are known to be missing. (code: 38 - INCONSISTENT_SUBJECTS)
                ./sub-01/anat/sub-01_run-02_inplaneT2.nii.gz
                ./sub-01/anat/sub-01_run-03_T1w.nii.gz
                ./sub-02/anat/sub-02_run-02_inplaneT2.nii.gz
                ./sub-02/anat/sub-02_run-03_T1w.nii.gz
                ./sub-03/anat/sub-03_run-02_inplaneT2.nii.gz
                ./sub-03/anat/sub-03_run-03_T1w.nii.gz
                ./sub-04/anat/sub-04_run-02_inplaneT2.nii.gz
                ./sub-04/anat/sub-04_run-03_T1w.nii.gz
                ./sub-05/anat/sub-05_run-02_inplaneT2.nii.gz
                ./sub-05/anat/sub-05_run-03_T1w.nii.gz
                ... and 18 more files having this issue (Use --verbose to see them all).

        Summary:                 Available Tasks:                  Available Modalities:
        249 Files, 2.11GB        abstract-concrete judgment        inplaneT2
        13 - Subjects                                              T1w
        1 - Session                                                bold
```

suyashdb:
let me work on it again..

suyashdb:
i tried implementing filter function https://github.com/suyashdb/bids-validator/blob/optimizeTraversalOfDirectory/utils/files.js#L102-L107 in 3 different ways inspired by previous issue similar to this where changing `forEach` to `for` helped us. 
But in this case, none of them worked. Any pointers that you think might be causing this. meanwhile i am going through various stack overflow questions.

chrisgorgo:
The code you wrote is not even called when operating in the browser: 
https://github.com/suyashdb/bids-validator/blob/optimizeTraversalOfDirectory/utils/files.js#L9
https://github.com/suyashdb/bids-validator/blob/optimizeTraversalOfDirectory/utils/files.js#L92

chrisgorgo:
In other words your filetering code should be outside of this `if` statements (so it would be applied in either case - if the validator is used as a command line or if the validator is used in the browser):
https://github.com/suyashdb/bids-validator/blob/optimizeTraversalOfDirectory/utils/files.js#L78

chrisgorgo:
`fs` and `zlib` are only available if the code is running via node.js (commandline) not in the browser. This is why the conditional import was put there to begin with.

suyashdb:
yes true. There is discussion on stackoverflow about other fs like module(for eg: browserify-fs) which will work with browser. Though it's not been useful or my implementation have been incorrect.

chrisgorgo:
I don't think you need any additional modules. Consider that the original `readDir` function called the `callback` function on an array of file objects. In node.js context the `readDir` function was taking a string - path to a directory - and traversing the file system recursively to create the array of file objects. In context of a browser this was not necessary and the `dir` argument of the `readDir` function is already an array of file objects. That is why in context of a browser the `readDir` [does not do much](https://github.com/INCF/bids-validator/blob/8d8c93eabb35ce10f72193a5f71953d07879fa54/utils/files.js#L93).

To implement the required functionality you need to filter the array of file objects before `callback` is called.

suyashdb:
may be my understanding in limited, but i haven't understood how "the `dir` argument of the `readDir` function is already an array of file objects". In the case of browser `fs` is not loaded control wont go inside if condition rather it will directly proceed to `else` clause where callback is called with `dir` passed as another parameter.

PR #352, opened 2017-10-27
Validate slice timing array with RepetitionTime

### LINKED ISSUE #302: Validate that each value in the SliceTiming array is smaller than RepetitionTime

suyashdb:
Closes #302.

### DIFF
--- a/tests/json.spec.js
+++ b/tests/json.spec.js
@@ -24,13 +24,4 @@ describe('JSON', function(){
 			assert(issues && issues.length === 1);
 		});
 	});
-
-	it('sidecar SliceTiming shouldnot be greator than RepetitionTime', function(){
-		var jsonObjInval = '{"RepetitionTime": 1.2, "echo_time": 0.005, "flip_angle": 90, "TaskName": "Rest", "SliceTiming":[0.0, 1.7759, 0.0517, 0.8276, 0.1034, 1.8793, 0.1552]}';
-		validate.JSON(file, jsonObjInval, function (issues) {
-			assert(issues.length === 1 && issues);
-		});
-	});
-
-
 });
--- a/tests/json.spec.js
+++ b/tests/json.spec.js
@@ -25,4 +25,13 @@ describe('JSON', function(){
 		});
 	});
 
+	it('sidecar SliceTiming shouldnot be greator than RepetitionTime', function(){
+		var jsonObjInval = '{"RepetitionTime": 1.2, "echo_time": 0.005, "flip_angle": 90, "TaskName": "Rest", "SliceTiming":[0.0, 1.7759, 0.0517, 0.8276, 0.1034, 1.8793, 0.1552]}';
+		validate.JSON(file, jsonObjInval, function (issues) {
+			console.log(issues.lenth);
+			assert(issues.length === 1 && issues);
+		});
+	});
+
+
 });
--- a/tests/json.spec.js
+++ b/tests/json.spec.js
@@ -28,7 +28,6 @@ describe('JSON', function(){
 	it('sidecar SliceTiming shouldnot be greator than RepetitionTime', function(){
 		var jsonObjInval = '{"RepetitionTime": 1.2, "echo_time": 0.005, "flip_angle": 90, "TaskName": "Rest", "SliceTiming":[0.0, 1.7759, 0.0517, 0.8276, 0.1034, 1.8793, 0.1552]}';
 		validate.JSON(file, jsonObjInval, function (issues) {
-			console.log(issues.lenth);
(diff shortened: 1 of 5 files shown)


suyashdb:
@chrisfilo: Ready for review.

utils/issues/list.js
```diff
@@ -326,5 +326,10 @@ module.exports = {
         key: 'SESSION_LABEL_IN_FILENAME_DOESNOT_MATCH_DIRECTORY',
         severity: 'error',
         reason: 'Session label in the filename doesn\'t match with the path of the file. File seems to be saved in incorrect session directory.'
+    },
+    66: {
+        key: 'SLICETIMING_VALUES_GREATOR_THAN_REPETITION_TIME',
+        severity: 'error',
+        reason: 'SliceTiming value/s in json contains invalid value as it is greator than RepetitionTime.'
```
chrisgorgo: Put the names of the fields in quotes (consistent with other error descriptions) and don't mention the JSON file.

tests/json.spec.js
```diff
@@ -25,4 +25,12 @@ describe('JSON', function(){
 		});
 	});
 
+	it('sidecar SliceTiming shouldnot be greator than RepetitionTime', function(){
```
chrisgorgo: should not

tests/json.spec.js
```diff
@@ -25,4 +25,12 @@ describe('JSON', function(){
...
+		var jsonObjInval = '{"RepetitionTime": 1.2, "echo_time": 0.005, "flip_angle": 90, "TaskName": "Rest", "SliceTiming":[0.0, 1.7759, 0.0517, 0.8276, 0.1034, 1.8793, 0.1552]}';
+		validate.JSON(file, jsonObjInval, function (issues) {
+			assert(issues.length === 1 && issues);
```
chrisgorgo: Please verify that the expected error was raised (via the error code) - not just any issue.

validators/json.js
```diff
@@ -60,6 +60,22 @@ function checkUnits (file, sidecar) {
             code: 2
         }));
     }
+    if (sidecar.hasOwnProperty('RepetitionTime') && sidecar.hasOwnProperty('SliceTiming')) {
```
chrisgorgo: This needs to be done on the level of merged dictionary not individual JSON file. RepetitionTime and SliceTiming could be defined in different files.

validators/json.js
```diff
@@ -60,6 +60,22 @@ function checkUnits (file, sidecar) {
...
+        var SliceTimingArray = sidecar["SliceTiming"];
+        var invalid_timesArray = [];
+        SliceTimingArray.filter(function(time){
+            if (time > sidecar['RepetitionTime']){
+                invalid_timesArray.push(time);
+            }
+        });
```
chrisgorgo: Why not do a for loop that goes over all values in SliceTiming and checks each against RepetitionTime. The look could break at the first incorrect value.

utils/issues/list.js
chrisgorgo: The error should also mention that SliceTiming values should be in seconds not milliseconds (common mistake).

chrisgorgo (changes requested):

tests/nii.spec.js
```diff
@@ -135,4 +135,37 @@ describe('NIFTI', function(){
...
+        var file_new = {
+            name: 'sub-15_task-mixedeventrelatedprobe_run-01_bold.nii.gz',
+            relativePath: '/sub-15/func/sub-15_task-mixedeventrelatedprobe_run-01_bold.nii.gz'
+        };
+        var events = [
+            '/sub-15/func/sub-14_task-mixedeventrelatedprobe_run-01_events.tsv',
+            '/sub-15/run-01_events.tsv'
+        ];
+        validate.NIFTI(null, file_new, jsonContentsDict_new, {}, [], events, function (issues) {
+            assert(issues[2].code === 66);
```
chrisgorgo: could you also add assertions for number of errors and warnings?

chrisgorgo (commented):
Thanks! Much better!
Code climat seems to be unhappy about nested ifs https://codeclimate.com/github/INCF/bids-validator/pull/352#issues. Here are some tips how to improve this. https://en.wikibooks.org/wiki/Computer_Programming/Coding_Style/Minimize_nesting

tests/nii.spec.js
suyashdb: Sure will do. 
Yes i am trying to resolve the code climate errors. :)

suyashdb:
@chrisfilo 
CircelCi and Code Climate both looks happy this time!

chrisgorgo (approved):

PR #361, opened 2017-11-01
check filemaps fields only when fmap files are present

suyashdb:


validators/nii.js
```diff
@@ -95,7 +95,7 @@ module.exports = function NIFTI (header, file, jsonContentsDict, bContentsDict,
         }
 
         // field map checks
-        if (path.includes("_bold.nii") || path.includes("_sbref.nii") || path.includes("_dwi.nii")) {
+        if ((path.includes("_bold.nii") || path.includes("_sbref.nii") || path.includes("_dwi.nii")) && (path.includes("fmap"))) {
```
chrisgorgo, line 98: this statement will never be true because _bold, _sbref, and _dwi are not stored in fmap folder

chrisgorgo (changes requested):

chrisgorgo:
Replaced by #363

PR #363, opened 2017-11-03
filter fieldmap issues even if they are errors

### LINKED ISSUE #362: fmap related json fields are checked when fmap files dont exist
Currently,  the implementation of the nii.js is such that, error codes 6,7,8, 9 are generated for all niftis. Latter these issues are filtered out when `fmap/` doesnt exist for that subject. 

However, when we raise these warnings to errors using config file, these errors doesn't get filtered in case the dataset does not contain fmap.

To solve this issue, may be it's a better option [not to generate](https://github.com/INCF/bids-validator/blob/master/validators/nii.js#L98) those warnings/error in cases where fmap doesn't exist.

chrisgorgo:
Closes #362

### DIFF
--- a/utils/issues/index.js
+++ b/utils/issues/index.js
@@ -24,19 +24,16 @@ var issues = {
      * Remove fieldmap related warnings if no fieldmaps
      * are present.
      */
-    filterFieldMaps: function (issueList, summary) {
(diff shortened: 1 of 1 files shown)


chrisgorgo:
Waiting for your review @suyashdb.

suyashdb:
Looks good to me.

Issue #366, opened 2017-11-07
bids-validator checking derivatives


jokedurnez:
Hey,

Is there a reason why the bids-validator checks the derivatives?
First I had put my preprocessed data in there, which results in the bids-validator running for >2h.
Now I'm running preprocessing, and I get this error somewhere between different steps of my pipeline.
This is part of a bids-app (HCPPipelines), so I don't think I can add an ignore flag for the derivatives.

```
bids-validator /oak/stanford/groups/russpold/data/Psychosis/0.0.2//
        1: Subject label in the filename doesn't match with the path of the file. File seems to be saved in incorrect subject directory. (code: 64 - SUBJECT_LABEL_IN_FILENAME_DOESNOT_MATCH_DIRE
CTORY)

                ./derivatives/HCPPipelines/sub-S0784SQF/MNINonLinear/Native/sub-S0784SQF.ArealDistortion_FS.native.dscalar.nii

                        Evidence: File: /derivatives/HCPPipelines/sub-S0784SQF/MNINonLinear/Native/sub-S0784SQF.ArealDistortion_FS.native.dscalar.nii is saved in incorrect subject directory as

per sub-id in filename.
```

chrisgorgo:
No - this is a bug, #346 should fix it. That branch should also have much better performance.

chrisgorgo:
Fixed by #368

Issue #348, opened 2017-10-23
Ignore files and folders starting with '.' in 


chrisgorgo:
It will reduce the noise in the NOT_INCLUDED warning.

@yarikoptic and @satra - what do you think?

Related to #293 and https://github.com/nipy/heudiconv/issues/106

satra:
@chrisfilo - isn't @mgxd suggestion of generating a bids config file sufficient to handle this?

chrisgorgo:
Yes it would, but it requires directing heudiconv users to perform a custom step which is not ideal. I would like the whole process to be as straightforward as possible for users. 

There are also other '.' files such as .DS_store on mac.

Do you see a substantial downside of ignoring '.' files in the validator?

satra:
@chrisfilo - it wouldn't require heudiconv users to do a custom step. whenever they use `-b`, a bids config file will be created with the hidden files ignored. i'm assuming this gets used automatically by the validator.

regarding hidden files, perhaps bids will use hidden files in the future, who knows !?!

to generalize this a bit more, one can consider that we use a `.bidsignore` file, but i see the config option as being sufficient for now.

chrisgorgo:
The config file is provided as command line option - it is not read automatically from the validated dataset (the config files were meant to be reused across datasets - for example by a repository that has specific, more strict metadata requirements). This means users will have to provide this command line option and point to the config file generated by heudiconv. It is also not possible to provide config files for the online version of the validator.

If BIDS will start using '.' files in the future we can modify the validator.

The concept of a `.bidsignore` file is definitely appealing and worth implementing at some point. However, changing the defaults might improve the user experience (especially for people with less expertise).

chrisgorgo:
Actually in theory this test should already ignore '.git' subdirectories.
https://github.com/INCF/bids-validator/blob/c82f0064ad95c8fee5a3f42776d590444acbcf48/utils/type.js#L63

called in:

https://github.com/INCF/bids-validator/blob/c82f0064ad95c8fee5a3f42776d590444acbcf48/validators/bids.js#L219

yarikoptic:
that one seems to be ignoring `.git` only on the top level. (what about any subdirectory being a git submodule?)
`.bidsignore` indeed sounds handy... what syntax would it be? ;)

chrisgorgo:
It ignores `/.git` and all of its subdirectories. We would have to add some thing for submodules (again I think ignoring all `.` and their subdirectories at any level would be most user friendly)

[...]

PR #368, opened 2017-11-13
Relative path and directory traversal optimization

chrisgorgo:
- harmonizes how relativePath is used across browser and command line
- optimizes path traversal to exclude folders we don't need to look into (as early as possible)

Closes #346, #348, and #366.

### DIFF
--- a/utils/files.js
+++ b/utils/files.js
@@ -65,6 +65,19 @@ function readFile (file, callback) {
     }
 }
 
+function harmonizeAndFilter(dir) {
+    var filesList = [];
+    for (var i = 0; i < dir.length; i++) {
+        var fileObj = dir[i];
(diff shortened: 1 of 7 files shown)


utils/files.js
```diff
@@ -65,6 +65,19 @@ function readFile (file, callback) {
     }
 }
 
+function harmonizeAndFilter(dir) {
```
suyashdb: This is called to harmonize dir paths(dir list) in case of browser only. 
Node side file path remain same. Right?

Just for future reference, can we add a small blurb explaining this function.

suyashdb:
It works well when i tested in two datasets using the browser validator. 
Though would like to understand better during our weekly call.

PR #372, opened 2017-12-05
No bids specific nifti files in the dataset

suyashdb:
Closes #325

PR #375, opened 2017-12-10
Added check for no compatible data.

### LINKED ISSUE #325: Raise an error if there are no subject specific BIDS compatible files
Here's a dataset that technically passes the validator:
image

But the dataset is defunct due to typos:
image

It would be useful that in such cases (where there are no BIDS compatible files that are specific to any given participant) the validator returned an error.

chrisgorgo:
Closes #325 and #372


chrisgorgo:
image

PR #381, opened 2017-12-22
Implementation of .bidsignore

chrisgorgo:
Checks for presence of /.bidsignore and reads the exclusion patters from it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,11 @@ If you would like to test individual files you can use the file specific checks 
 
 Additionally you can reformat stored errors against a new config using `validate.reformat()`
 
+#### .bidsignore
+Optionally one can include a `.bidsignore` file in the root of the dataset. This file lists patterns (compatible
+with the [.gitignore syntax](https://git-scm.com/docs/gitignore)) defining files that should be ignored by the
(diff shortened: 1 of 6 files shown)


chrisgorgo:
@yarikoptic, @satra  - would love to hear what do you think

yarikoptic:
Well, I think it is/will be great - exactly what we need! ;)
I forgot though already (if ever knew) how to "try" anything from the land of node

chrisgorgo:
Try this out: https://github.com/INCF/bids-validator#development

yarikoptic:
I guess nmp install installs it system or user wide... Don't want to overload default one... How to create some kind of sandbox like virtualenv in python?

chrisgorgo:
Unless you pass `-g` `npm install` performs a local install in the current directory.

yarikoptic:
Ah, great to know, thanks. I Will try when get back to the laptop and if would still be able to type

yarikoptic:
FWIW -- seems to work lovely for me (thanks!).  Although with the already implemented `ignore all .files` my usecases are largely addressed without requiring .bidsignore ;)

chrisgorgo:
Awesome! Thanks for testing!

PR #391, opened 2018-01-24
enh: use `node:8.9.4-alpine` base image

kaczmarj:
This pull request proposes using `node:8.9.4-alpine` as the base image. This has the benefit of removing a `RUN` layer and minimizing the footprint of the Docker image (83.8MB from 332MB, both uncompressed sizes).


chrisgorgo:
Lean and mean! Thanks!

Issue #347, opened 2017-10-23
Distinguish ERRORS and WARNINGS with text in addition to color(command line version only)


chrisgorgo:
Right now if a dataset only generates warnings users can get the impression that they are errors since nothing other than yellow color on the console indicates that they are only errors. We should explicitly write if something is a warning or an error.

SetCodesToFire:
Hi @chrisfilo , 
PR added for this issue. Added ERR for printing errors and WARN for Warnings

-R

SetCodesToFire:
@chrisfilo  can you tell me what label does config issues should be attached with. I did them with the ERR only

chrisgorgo:
See #421

PR #415, opened 2018-02-27
Added distinguish errors and warnings as text ERR and WARN

SetCodesToFire:
Distinguish ERRORS and WARNINGS with text in addition to color(command line version only) #347


chrisgorgo:
Looks great! Thanks!

SetCodesToFire:
😄

Issue #392, opened 2018-01-30
Error when _events.tsv files have only two columns


chrisgorgo:
Rare but valid `_events.tsv` files with only two columns
```
onset	duration
1	3
2	3
```
lead to an error

```
        1: Second column of the events file must be named 'duration' (code: 21 - EVENTS_COLUMN_DURATION)
                ./sub-CTS202/func/sub-CTS202_task-restbaseline_events.tsv
                        @ line: 1 character: 6
                        Evidence: onset,duration
```

SetCodesToFire:
Hi @chrisfilo ,
I tried to replicate this issue and I am getting another error which is shown here 

![tsv](https://user-images.githubusercontent.com/25459509/36594073-ad036cf0-18c2-11e8-9b7b-602fab02b8ec.PNG)

I tried to debug this but found nothing. Can you identify the mistake?

SetCodesToFire:
where as trying to do the same function call for BIDS, NIFTI, JSON are working

PR #417, opened 2018-02-27
Bug fix for when only two columns are there in valid events.tsv as well as test for it

SetCodesToFire:
Solving #392

### DIFF
--- a/tests/tsv.spec.js
+++ b/tests/tsv.spec.js
@@ -42,7 +42,7 @@ describe('TSV', function(){
     };
 
     it('should require events files to have "onset" as first header', function () {
-        var tsv = 'header-one\tduration\t4eader-three\n' +
+        var tsv = 'header-one\tduration\theader-three\n' +
                   'value-one\tvalue-two\tvalue-three';
         validate.TSV.TSV(eventsFile, tsv, [], function (issues) {
             assert(issues.length === 1 && issues[0].code === 20);
@@ -50,7 +50,7 @@ describe('TSV', function(){
     });
 
     it('should require events files to have "duration" as second header', function () {
-        var tsv = 'onset\theader-two\t4eader-three\n' +
+        var tsv = 'onset\theader-two\theader-three\n' +
                   'value-one\tvalue-two\tvalue-three';
         validate.TSV.TSV(eventsFile, tsv, [], function (issues) {
             assert(issues.length === 1 && issues[0].code === 21);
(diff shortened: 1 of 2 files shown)


tests/tsv.spec.js
```diff
@@ -42,29 +42,37 @@ describe('TSV', function(){
     };
 
     it('should require events files to have "onset" as first header', function () {
-        var tsv = 'header-one\tduration\t4eader-three\n' +
```
chrisgorgo, line 45: `t4eader`?

validators/tsv.js
```diff
@@ -85,7 +85,7 @@ var TSV = function TSV (file, contents, fileList, callback) {
                 code: 20
             }));
         }
-        if (headers[1] !== "duration"){
+        if (!(headers[1] == "duration" || headers[1] == "duration\\n")){
```
chrisgorgo: I'm unclear that this line changes anything. Could you show that your new test fails without the change in line 88?

SetCodesToFire:
Yes, It will fail. Because when there are only two columns in the file then while splitting them using \t as the delimiter will create 
col-1 col-2\n

and hence checking col-2 to be duration will throw an error , while the events.tsv file looks correct to the user.

validators/tsv.js
SetCodesToFire: Yes, It will fail. Because when there are only two columns in the file then while splitting them using \t as the delimiter will create
col-1 col-2\n

and hence checking col-2 to be duration will throw an error , while the events.tsv file looks correct to the user.

tests/tsv.spec.js
SetCodesToFire: this is just a typo change. Do not consider it relevant.

validators/tsv.js
chrisgorgo: Ok so just replace `headers[1]` with `headers[1].trim()`. IT will  simplify the logic.

tests/tsv.spec.js
chrisgorgo: Could you revert it then?

validators/tsv.js
SetCodesToFire: Okay 😄

tests/tsv.spec.js
SetCodesToFire: I don't think that it would anyways create any problem. Because the programmer of that line must be typing header only. And 4eader is not done purposefully.
I think we can keep it either way. There is no issue.

chrisgorgo: ok

validators/tsv.js
SetCodesToFire: .trim() will just remove the extra white-space but when we parse the content of the .tsv file it is in the format of x\ty\tz\na\tb\tc
x    y    z
a    b    c

 hence the trim won't help here

SetCodesToFire: a

SetCodesToFire: b

SetCodesToFire: b

chrisgorgo: You had a double negative in your `trim()` example.

SetCodesToFire: Oops. Changed it.

chrisgorgo:
Thanks!

PR #428, opened 2018-03-05
Fix incorrect dataset summary

olgn:
Ignored files (/.bidsignore, /derivatives, /sourcedata, /code, .*, etc) are now included in the summary stats. 

Fixes a bug referenced [here](https://github.com/OpenNeuroOrg/openneuro/issues/391) where ignored files were causing an inaccurate estimation of dataset size, as well as an incorrect count of files in the dataset.

### DIFF
--- a/utils/files.js
+++ b/utils/files.js
@@ -163,7 +163,7 @@ function preprocessBrowser(filesObj, ig) {
         var fileObj = filesObj[i];
         fileObj.relativePath = harmonizeRelativePath(fileObj.webkitRelativePath);
         if (ig.ignores(path.relative('/', fileObj.relativePath))) {
-            continue;
+            fileObj.ignore = true;
         }
         filesList.push(fileObj);
     }
@@ -193,9 +193,7 @@ function getFiles(dir, files_, rootpath, ig){
         var fullPath = dir + '/' + files[i];
         var relativePath = fullPath.replace(rootpath, '');
         relativePath = harmonizeRelativePath(relativePath);
-        if (ig.ignores(path.relative('/', relativePath))) {
-            continue;
-        }
+        
         var fileName = files[i];
 
         var fileObj = {
@@ -203,6 +201,10 @@ function getFiles(dir, files_, rootpath, ig){
             path: fullPath,
             relativePath: relativePath
         };
+
+        if (ig.ignores(path.relative('/', relativePath))) {
+            fileObj.ignore = true;
+        }
 
         if (fs.lstatSync(fullPath).isDirectory()) {
             getFiles(fullPath, files_, rootpath, ig);
--- a/validators/bids.js
+++ b/validators/bids.js
@@ -216,8 +216,14 @@ BIDS = {
                 hasDatasetDescription = true;
             }
 
+            // ignore files flagged by utils.files.getBIDSIgnore()
+            if (file.ignore) {
+                console.log('ignored file: ', file)
+                process.nextTick(cb);
+            }
+
             // ignore associated data
-            if (utils.type.isStimuliData(file.relativePath)) {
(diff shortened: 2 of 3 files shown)


utils/files.js
```diff
@@ -163,7 +163,7 @@ function preprocessBrowser(filesObj, ig) {
         var fileObj = filesObj[i];
         fileObj.relativePath = harmonizeRelativePath(fileObj.webkitRelativePath);
         if (ig.ignores(path.relative('/', fileObj.relativePath))) {
-            continue;
+            fileObj.ignore = true;
```
chrisgorgo, line 166: This was introduced to speed up validation of datasets with thousands of files in /derivatives folders. See https://github.com/INCF/bids-validator/issues/344. I might be worth checking if this PR does not reintroduce the performance issues.

olgn: @chrisfilo is it the number of files, or the total size of the folder that effects this? could you point me to a dataset that you could imagine being problematic?

chrisgorgo: It's unclear if it is the sheer number of files or their size - this would have to be tested. The dataset mentioned in the original issue could be found here: https://openfmri.org/dataset/ds000030/

olgn: gotcha. I am downloading some pieces of that dataset now, and will test this new implementation out. do you by chance have any information as to the operating system of the user who had trouble uploading originally?

chrisgorgo: I believe it was Linux via Node.

olgn: running validation on 1/10 of that dataset, (~60 gigs) takes five seconds on my mac via node. will check with linux (and a slightly larger chunk of the dataset) tomorrow and get back to you.

olgn: here is the output on a larger portion of the dataset using the following code (test.js):

```
const validator = require('/Users/teal/Git/bids-validator');
const filesize = require('filesize');

const derivatives_folder = '/Users/teal/Documents/openneuro/datasets/derivative_stuff';
console.log('---------- BIDS VALIDATOR TEST STARTING  ----------');
console.log('...');
console.time('validation duration');

validator.BIDS(derivatives_folder, {}, function(errors, stats) {
        console.log('---------- BIDS VALIDATOR TEST COMPLETE ----------');
        console.log('directory:', derivatives_folder);
        console.log('directory size:', filesize(stats.size));
        console.timeEnd('validation duration');
});
```

screen shot 2018-03-06 at 12 24 47 pm

chrisgorgo: LGTM.

(now I got this stuck in my head: https://www.youtube.com/watch?v=wqCpjFMvz-k)

validators/bids.js
```diff
@@ -216,8 +216,13 @@ BIDS = {
                 hasDatasetDescription = true;
             }
 
+            // ignore files flagged by utils.files.getBIDSIgnore()
+            if (file.ignore) {
+                process.nextTick(cb);
+            }
```
chrisgorgo: It might be worth moving this to the top of the eachOfLimit loop to skip a couple extra string comparisons.

chrisgorgo (approved):

validators/bids.js
olgn: good point. taken care of with [this commit](https://github.com/INCF/bids-validator/pull/428/commits/53d7d495653eef49b604e7c7586b7c9dd3278baf)

utils/files.js
olgn: pretty good tune, i must say

validators/bids.js
olgn: @chris perhaps i was a little quick with this - upon inspection, moving this to the top caused process.nextTick() to be called twice - and therefore failed to perform properly. reverted in 9f87fd1

PR #440, opened 2018-03-15
Add global install command to readme

choldgraf:
Makes it clearer to install the validator globally, in case you need to run it on other folders (e.g. as part of a testing pipeline for another package)


chrisgorgo:
Thanks!

PR #441, opened 2018-03-19
Fixed linking issues and removed redundant steps

wazeerzulfikar:
Fixed linking issues and removed redundant steps in the instructions for local development in the README file


chrisgorgo:
Thanks! 

BTW - the web version of the validator is also being built automatically on CircleCI for each commit/PR.

PR #447, opened 2018-03-29
file size-check to top of validator; remove ignored files from fileslist

olgn:
* the size of the dataset is calculated at the beginning of the bids test
* after getting the size, unused files are deleted from the filesList


chrisgorgo:
Thanks!

PR #451, opened 2018-04-04
[WIP] iEEG validator extension

choldgraf:
This PR adds validation for the iEEG specification. It's still WIP and is basically a riff off of @jasmainak 's MEG PR. Still a WIP but comments are definitely welcome.

**NOTE: the following list is now in  https://github.com/INCF/bids-validator/issues/510 **
## To-do list for this PR to get merged

- [ ] Agree on how to reference the coordinate systems in the main BIDS spec
- [ ] Agree on nomenclature around "electrodes" vs "contacts" (e.g. for electrode shanks vs single electrodes)
- [ ] Decide on a final list of file extensions to support.
  - Current extensions suggested: EDF, VHDR
  - Current extensions allowed in addition: MEF3, SET, NWB
  - In particular @jasmainak and @teonbrooks showed concerns with having EDF as a primary format we recommend people use.
- [ ] Decide whether we want javascript readers to be a blocker on the specification
- [ ] Decide whether we want python readers to be a blocker on the specification
- [ ] Figure out prefixes for units
- [ ] Relatively minor spot-checks in the google doc
- [ ] Final pass
- [ ] Merge!

<details>
<summary>Old comments from before</summary>

A few thoughts that came up with this first pass:

* Here: do we need `_ieeg` and `_meg` for each of the electrodes/channels files (e.g., should we do `channels.tsv` -> `channels_meg.tsv`? Here are some points where this might be relevant:
  * https://github.com/INCF/bids-validator/pull/301/files#diff-bca076162864b02c2dbe628ed44a653cR61
  * https://github.com/INCF/bids-validator/pull/301/files#diff-bca076162864b02c2dbe628ed44a653cR108
  * https://github.com/INCF/bids-validator/pull/301/files#diff-bca076162864b02c2dbe628ed44a653cR196
  * https://github.com/INCF/bids-validator/pull/301/files#diff-bb569ec823383674890ed051c0aab5f5R42
* Or can we just update the coordsystem schema to include ieeg?
  * https://github.com/INCF/bids-validator/pull/301/files#diff-edddc921ee448ec0b00eb10faf4ac9d6R3
* Does it make sense to merge some of the JSONSchemas into a "shared" schema and
  then use the same "inheritance" principle so we only define the schema items
  that differ for each type of data? At the least we could have a `base.json`
  that holds fields to be shared between all data
* Does the MEG spec have "electrodes.tsv" or is this just iEEG? I didn't see
  anything about `electrodes.tsv` in Mainak's PR.
* Where is the check for optional fields in TSV files? E.g., can we create a
  JSONSchema for TSV files where keys map onto columns, and "type" maps on to the
  type for each column?
</details>

cc @chrisfilo @dorahermes @tyarkoni @jasmainak who may be interested or have thoughts


tests/json.spec.js
```diff
@@ -32,23 +32,43 @@ describe('JSON', function(){
...
-        validate.JSON(meg_file, jsonObjInval, function(issues){
-            assert(issues && issues.length === 1);
-        });
-    });
+  var meg_file = {
+      name: 'sub-01_run-01_meg.json',
+      relativePath: '/sub-01_run-01_meg.json'
+  };
+
+  it('*_meg.json sidecars should have required key/value pairs', function(){
```
choldgraf, line 40: the meg `it(` section was at a different indentation level and I couldn't figure out why, so I aligned them

dorahermes:
Really cool that you made a start! I can only comment on one of your points, which is that the MEG spec does not have an _electrodes.tsv file, but a digitization file: _headshape.<manufacturer specific extension>.

tests/json.spec.js
jasmainak: It's javascript, indentation doesn't matter :-P

validators/json.js
```diff
@@ -39,6 +39,8 @@ function checkUnits (file, sidecar) {
             schema = require('./schemas/dataset_description.json');
         } else if (file.name.endsWith("meg.json")) {
             schema = require('./schemas/meg.json');
+        } else if (file.name.endsWith("ieeg.json")) {
+            schema = require('./schemas/ieeg.json');
```
jasmainak, line 45: you need to update here.

coordsystem.json -> coordsystem_meg.json etc.

utils/issues/list.js
```diff
@@ -372,4 +372,24 @@ module.exports = {
...
     }
+    75: {
+        key: 'CHANNELS_COLUMN_SFREQ',
+        severity: 'error',
+        reason:   "Fourth column of the channels file must be named 'sampling_frequency'"
+    },
+    76: {
+        key: 'CHANNELS_COLUMN_LOWCUT',
+        severity: 'error',
+        reason:   "Third column of the channels file must be named 'low_cutoff'"
```
jasmainak: Fifth?

utils/issues/list.js
```diff
@@ -372,4 +372,24 @@ module.exports = {
...
+    },
+    77: {
+        key: 'CHANNELS_COLUMN_HIGHCUT',
+        severity: 'error',
+        reason:   "Third column of the channels file must be named 'high_cutoff'"
```
jasmainak: Sixth?

utils/issues/list.js
```diff
@@ -372,4 +372,24 @@ module.exports = {
...
+    },
+    78: {
+        key: 'CHANNELS_COLUMN_NOTCH',
+        severity: 'error',
+        reason:   "Third column of the channels file must be named 'notch'"
```
jasmainak: Seventh? :-P

jasmainak:
> Does it make sense to merge some of the JSONSchemas into a "shared" schema and
then use the same "inheritance" principle so we only define the schema items
that differ for each type of data? At the least we could have a base.json
that holds fields to be shared between all data

Wouldn't this be sorted out when the iEEG spec gets merged into the main spec? So, I suppose nothing to worry about right now

jasmainak:
> do we need _ieeg and _meg for each of the electrodes/channels files (e.g., should we do channels.tsv -> channels_meg.tsv?

again, seems like a discussion for the google doc of iEEG spec, no?

utils/issues/list.js
```diff
@@ -372,4 +372,24 @@ module.exports = {
         severity: 'error',
         reason:   "Nifti file exist with both '.nii' and '.nii.gz' extensions."
     }
+    75: {
```
jasmainak: I wonder if we should pool these codes together into one: MISSING_TSV_COLUMN ? What do you think? There seems to be some redundancy there

jasmainak:
looks like a good start indeed !

you need to update tests

jasmainak:
> Where is the check for optional fields in TSV files? E.g., can we create a
JSONSchema for TSV files where keys map onto columns, and "type" maps on to the
type for each column?

There is none at the moment. That would be a welcome addition!

tests/json.spec.js
choldgraf: it may not matter to the machine, but to some humans it does ;-)

validators/json.js
choldgraf: I wanted to make sure this was the path forward we all agreed on before implementing all of these changes. Do you think it's a good decision to require `coordsystem.json` to be instead `coordsystem.{{ieeg | meg | eeg}}.json`? What do you think @chrisfilo ?

utils/issues/list.js
choldgraf: Seems reasonable to me, I guess there could be one error, and it'd say "The first N columns must be these values: `[XXX, ..., XXX]`" yeah?

validators/json.js
chrisgorgo: Are you referring to the name of the schema file? As long as content is correct and the right file is being called in the code it does not matter much.

choldgraf: yep, just the name. Just a question of whether to force people to use `_<insert-modality-type>` at the end of `coordsystem`

chrisgorgo: I am still a bit confused - which people? bids-derivatives developers or BIDS iEEG users?

In other words is this a question about internal bids-derivatives implementation detail or the shape of the BIDS iEEG proposal?

choldgraf: The question is: should the BIDS-iEEG specification require that the file be called `coordsystem_ieeg.json`, and that the BIDS-MEG specification should be called `coordsystem_meg.json`, rather than allowing each one to call the file `coordsystem.json`

choldgraf: The way I see it, these are the two options:

1. Both specifications can have a file called `coordsystem.json`.
    * We then need to change the regex checkers to look not *just* for a file called `coordsystem.json`, since that will be ambiguous as to whether it's for `meg`/`ieeg`/etc.
    * Instead, we need to look for `ieeg/coordsystem.json` or `meg/coordsystem.json`.
2. The two specifications must change their coordsystem json file to include the spec name in the file name (e.g. `coordsystem_ieeg.json` and `coordsystem_meg.json`).
    * In this case, we need to go through the regexes and have them look for these specification-specific file names.

utils/issues/list.js
```diff
@@ -353,23 +353,14 @@ module.exports = {
...
-    72: {
-        key: 'CHANNELS_COLUMN_TYPE',
-        severity: 'error',
-        reason:   "Second column of the channels file must be named 'type'"
-    },
-    73: {
-        key: 'CHANNELS_COLUMN_UNITS',
-        severity: 'error',
-        reason:   "Third column of the channels file must be named 'units'"
+        reason:   "The column names of the channels file must begin with ['name', 'type', 'units', 'sampling_frequency', 'low_cutoff', 'high_cutoff', 'notch']"
```
jasmainak: channels file -> channels.tsv file

Is it the same for MEG channels.tsv?

jasmainak:
also need tests for the regexps

validators/json.js
chrisgorgo: Few comments:

1. Even though at this stage it is still possible to change the MEG part of spec it would be quite difficult (as you know the MEG crowd would like to get it over with and submit the revision of the paper) and not worth it for this change.

2. You have to be more inclusive when discussing changes to the spec (in contrast to validator implementation details). Especially when they involve multiple modalities.

3. Related to above - this might not be the best forum to discuss spec details. This information gets lost in flurry of other emails concerning validator implementation details and very few people monitor this. I would leave such discussions to the mailing list.

choldgraf:
What's a good way to debug the validator? E.g. I'm getting a cryptic error message and I _think_ it's tied to a specific schema that's being validated, but I don't get any information about which file this is (here's the message:

```
Error: schema is invalid: data.required should NOT have duplicate items (items ## 2 and 11 are identical)
    at Ajv.validateSchema (/mnt/c/Users/chold/Dropbox/github/forks/python/neuro/bids-validator/node_modules/ajv/lib/ajv.js:188:16)
    at Ajv._addSchema (/mnt/c/Users/chold/Dropbox/github/forks/python/neuro/bids-validator/node_modules/ajv/lib/ajv.js:317:10)
    at Ajv.compile (/mnt/c/Users/chold/Dropbox/github/forks/python/neuro/bids-validator/node_modules/ajv/lib/ajv.js:117:24)
    at checkUnits (/mnt/c/Users/chold/Dropbox/github/forks/python/neuro/bids-validator/validators/json.js:50:32)
    at /mnt/c/Users/chold/Dropbox/github/forks/python/neuro/bids-validator/validators/json.js:22:44
    at Object.parse (/mnt/c/Users/chold/Dropbox/github/forks/python/neuro/bids-validator/utils/json.js:29:17)
    at module.exports (/mnt/c/Users/chold/Dropbox/github/forks/python/neuro/bids-validator/validators/json.js:20:16)
    at /mnt/c/Users/chold/Dropbox/github/forks/python/neuro/bids-validator/validators/bids.js:334:21
    at /mnt/c/Users/chold/Dropbox/github/forks/python/neuro/bids-validator/utils/files.js:51:47
    at process._tickCallback (internal/process/next_tick.js:112:11)
```

Is there a way to get `--verbose` to say which file its validating as it scans through a directory?  @jasmainak what helped you in these little nitpicky things?

chrisgorgo:
I use the graphical debugger included in WebStorm and set some breakpoints.

Best,
Chris

jasmainak:
yeah use the webstorm debugger. It's really helpful to spend a few moments to understand how the debugger works :-)

choldgraf:
You guys have any suggestions for learning how to use webstorm? This is the first I've heard of it...

jasmainak:
I learned it by playing around with the menu buttons, et voila :-P

and for the debugger, you can click on the line number just like in any other IDE

choldgraf:
Wow that took me way too long to figure out. So it turns out that the problem is that one of the JSON files didn't have its final line end with `,`. 

E.g., if the final line was:

```
{
    "TaskName": "visual",
   ...
    "iEEGReferenceScheme": "intracranial channel not included with data"
}
```

it would throw the error:

`Error: schema is invalid: data.required should NOT have duplicate items (items ## 2 and 11 are identical)`

but if the final line was:

```
{
    "TaskName": "visual",
   ...
    "iEEGReferenceScheme": "intracranial channel not included with data",
}
```

then it would pass fine.

This took me a long time to figure out because ending the JSON with a `,` actually doesn't pass many JSON linters (e.g. https://jsonlint.com/). Have you guys run into this problem as well?

chrisgorgo:
I did not, but indeed the error message is counter intuitive.

Best,
Chris

PS Apologies for brevity and potential typos. This message was composed on
a phone.

choldgraf:
ahh I figured it out - the `,` actually wasn't the error, but if the file included a `,` then it would trigger a _different_ error and the script would break then. If the `,` was removed, it would pass this initial check and then a later check would raise the _actual_ error.

jasmainak:
@choldgraf why do you use an online linter when it is already in sublime text? ;-) maybe that's why I never ran into the error ...

choldgraf:
I use atom for linting, I had just reached that point of debugging where you think to yourself "I have no idea why this isn't working I'm going to start desperately trying things until it works again"

validators/schemas/ieeg.json
```diff
@@ -0,0 +1,47 @@
+{
+        "type": "object",
+        "properties": {
```
choldgraf, line 3: Not sure whether we need to explicitly define all of the non-ieeg channel counts since MEG didn't do this, but this is what we have in the specification for ieeg so ¯\_(ツ)_/¯

choldgraf:
ok - I think the ieeg validator is nearly there (at least for this version of the ieeg specification). A couple dangling questions:

1. What's the process around adding ieeg datasets to the `bids-examples` repository? It seems like we need to add them and then make the tests in this PR pull the archive from the ieeg branch?
2. @dorahermes and I should make sure that the datasets on the BIDS-iEEG branch both adhere to the current specification and pass the validator (@jasmainak what would you recommend for this? Perhaps Dora can just checkout my version of the validator so she can spot-check problems on her data locally?)
3. Once that's done I can build in the new bids-examples tests into the validator tests for this branch

chrisgorgo:
This is the current BIDS IEEG examples branch: https://github.com/INCF/BIDS-examples/tree/bep010_ieeg

For a while it will be an iterative process, examples will be updated and tested against this PR, the PR might need updating, the examples might need updating.

When we will be ready to merge and release BIDS 1.2 (with iEEG) we will 1) merge https://github.com/INCF/BIDS-examples/tree/bep010_ieeg 2) tag new version of bids-examples 3) update this PR to use new version of examples (that will include ieeg).

BTW could you make sure that the validator will correctly detect presence of ieeg modalities and list them in summary (like it does for T1w, bold etc.)? It was missing in the MEG PR.

choldgraf:
@chrisfilo sounds good - re: the summary stuff, where is that code in the validator? And should I add the same for MEG in this PR or has that already been done?

jasmainak:
Feel free to fix for MEG too

chrisgorgo:
It's here: https://github.com/INCF/bids-validator/blob/88a30a31babf6d1f91d2a9d54efab8c150a7c3b8/validators/bids.js#L249

I would recommend doing MEG fix in separate PR so it could be merged sooner.

choldgraf:
hey all - I've gotten this working with the `npm` tests as well as have my own dataset passing. I believe that @dorahermes is working on getting her own dataset to pass as well.

What is the path forward on this PR? Obviously make sure that circle is passing, but beyond that is there a specific to-do item? (also happy to have another review from somebody to make sure I'm not missing something). I'm happy to keep making PRs to spot-check problems as they come up if it looks like we didn't get something quite right this time around...

dorahermes:
I updated by datasets so they should pass, @choldgraf can you check?

chrisgorgo:
If the branch is feature complete we wait for the BEP to be merged then:

- merge the bids-examples branch
- mint a new release of bids-examples
- update this PR to make circle happy

One thing that you might want to add here (that MEG PR was sadly missing) is some form of light validation of the binary data formats. For example for each NIFTI file the validator is reading the header and checking validity and consistency of its metadata.

choldgraf:
@chrisfilo interesting point re: validating the binary files. @jasmainak or @teonbrooks did you guys think up a way that you could validate a `.fif` / `.edf` / `.gdf` file?

choldgraf:
ping to @larsoner as well - Eric do you know of a good way we could do a quick "is this a proper `.fif` / `.gdf` / `.edf`?" check with javascript by looking at the headers of the files?

jasmainak:
I don't know any javascript way to do it unfortunately :\ Currently the MEG specification supports 6 or 7 formats and I am not sure if there is a javascript reader for each of them. At least I could find an edf reader on a quick google search: https://www.npmjs.com/package/edf-parser. Maybe you could cook up something by reading the IO code in mne-python. I don't know ...

choldgraf:
@chrisfilo do you have any suggestions for how we could validate edf/gdf/fif files from javascript? Hopefully nothing too crazy - I can write regexs, but writing header readers for binary data formats is pretty far outside of my javascript / coding skillset...

chrisgorgo:
I don't know these formats, so I cannot help. Would be good if at least
some of them had I/O libraries in JS. I have not looked myself though.

dorahermes:
Does the biosig toolbox run in Java?  
https://github.com/donnchadh/biosig/tree/master/biosig4java

See: http://biosig.sourceforge.net/

teonbrooks:
> @chrisfilo interesting point re: validating the binary files. @jasmainak or @teonbrooks did you guys think up a way that you could validate a .fif / .edf / .gdf file?

we don't have a validation tool for these (yet). are you asking if there is a js implementation for reading header info from files? and then, I guess, comparing that to the sidecar files?

choldgraf:
yeah, mostly. @chrisfilo mentioned that we need to validate that the `.edf` `.fif` etc files are actually what they seem. I don't think we need to really strongly validate them against the JSON data, but at the least we could return that they are a "proper" version of the file type.

That said, I have no idea how to go about reading in these files from scratch in Python much less in Javascript...

teonbrooks:
@chrisfilo, could you point us in the direction of how you validate the nifti file? I could use that as a starting block to look into `fif` for starters for meg, and then `edf` for eeg. 

also, just found [this](https://github.com/NeuroTechX/node-edf) which might also be a good starter

choldgraf:
perhaps we could shoot for EDF + FIF validation as a starting point for the validator (that would at least cover part of MEG and iEEG) and add subsequent validator support as updates in the future

chrisgorgo:
Here you go: https://github.com/INCF/bids-validator/blob/master/utils/files.js#L225

+1 for starting small and adding more file formats later.

Please mind that not all node modules are pure JS - some are just bindings to external libraries.

choldgraf:
I'm worried that this PR is becoming stale because it's currently blocking on a javascript EDF/FIF/etc reader that will likely take some time to implement.

It sounds like the first pass will be to validate `.edf` files for this round, and to add support for other file formats in the coming months. Since EDF cuts across almost all electrophysiology modalities, can we get this PR merged without blocking on an EDF javascript parser? I think that it will be complex enough that it would make this PR significantly more complicated, and is also somewhat orthogonal to the file name / metadata checking that's already in this PR. Getting this merged would also make it easier to spot where we can reduce redundancy with the EEG validator...

WDYT @chrisfilo ?

(happy to make other edits etc per the other parts of the PR, however)

choldgraf:
alternatively, we could wait until the specification is frozen as the primary blocker (as opposed to EDF javascript validation)

tests/bids.spec.js
```diff
@@ -160,4 +160,22 @@ var suite = describe('BIDS example datasets ', function() {
             isdone();
         });
     });
+
+    it('validates ieeg modalities', function(isdone) {
+        var options = {ignoreNiftiHeaders: true};
+        validate.BIDS("tests/data/BIDS-examples-" + test_version + "/ds001", options, function (issues, summary) {
```
choldgraf, line 166: I am thinking that this is cruft from an earlier version of this PR (this whole section)...does that seem true to folks?

chrisgorgo:
I understand your concern, but I would prefer waiting until the iEEG BEP is
frozen and merged into the main spec.

One alternative is to set up a branch for all iEEG related validator fixes
and send PRs against that. This could help organize the work, but will not
avoid having to keep up with the upstream. Luckily upstream is not
undergoing any major refactors.

Best,
Chris

jasmainak:
@choldgraf: 

http://webmeg.net/js/formatReaders/fiffReader.js
http://webmeg.net/js/formatReaders/edfReader.js

I think you might like this :)

I'm also happy to collaborate on writing separate node libraries for these readers if they don't exist :-)

choldgraf:
nice! very cool. I'm going to have minimal time to look into javascript readers/writers (getting married in 2 3 weeks), so I am totally welcome to new PRs or pushes to this PR.

choldgraf:
Hey all - I took a pass through our outstanding comments / issues and tried to collect them into a list at the top, so that it's clearer what needs to be done for this to get merged.

https://github.com/INCF/bids-validator/pull/451#issue-179479214

Take a look and let me know if I missed something / if we should remove or edit something.

In particular I am curious what people think about making Javascript readers or Python readers to be a blocker for the specification.

jasmainak:
> I'm going to have minimal time to look into javascript readers/writers (getting married in 2 3 weeks), so I am totally welcome to new PRs or pushes to this PR.

congratulations @choldgraf ! It's a bummer I'm not in Paris at the moment. It's one of the things I am interested at looking into with @teonbrooks in the upcoming hackathon at Neuroinformatics. I will keep you posted.

choldgraf:
Hey all - I just rebased and pushed a commit that replicated what @sappelhoff did here: https://github.com/robertoostenveld/bids-validator/pull/15

It looks like tests aren't passing but I'm not sure that I'll have time to fix them before leaving for my wedding on Sunday, after which I'll be offline for about a month. If somebody can give me feedback soon I can try to implement something, but I don't have time to disentangle a bunch of javascript :-( 

maybe @jasmainak can help out? anybody feel free to push directly to this branch

nellh:
I got the tests working but there's a couple things I ended having to change in the actual tests:

1. The removed electrodes.tsv test was looking for a type column which isn't present in the example or the spec.
1. The ieeg.json sidecar test example was missing a couple fields required by the JSON schema, SamplingFrequency and iEEGReference. I filled in some dummy values in the test but it might be better to fix the schema to not require these. Let me know and I can make that change instead.

jasmainak:
I think dummy values are fine in the test of the schema! Since these are required fields in the specification, perhaps it's better not to change the schema?

choldgraf:
@nellh wow - you are awesome! Thanks so much :-)

sappelhoff:
@choldgraf : Bringing the discussion back to file format / header validation, I started a validator for the BrainVision (vhdr) format here: https://github.com/sappelhoff/brainvision-validator

We could make good use of that for both BEP006 and BEP010 :-) feel free to contribute

choldgraf:
@sappelhoff ah that looks great! I'm looking forward to getting more brainvision tools in the open source ecosystem. I'll be of limited help in this space for the next few weeks (getting married and then a rare break from work for a few weeks). But I can help out when I get back.

nellh:
I got this up to date with master after merging #464 - if someone can review the conflicts resolution and all else looks ready, I can merge this.

I think we can set aside the Code Climate reports here, a lot of code formatting with change in #504 anyways.

utils/type.js
```diff
@@ -113,9 +116,15 @@ module.exports = {
...
+            '\\/(?:(ses-[a-zA-Z0-9]+)' +
+            '\\/)?\\1(_\\2)?(?:_task-[a-zA-Z0-9]+)?(?:_acq-[a-zA-Z0-9]+)?(?:_proc-[a-zA-Z0-9]+)?(?:_space-[a-zA-Z0-9]+)?'
+            + '(_events.tsv|_channels.tsv|_electrodes.tsv|_ieeg.json|_coordsystem.json|_photo.jpg|_headshape.pos)$');
+
         return conditionalMatch(scansRe, path) || conditionalMatch(funcSesRe, path) ||
             conditionalMatch(anatSesRe, path) || conditionalMatch(dwiSesRe, path) ||
-            conditionalMatch(megSesRe, path) || conditionalMatch(eegSesRe, path);
+            conditionalMatch(megSesRe, path) || conditionalMatch(eegSesRe, path) ||
+            conditionalMatch(megSesRe, path) || conditionalMatch(ieegSesRe, path);
     },
```
sappelhoff: That's one `conditionalMatch(megSesRe, path) ||` too much, should be:

```js
            conditionalMatch(megSesRe, path) || conditionalMatch(eegSesRe, path) ||
            conditionalMatch(ieegSesRe, path);
```

validators/bids.js
```diff
@@ -235,7 +238,9 @@ BIDS = {
 
         // validate individual files
         async.eachOfLimit(fileList, 200, function (file, key, cb) {
-            var path = file.relativePath;
+            const path = file.relativePath;
+            const pathParts = path.split('_');
+            const suffix = pathParts[pathParts.length - 1];
 
             // Make RegExp for detecting modalities from data file extensions
             var dataExtRE = new RegExp (['^.*\\.(',
```
sappelhoff, line 246: the `dataExtRE` should be adjusted for iEEG as well (adding `dat`). E.g.,

```js
            // Make RegExp for detecting modalities from data file extensions
            var dataExtRE = new RegExp (['^.*\\.(',
                                         'nii|nii\\.gz|', // MRI
                                         'fif|fif\\.gz|sqd|con|kdf|chn|trg|raw|raw\\.mhf|', // MEG
                                         'eeg|vhdr|vmrk|edf|cnt|bdf|set|fdt', // EEG
                                         'some|extensions|for|ieeg|that|are|not|yet|covered', //iEEG
                                         ')$'].join(''));
```

... so that iEEG is added as a modality.

validators/json.js
```diff
@@ -40,6 +41,12 @@ function checkUnits (file, sidecar) {
...
+        } else if (file.path.includes('/meg/') && file.name.endsWith("coordsystem.json")) {
+            schema = require('./schemas/coordsystem_meg.json');
+        } else if (file.path.includes('/ieeg/') && file.name.endsWith("coordsystem.json")) {
+            schema = require('./schemas/coordsystem_ieeg.json');
         } else if (file.name.endsWith("eeg.json")) {
```
sappelhoff, line 50: Note: there is no `coordsystem_eeg.json` yet ... we should add that ... but after merging this PR

validators/bids.js
```diff
@@ -267,16 +272,27 @@ BIDS = {
...
+            if (summary.modalities.indexOf(modality) === -1) {
+                summary.modalities.push(modality);
+            }
+
+                process.nextTick(cb);
+                }
+
+            // capture ieeg files for summary
+            else if (file.name.endsWith('.edf') || file.name.endsWith('.vhdr') || file.name.endsWith('.vmrk') || file.name.endsWith('.dat')) {
+                ephys.push(file);
```
sappelhoff: the capture of ieeg files for summary should happen on line 275 ... below the capture of nifti files for validation.

Also, if the array is called `ephys`, the EEG files should be captured as well, this involves adding `cnt|bdf|set` as possible extensions using the `.endsWith` style --> perhaps there is a smarter way to program this, apart from using many ORs

sappelhoff (changes requested):

utils/type.js
nellh: Pushed a fix for this. Thanks for catching that.

jasmainak:
has this been validated against the new iEEG dataset [offline](https://github.com/INCF/bids-validator/blob/88a30a31babf6d1f91d2a9d54efab8c150a7c3b8/tests/bids.spec.js). Otherwise, you'll end up with lots of failures in BIDS-examples ...

validators/bids.js
nellh: What extensions would be useful here? It doesn't seem clear from the BEP010 spec.

nellh: I'll add 'dat' but I'm unclear if there's any others.

sappelhoff: @nellh, that's something @choldgraf or @dorahermes could say more about.

dorahermes: In the current spec iEEG also allows the following:
nwb|tdat|tidx|tmet

(I will update the visual and motor BIDS-example sets this week)

jasmainak: @sappelhoff this looks outdated for MEG. We should fix it.

@nellh there is an intersection between the file formats for EEG and iEEG. I think it will pretty hard to tell the modality without reading the files ... you could probably say EEG/iEEG or something like that.

jasmainak: It comes back to the discussion we had with @teonbrooks about having javascript readers for ephys files

nellh: Thanks, I've added those.

sappelhoff: @jasmainak:

> his looks outdated for MEG. We should fix it.

Could you perhaps make a specific issue?

> It comes back to the discussion we had with @teonbrooks about having javascript readers for ephys files

See this [project](https://github.com/sappelhoff/brainvision-validator) I have started for validation of BrainVision files

nellh:
I'm going to merge this so we can continue with refactoring. Further improvements can work towards enabling the `--bep010` flag by default.

choldgraf:
Sounds great! Thanks everybody for pitching in to get this one merged! I've opened up https://github.com/INCF/bids-validator/issues/510 to coordinate the things that are remaining in order to "activate" the iEEG spec in the validator. Just FYI!

PR #472, opened 2018-05-05
Add --json flag to produce machine readable output.

nellh:
This is useful to run the validator from Python (or other environments) and work with the resulting errors, warnings, and summary.

This mode also does not set exit codes, the reader is expected to check for errors or warnings.


chrisgorgo:
0.26.11 landed in npm

PR #487, opened 2018-07-19
MINOR: add remaining optional fields

sappelhoff:
added the optional "DatasetDOI" and "Acknowledgements" fields to `dataset_description.json` as specified in the specification  section [8.1](https://docs.google.com/document/d/1HFUkAEE-pB-angVcYe6pf_-fVf4sCpOHKesUvfb8Grc/edit#heading=h.ie43v2mfgaeq).

I furthermore took the liberty and rearranged the order of fields to match the order as described in the BIDS spec. This makes it easier for developers to go through the list top to bottom using a combination of the specification and the validator code.

### DIFF
--- a/validators/schemas/dataset_description.json
+++ b/validators/schemas/dataset_description.json
@@ -21,7 +21,10 @@
             "items": {
                 "type": "string"
             },
-            "type": "array"
+            "type": [
+                "string",
+                "array"
+            ]
         },
         "HowToAcknowledge": {
             "items": {
--- a/validators/schemas/dataset_description.json
+++ b/validators/schemas/dataset_description.json
@@ -2,39 +2,54 @@
     "$schema": "http://json-schema.org/draft-06/schema#",
     "$id": "http://example.com/example.json",
     "properties": {
+        "Name": {
+            "type": "string"
+        },
+        "BIDSVersion": {
+            "type": "string"
+        },
+        "License": {
+            "type": "string"
+        },
         "Authors": {
             "items": {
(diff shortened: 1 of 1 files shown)


sappelhoff:
btw, am I using the minor tag correctly? What I mean is "it's a minor change"

chrisgorgo:
There is no conventions about "MINOR:" in this repo, but I got what you mean. Thanks for the contribution!

chrisgorgo:
It seems to be failing on ds000117 - could you look into this?

sappelhoff:
yes, they do not have their `Acknowledgements` as array, see [here](https://github.com/INCF/BIDS-examples/blob/master/ds000117/dataset_description.json).

This can be fixed by enclosing their string with square brackets in the dataset `[ ]`.

According to the spec ([link](https://docs.google.com/document/d/1HFUkAEE-pB-angVcYe6pf_-fVf4sCpOHKesUvfb8Grc/edit#heading=h.qza727avafid)), Acknowledgements should be an array ... at least it reads like it.

validators/schemas/dataset_description.json
```diff
@@ -2,39 +2,54 @@
...
         },
-        "BIDSVersion": {
-            "type": "string"
-        },
-        "Funding": {
+        "Acknowledgements": {
             "items": {
                 "type": "string"
             },
-            "type": ["array", "string"]
```
chrisgorgo, line 18: Please leave it as a string or array - this will make the tests pass again.

PR #488, opened 2018-07-21
Added Neurostars message

franklin-feingold:
Added the Neurostars message for all CLI outputs. This is to bring attention to a location that they are able to ask questions. Even if the user has a validated dataset, it may be helpful to still have this link.


chrisgorgo:
Thanks!

Issue #475, opened 2018-05-26
Integrate validation of BrainVision data files


sappelhoff:
Some data are saved in a multi-file system. For example the Brainvision format for EEG data contains three files:

- `.eeg` binary eeg data
- `.vhdr` meta data in text format
- `.vmrk` event markers in text format relating to the `.eeg`

I don't know if there are multi-file systems accepted somewhere in BIDS up to now ... but as soon as BEP006_EEG will be merged, there will be that case (at the latest).

Once the BIDS-validator encounters data of a "multi-file" system, it might make sense to kick in some assertion that all files of the multi-file system are present such as in this pseudo-code (python-like because I don't know much about js):

``` python
# import some list of what multi_file_endings are in BIDS
import multi_file_endings

# Check if current file ending is part of multi_file_endings
if file_ending in multi_file_endings:
   
    # Check which other files should be present
    # e.g., if file_ending=='.vhdr', then 
    # other_file_endings = ['.eeg', '.vmrk'] 
    other_file_endings = find_multi_file_system(f_ending, multi_file_endings)

    # Assert all necessary files are present ... else, raise some error message
    for other_file_ending in other_file_endings:
        assert other_file_ending in dir_that_is_currently_checked
```

This might be relevant in general and not just for EEG, so I am posting it as an issue here.

sappelhoff:
Apparently the BrainVision file system will (as for now) be the only multi-file system in BIDS once the EEG spec is merged.

> EDIT: in MEG we have the file format ending in `.ds`, which is a directory ... so also kind of multifile

I have started to write some code to validate the BrainVision file triplets [here](https://github.com/sappelhoff/brainvision-reader). The aim is to make this an npm package for easy integration into the BIDS-validator.

It could then conveniently go right below this line:

https://github.com/INCF/bids-validator/blob/fc7ac71b1080918a4d17316389d54b89d6bf3ddb/validators/bids.js#L251-L253

@robertoostenveld  @CPernet

sappelhoff:
edit: This is now located here:
https://github.com/bids-standard/bids-validator/blob/c9901bc9c53480a8b9b13fedbe6ab5ae7a265b5c/validators/bids/groupFileTypes.js#L48

and needs some work

olgn:
reopening because the changes implemented in #701 broke the browser version of the validator and were reverted

sappelhoff:
relevant conversation in https://github.com/bids-standard/bids-validator/issues/718

PR #490, opened 2018-07-22
search for MEG modalities

### LINKED ISSUE #489: no modalities for MEG (and EEG) in summary
When running the BIDS-validator successfully, there is a nice summary output at the end such as:
```


        Summary:                 Available Tasks:        Available Modalities: 
        57 Files, 19.34kB        rhyme judgment          T1w                   
        13 - Subjects                                    inplaneT2             
        1 - Session                                      bold   
```
When testing data that contain MEG modalities (e.g., ds000117), that modality does not show up. Same thing for the EEG validator, which has not yet been merged but is even with master besides the few EEG-specific commits.

I was looking for why this might be the case and stumbled over these lines:

https://github.com/INCF/bids-validator/blob/88a30a31babf6d1f91d2a9d54efab8c150a7c3b8/validators/bids.js#L249-L258

Why is the modality part indented so as to only occur, if a nifti file was found? Is that perhaps a bug?

If not, what else could be the reson that MEG and EEG modalities are not shown in the summary?

sappelhoff:
fixes #489

### DIFF
--- a/validators/bids.js
+++ b/validators/bids.js
@@ -246,7 +246,7 @@ BIDS = {
             }
 
             // check nifti and MEG files
-        else if (RegExp('\.nii|i\.gz|\.fif|f\.gz|\.sqd|\.con|\.kdf|\.chn|\.trg|\.raw|\.mhf').test(file.name.slice(-4))) {
+        else if (RegExp('^.*\.(nii|nii\.gz|fif|fif\.gz|sqd|con|kdf|chn|trg|raw|raw\.mhf)$').test(file.name)) {
 
             // capture nifties for later validation
             if (file.name.endsWith('.nii') || file.name.endsWith('.nii.gz')) {niftis.push(file);}
--- a/validators/bids.js
+++ b/validators/bids.js
@@ -246,12 +246,7 @@ BIDS = {
             }
 
             // check nifti and MEG files
-            else if (file.name.endsWith('.nii') || file.name.endsWith('.nii.gz') ||
-                     file.name.endsWith('.fif') || file.name.endsWith('.fif.gz') ||
-                     file.name.endsWith('.sqd') || file.name.endsWith('con') ||
-                     file.name.endsWith('kdf') || file.name.endsWith('chn') ||
-                     file.name.endsWith('trg') || file.name.endsWith('raw') ||
-                     file.name.endsWith('raw.mhd')) {
+        else if (RegExp('\.nii|i\.gz|\.fif|f\.gz|\.sqd|\.con|\.kdf|\.chn|\.trg|\.raw|\.mhf').test(file.name.slice(-4))) {
 
             // capture nifties for later validation
             if (file.name.endsWith('.nii') || file.name.endsWith('.nii.gz')) {niftis.push(file);}
--- a/validators/bids.js
+++ b/validators/bids.js
@@ -176,7 +176,7 @@ BIDS = {
                 }
                 summary.size += file.stats.size;
             }
-        }); 
+        });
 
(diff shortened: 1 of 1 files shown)


validators/bids.js
```diff
@@ -245,19 +245,26 @@ BIDS = {
...
-                    summary.modalities.push(suffix);
-                }
+            // check nifti and MEG files
+            else if (file.name.endsWith('.nii') || file.name.endsWith('.nii.gz') ||
+                     file.name.endsWith('.fif') || file.name.endsWith('.fif.gz') ||
+                     file.name.endsWith('.sqd') || file.name.endsWith('con') ||
+                     file.name.endsWith('kdf') || file.name.endsWith('chn') ||
+                     file.name.endsWith('trg') || file.name.endsWith('raw') ||
+                     file.name.endsWith('raw.mhd')) {
+
```
sappelhoff: This fix works (at least for for ds000117) and is pretty straight forward. It can also be expanded once the EEG-validator is merged. We could then simply add a few more `|| (file.name.endsWith('.eeg))`. A more elegant solution would be something like (pseudocode): `if file.name.endsWith.anyOf([*array of file endings*])` But I don't know (yet) how to do that in JS.

chrisgorgo: No worries - it's a good start.

validators/bids.js
```diff
@@ -245,19 +245,26 @@ BIDS = {
...
+            // capture nifties for later validation
+            if (file.name.endsWith('.nii') || file.name.endsWith('.nii.gz')) {niftis.push(file);}
+
+            // collect modality summary
```
sappelhoff: @robertoostenveld ... the line below "capture nifties for later validation" would be an excellent entry point to "capture BrainVision files for validation" --> i.e., a quick check whether the file triplets are present (.eeg, .vhdr, .vmrk).

chrisgorgo: Good pointer for a future PR adding basic EEG header validation!

validators/bids.js
chrisgorgo: Seems that CodeClimat is not happy with such a long logical expression - could you try to replace it with a regular expression. Something similar to https://github.com/INCF/bids-validator/blob/master/utils/type.js#L47

validators/bids.js
```diff
@@ -246,12 +246,7 @@ BIDS = {
             }
 
             // check nifti and MEG files
-            else if (file.name.endsWith('.nii') || file.name.endsWith('.nii.gz') ||
-                     file.name.endsWith('.fif') || file.name.endsWith('.fif.gz') ||
-                     file.name.endsWith('.sqd') || file.name.endsWith('con') ||
-                     file.name.endsWith('kdf') || file.name.endsWith('chn') ||
-                     file.name.endsWith('trg') || file.name.endsWith('raw') ||
-                     file.name.endsWith('raw.mhd')) {
+        else if (RegExp('\.nii|i\.gz|\.fif|f\.gz|\.sqd|\.con|\.kdf|\.chn|\.trg|\.raw|\.mhf').test(file.name.slice(-4))) {
```
chrisgorgo: What about using the end of line special character `$`?

sappelhoff: @chrisfilo, I changed the RegExp a bit so that I don't have to use the `.slice`.

chrisgorgo:
Awesome thanks!

validators/bids.js
sappelhoff: Pointing to issue #475

sappelhoff:
Happy to help :-) What's the merging policy?

PR #503, opened 2018-08-09
Add warning for .tsv files without accompanying data dictionary (json)

bGrass:


nellh (changes requested):
This looks good except the error code should be updated to 78 so it can cleanly merge now that #501 is in.

nellh:
Tests do pass with the es6 linting fix that's now on master.

nellh:
Actually lets change this to code 82 to avoid conflicts with #451

nellh (approved):

PR #506, opened 2018-08-13
Sane echo times

### LINKED ISSUE #191: Check EchoTime1 and EchoTime2 for sane values
The value of (EchoTime2 - EchoTime1) should be within the range of 0.0001 and 0.01.

olgn:
fixes #191 

* throws an error if (EchoTime2 - EchoTime1) > 0.01 or (EchoTime2 - EchoTime1) < 0.0001 in sidecar dictionaries associated with _phasediff.nii files (these values are 'unreasonable'). 
* implements unit tests that demonstrate this idea


nellh (approved):
Looks good, matches spec definitions and tests are correct.

PR #507, opened 2018-08-14
Proper acq time format

### LINKED ISSUE #461: acq_time in scans.tsv
Hi,

I'm not sure the validator is checking the acq_time field properly.

Acording to the BIDS specs: "Datetime should be expressed in the following format 2009­-06­-15T13:45:30 (year, month, day, hour (24h), minute, second; this is equivalent to the RFC3339 “date-time” format, time zone is always assumed as local time)."

Martin @martcous discovered our MEG-BIDS datasets: 
e.g. https://openneuro.org/datasets/ds000247
SUB-0002_SES-0001_SCANS.TSV
Has a value of: 18901014T123300
Instead of: 1890-10-14T12:33:00

Cheers,

olgn:
fixes #461 

* checks the acq_time column of _scans.tsv, if present and returns an issue if the value is not properly formatted (in YYY-MM-DDTHH:mm:ss style)
* adds unit tests for properly + improperly formatted _scans.tsv files

to pass tests, this requires changes to improperly formatted BIDS-example datasets found [here](https://github.com/INCF/BIDS-examples/pull/114). once this change is merged I will re-run circle ci tests to get this through.


validators/tsv.js
```diff
@@ -227,6 +234,28 @@ var checkage89_plus = function(rows, file, issues){
...
+const checkAcqTimeFormat = function(rows, file, issues) {
+    const format = 'YYYY-MM-DDTHH:mm:ss';
+    const header = rows[0].trim().split('\t');
+    const acqTimeColumn = header.indexOf('acq_time');
+    const testRows = rows.slice(1);
+    for (let i = 0; i < testRows.length; i++) {
+        const line = testRows[i];
+        const lineValues = line.trim().split('\t');
+        const acqTime = lineValues[acqTimeColumn];
+        if (acqTime && !moment(acqTime, format, true).isValid()) {
```
nellh: Moment is overkill for this and this impacts bundle size significantly. We should either perform this check with our own function or use a lighter library like date-fns (I'd use date-fns@next since 2.0 is close). It has a very similar parse function that takes the same format string.

nellh (changes requested):

olgn:
@nellh replaced moment.js with date-fns. this will still need the BIDS-example pr to go through before this can pass tests

nellh (approved):

nellh:
I'll merge this once https://github.com/INCF/BIDS-examples/pull/114 lands.

PR #509, opened 2018-08-15
304 slice timing length

### LINKED ISSUE #304: Validate that each value in the SliceTiming array has the right number of elements
Should match to the 'k' dimension of the corresponding nifti volume.

olgn:
fixes #304 

* checks the length of the SliceTiming array and throws a warning if the length is not the same as the 'k' dimension of the associated nifti header.
* includes unit tests


nellh (approved):
This looks good, can you resolve the conflicts?

PR #513, opened 2018-08-20
more robust .bvec and .bval validation

### LINKED ISSUE #229: More robust validation of bvecs / bvals
[Current code](https://github.com/INCF/bids-validator/blob/master/validators/nii.js#L47-L50) loads the bvecs & bvals files assuming that they each have the correct number of lines, and tests the line lengths only. If the files are completely malformed (e.g. somebody gets the bvecs and bvals files mixed up), this can result in a "Object 'undefined' does not have function '`replace()`'" error (or something like that). Better to test the number of non-empty lines first.

olgn:
fixes #229 

a bug in the bval.js + bvec.js validation logic failed to check if the `contents` of .bval and .bvec files are defined + of type string before calling string-ony function `.replace()`.

* checks to make sure that contents of .bval and .bvec files are defined and of type string
* throws error if the above condition is not met (88 for .bval, 89 for .bvec)
* implements unit tests for these conditions in bval.spec.js and bvec.spec.js

### DIFF
--- a/tests/bval.spec.js
+++ b/tests/bval.spec.js
@@ -22,4 +22,18 @@ describe('bval', function() {
       assert(issues.length == 1 && issues[0].code == 47)
     })
   })
+
+  it('should not allow undefined bvals', function() {
+    const bval = undefined
+    validate.bval({}, bval, function(issues) {
+      assert(issues.length == 1 && issues[0].code == 89)
+    })
+  })
+
+  it('should not allow bvals of types other than string', function() {
+    const bval = [0, 1, 2, 3]
+    validate.bval({}, bval, function(issues) {
+      assert(issues.length == 1 && issues[0].code == 89)
+    })
+  })
 })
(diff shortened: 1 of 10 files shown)


bGrass (approved):

chrisgorgo:
I'm not sure how content of bvec/bval file might not be read as a string. The original issue was mentioning mixed up bvec with bval as a motivation - maybe that should be the in the tests?

olgn:
@chrisfilo I believe these tests would be redundant - as the existing tests cover cases where the bval / bvec are interchanged:

- bvec submitted to bval:
https://github.com/INCF/bids-validator/pull/513/files#diff-ea943486d593a4c92ecfdd93733c75e3L12

- bval submitted to bvec:
https://github.com/INCF/bids-validator/pull/513/files#diff-6bef18600f49499cb535eebd20203d07L12

let me know if im missing something here

chrisgorgo:
Ok - lets merge it then as soon as codeclimate is happy.

PR #525, opened 2018-08-23
Json sidecar nodatafile error

### LINKED ISSUE #358: Raise an error if a JSON sidecar is found with no corresponding data file
For example:
image
https://openneuro.org/datasets/ds001110/versions/00001

bGrass:
Fixes #358


chrisgorgo:
That is a lot of files being modified... perhaps something went wrong when branching?

bGrass:
Most of the modified files are a test dataset that I added in order to test the validation of this condition.

utils/files.js
```diff
@@ -27,9 +27,24 @@ var fileUtils = {
...
   getBFileContent: getBFileContent,
+  dataExtRE: dataExtRE,
 }
 
 // implementations ----------------------------------------------------------------
+/**
+ * Make RegExp for detecting modalities from data file extensions
+ */
+function dataExtRE() {
+  return new RegExp(
```
nellh, line 38: The RegExp here could just be exported, it doesn't need to be a function.

nellh (approved):
Made a comment about the regexp export but that's changing in another issue anyways, so I think this is good to go.

PR #529, opened 2018-08-28
Move any string output from character -> reason fields.

nellh:
This fixes some inconsistent use of the Issue.character field so it always contains integers. These messages are instead provided as reasons for the issue.


nellh:
Related to https://github.com/OpenNeuroOrg/openneuro/issues/858

olgn (approved):
lgtm

PR #533, opened 2018-08-29
[MRG] Add coordsystem eeg

sappelhoff:
so far, the eeg coordsystem definition was squashed into the meg coordsystem using "anyOf". This is now fixed by introducing an eeg specific file, adjusting the meg file, and making the appropriate calls in validators/json.js

### DIFF
--- a/validators/schemas/coordsystem_eeg.json
+++ b/validators/schemas/coordsystem_eeg.json
@@ -1,0 +1,15 @@
+{
+  "type": "object",
+  "properties": {
+    "IntendedFor": { "type": "string", "minLength": 1 },
+    "FiducialsDescription": { "type": "string", "minLength": 1 },
+    "EEGCoordinateSystem": { "type": "string", "minLength": 1 },
+    "EEGCoordinateUnits": { "type": "string", "minLength": 1 },
+    "EEGCoordinateSystemDescription": { "type": "string" },
+    "AnatomicalLandmarkCoordinateSystem": { "type": "string", "minLength": 1 },
+    "AnatomicalLandmarkCoordinateUnits": { "type": "string", "minLength": 1 },
+    "AnatomicalLandmarkCoordinateSystemDescription": { "type": "string" }
+  },
+  "required": ["EEGCoordinateSystem", "EEGCoordinateUnits"],
+  "additionalProperties": false
+}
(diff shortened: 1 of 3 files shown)


sappelhoff:
Time for a short review @robertoostenveld ?

This is pretty straight forward though, so I think it can be merged @chrisfilo

validators/schemas/coordsystem_eeg.json
```diff
@@ -0,0 +1,15 @@
...
+    "IntendedFor": { "type": "string", "minLength": 1 },
+    "FiducialsDescription": { "type": "string", "minLength": 1 },
+    "EEGCoordinateSystem": { "type": "string", "minLength": 1 },
+    "EEGCoordinateUnits": { "type": "string", "minLength": 1 },
+    "EEGCoordinateSystemDescription": { "type": "string" },
+    "AnatomicalLandmarkCoordinateSystem": { "type": "string", "minLength": 1 },
+    "AnatomicalLandmarkCoordinateUnits": { "type": "string", "minLength": 1 },
+    "AnatomicalLandmarkCoordinateSystemDescription": { "type": "string" }
+  },
+  "required": ["EEGCoordinateSystem", "EEGCoordinateUnits"],
```
robertoostenveld, line 13: electrode positions are not always present (e.g. clinical EEG data, only standard 10-20 electrode names), hence I think that the coordinate system should not be required, only recommended.

robertoostenveld: Just traveling back from BIOMAG, where many groups presented pilot results from OPMs. It could very well be that also for OPM-based MEG systems we will get the situation that channel positions are not numerically specified.

sappelhoff: The "required" key just kicks in IF there is a coordsystem_eeg.json. And if we do have coordinates, there should be metadata on the system and its units.

If no coordsystem.json is specified, this json schema is not loaded and thus, nothing will be required.

robertoostenveld: perfect, thanks!

sappelhoff:
> Just traveling back from BIOMAG, where many groups presented pilot results from OPMs. It could very well be that also for OPM-based MEG systems we will get the situation that channel positions are not numerically specified.

I don't really know what you mean by OPM, but it sounds like something to be discussed in our Spec.

Thanks for the review.

robertoostenveld:
OPM stands for Optically Pumped Magnetometer, which is a new type of MEG sensor. See here for a recent review https://www.nature.com/articles/nature26147. Rigid 3D printed arrangements are now mostly used, but flexible OPM caps (like EEG caps) are already being piloted.

sappelhoff:
@chrisfilo, please merge or object :-)

PR #545, opened 2018-09-06
failing fs access in web validator

### LINKED ISSUE #544: TypeError: Cannot set property 'files' of undefined
Hi all, ran the bids validator last week without issues. Today it appears to be failing across all two servers that have access to the datasets I'm working with. Interestingly, one of the newer builds (e.g. https://1456-37161308-gh.circle-artifacts.com/0/root/web_version/index.html) that I spotted in an unrelated question on neurostars works just fine (see: https://neurostars.org/t/surprising-error-with-bids-validator/2347/2). 

The developer console error is beyond my ken:

**Here's the one in Chrome:**

Uncaught TypeError: Cannot set property 'files' of undefined
    at Object.format (app.min.js:1)
    at app.min.js:1
    at app.min.js:1
    at u (app.min.js:1)
    at app.min.js:1
    at m.run (app.min.js:1)
    at h (app.min.js:1)

**And in firefox:**

TypeError: u[f.code] is undefined[Learn More] app.min.js:1:148734

olgn:
fixes #544 (a regression caused by merge conflict resolution [here](https://github.com/INCF/bids-validator/commit/abe53f6d8728f598f0ddde60454897f697748f73#diff-fb0b7896fb16689756863e176a24a827R19)). 

we now use typeof window === 'undefined' (`isNode` vs `fs`) as a check for node vs web environment

### DIFF
--- a/utils/files/readFile.js
+++ b/utils/files/readFile.js
@@ -16,7 +16,7 @@ const isNode = typeof window === 'undefined'
  */
 function readFile(file) {
   return new Promise((resolve, reject) => {
-    if (fs) {
+    if (isNode) {
       testFile(file, function(issue) {
         if (issue) {
           process.nextTick(function() {


chrisgorgo:
CircleCI now generates web version of the validator for each PR (see in artifacts). You can grab that link and ask the person reporting the issue if this solves it for them.

olgn:
@chrisfilo that's rad!

nellh (approved):

PR #546, opened 2018-09-06
Abstract away regular expressions for path validation in a programming language agnostic way

### LINKED ISSUE #429: Abstract away regular expressions for path validation in a programming language agnostic way
Replacing all of the regular expressions from https://github.com/INCF/bids-validator/blob/master/utils/type.js with an import of an external package that provides only the regular expressions. See https://github.com/INCF/pybids/issues/131#issuecomment-370916864 for context.

bGrass:
Replaced local RegExps with updated RegExps in bids-validator-common.

Fixes #429 

Depends on #560

### DIFF
--- a/utils/bids_files.js
+++ b/utils/bids_files.js
@@ -3,7 +3,7 @@
  * eg. corresponding files
  */
 // dependencies -------------------------------------------------------------------
-const dataExtRE = require('./data_extensions.js')
+const type = require('./type.js')
 
 // public API ---------------------------------------------------------------------
 const bidsFileUtils = {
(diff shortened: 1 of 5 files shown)


chrisgorgo:
What's the plan for handling regular expressions for work-in-progress extensions (such as tests only enabled with `--bep006`)?

bGrass:
@chrisfilo My thinking was that it would make sense to update bids-validator-common with any new or modified regular expressions so it will serve as a single source of truth that's language agnostic.

olgn (approved):

nellh (approved):

PR #556, opened 2018-09-12
change help message color to white

### LINKED ISSUE #554: Change color of "further questions" link in validator output
Currently we have a red

`If you have any questions please post on https://neurostars.org/tags/bids`

output with every print that the BIDS-validator does. I think that this comes at the expense of the user experience because we are used to associating red with errors ... 

IMO we should reserve:
- green if something is good (validation correct)
- yellow for warnings
- red for errors.

Let's change the red information `If you have any questions please post on https://neurostars.org/tags/bids` to white:
https://github.com/INCF/bids-validator/blob/6349fdbbac18688df7c5f9232eced9a7525268f9/cli.js#L154-L159

sappelhoff:
fixes #554

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -153,7 +153,7 @@ function logSummary(summary) {
 
     //Neurostars message
     console.log(
-      colors.red(
+      colors.white(
         'If you have any questions please post on https://neurostars.org/tags/bids',
       ),
     )


chrisgorgo:
Thanks!

PR #560, opened 2018-09-12
Merge bids-validator-common into bids-validator

bGrass:


nellh (changes requested):
The Python configuration can be at the top level (no need to nest bids_validator/bids_validator) and would be more intuitive to users expecting to consume the repo as a Python module.

bGrass:
@nellh Does this look better?

nellh (approved):

Issue #570, opened 2018-09-18
refactor bids.js


olgn:
bids.js, most notably fullTest(), has become too big for its britches! the goal of this refactor is to make this code cleaner + more maintainable, and to get rid of the outdated `async` library and use promises instead. maybe codeclimate will stop yelling at us, too...

- [x] separate bids.js into a module, with the different tests (fullTest, quickTest, etc) as module exports.
- [x] refactor and further divide fullTest into more maintainable chunks 
- [x] remove the async library from the bids-validator and substitute promises

nellh:
Maybe we can start moving some tests closer to the code under testing? Especially as things get more complex, it's helpful to find the tests near the code involved. This applies more to unit tests than integration tests which might not have an obviously related file. The usual convention for this is a __tests__ directory in the same directory as the source files being tested.

olgn:
@nellh sure, added to the to-do list

olgn:
as of #574 all this needs is the updates to the tests allowed by the refactor

olgn:
closing this, as the remaining test refinement is ongoing and convenient to do over time as we develop, rather than a heavy lift to increase coverage.

PR #572, opened 2018-09-19
Bids.js refactor

olgn:
start of work on #570. separates validators/bids.js into a module, with several individual test files: fullTest.js, quickTest.js, quickTestError.js, reset.js, start.js, and subSesMismatchTest.js. ensures codeclimate compliance for all functions and increased readability for all files except fullTest.js (future pr for this, as it is a big haul)


nellh (approved):
Looks good to me, I did some quick testing to see if this would break some consumers of the library but it looks like the API is maintained.

bGrass (approved):

PR #573, opened 2018-09-20
Add the ce label to anat type.spec and the corresponding test

naveau:
The optional contrast enhancement label  (`[_ce-<label>]`) was missing in `bids_validator/rules/file_level_rules.json`

See spec 1.0.2 : 8.3 Anatomy imaging data

I added the corresponding test.

Hope that helps


chrisgorgo:
Thanks!

PR #574, opened 2018-09-20
Remove async

olgn:
working toward #570. tests can + will be improved (in a future pr) to move away from the integration test style necessitated by fullTest.js before this refactor

* gets rid of outdated async library and replaces it with promise-based validation
* separates fullTest.js into several maintainable chunks
* refactors several clumsy validation modules
* brings the summary collection into its own util module
* updates _most_ functions to no longer rely on global variables to function (self.issues getting reassigned within functions and whatnot)
* moves all of the nifti validations (except for the niftiheader check) into one validation function
* all of the files get sorted by type _before_ the validation process begins, increasing readability of fullTest.js significantly

@nellh @bGrass please give this a good test run, as much of the code base has changed.


utils/files/collectDirectoryStatistics.js
```diff
@@ -1,16 +1,19 @@
-const async = require('async')
 const fs = require('fs')
 
-const collectDirectoryStatistics = (fileList, summary) => {
-  async.eachOfLimit(fileList, 200, function(file) {
+const collectDirectoryStatistics = fileList => {
```
nellh: If all this does is get the size, I'd rename it to collectDirectorySize.

utils/files/illegalCharacterTest.js
```diff
@@ -15,8 +14,11 @@ const illegalchar_regex_list = [
   [re.ses_re, 63, 'ses name contains illegal character:'],
 ]
 
-const illegalCharacterTest = (fileList, issues) => {
-  async.eachOfLimit(fileList, 200, function(file) {
+const illegalCharacterTest = fileList => {
+  let issues = []
```
nellh: This is a const, there's quite a few instances of let being used for variables that are not reassigned.

utils/summary/index.js
```diff
@@ -0,0 +1,11 @@
...
+const collectModalities = require('./collectModalities')
+const collectSessions = require('./collectSessions')
+const collectSubjects = require('./collectSubjects')
+
+module.exports = {
+  collectSummary,
+  collectModalities,
+  collectSessions,
+  collectSubjects,
+}
```
nellh: It's better to import these as `require('./summary/collectSummary.js')` than to provide this index module. This is important if we want to use ES6 modules (no index, static imports, tree shaking possible) without too much refactoring later. Pretty minor issue since it is a problem throughout the validator anyways though.

nellh (commented):
I took a look at the performance characteristics of this change, it does result in a small increase in memory usage with very large datasets but no more than 10% - so I think this is acceptable for removing a dependency and avoiding the complexity of nesting async calls. Performance is unchanged.

A few notes below, I'm still reviewing the code changes but here's some other initial feedback.

olgn:
@nellh those issues have been addressed

bGrass (approved):

nellh:
:+1: I think this is good once merge conflicts are resolved.

PR #579, opened 2018-09-21
[FIX #575] readDir.js now follow Symlink directories

### LINKED ISSUE #575: BIDS cli does not follow Symlink directories inside the dataset
I came to this error trying to validate a "sub"-dataset built by linking only some session directories from a main dataset in each subject directories.

Here is the tree of the "sub"-dataset built from `ds001`:
```
[...]
|-- sub-active36                                                                                                                                                                                               
|   |-- ses-posttraining -> ../../ds001/sub-active36/ses-posttraining                                                                                                                                          
|   `-- ses-pretraining -> ../../ds001/sub-active36/ses-pretraining                                                                                                                                            
|-- sub-active37                                                                                                                                                                                               
|   |-- ses-posttraining -> ../../ds001/sub-active37/ses-posttraining                                                                                                                                          
|   `-- ses-pretraining -> ../../ds001/sub-active37/ses-pretraining                                                                                                                                            
|-- sub-inhibition01                                                                                                                                                                                           
|   |-- ses-posttraining -> ../../ds001/sub-inhibition01/ses-posttraining                                                                                                                                      
|   `-- ses-pretraining -> ../../ds001/sub-inhibition01/ses-pretraining                                                                                                                                        
|-- sub-inhibition02                                                                                                                                                                                           
|   |-- ses-posttraining -> ../../ds001/sub-inhibition02/ses-posttraining                                                                                                                                      
|   `-- ses-pretraining -> ../../ds001/sub-inhibition02/ses-pretraining
[...]
```

Everything's working fine when I ran the validator in the browser but the cli version failed at the `Quick Test` :

```
[ERR]  The directory dsproject01/ failed an initial Quick Test.
```

I assumed that it was due to the symbolic link and attempt a monkey patch that work in my situation : https://github.com/naveau/bids-validator/commit/ce9b660abe5587f021ff015f79d80bbeab2d4c85

Basically, I test if the files read by `fs.readDirSync` are Symbolic Links pointing to a directory.

I don't know if this is a case you want to cover in the code and if there is a better way to handle it. I can create a WIP pull request if you provide me some hints.

naveau:
Should fix #575 

I don't have any test for this PR. Mocking symlink (https://github.com/tschaub/mock-fs) is probably an overkill.

### DIFF
--- a/utils/files/readDir.js
+++ b/utils/files/readDir.js
@@ -108,7 +108,7 @@ function getFiles(dir, files_, rootpath, ig) {
       fileObj.ignore = true
     }
 
-    if (isDirectory(fullPath) {
+    if (isDirectory(fullPath)) {
       getFiles(fullPath, files_, rootpath, ig)
     } else {
       files_.push(fileObj)
--- a/utils/files/readDir.js
+++ b/utils/files/readDir.js
@@ -108,12 +108,18 @@ function getFiles(dir, files_, rootpath, ig) {
       fileObj.ignore = true
     }
 
-    if (fs.lstatSync(fullPath).isDirectory()) {
+    var isDirectory = fs.lstatSync(fullPath).isDirectory()
+    if (fs.lstatSync(fullPath).isSymbolicLink()) {
+      targetPath = fs.realpathSync(fullPath)
+      isDirectory = fs.lstatSync(targetPath).isDirectory()
+    }
+    if (isDirectory) {
       getFiles(fullPath, files_, rootpath, ig)
     } else {
       files_.push(fileObj)
     }
   })
+
   return files_
 }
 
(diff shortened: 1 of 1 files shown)


utils/files/readDir.js
```diff
@@ -108,15 +108,28 @@ function getFiles(dir, files_, rootpath, ig) {
...
   })
+
   return files_
 }
 
+/**
+ * Leverage fs.isDirectory by following Symbolic Links
+ */
+function isDirectory(path) {
+  var isDir = fs.lstatSync(path).isDirectory()
```
olgn: can we set fs.lstatSync(path) to a variable and then reuse it on this line (125) and this line (126)? for performance reasons it makes sense to only call this function once.

olgn (changes requested):
@naveau this looks great. 

I have requested one small change for performance reasons

utils/files/readDir.js
naveau: Done. Sounds indeed reasonable !

olgn (approved):

Issue #510, opened 2018-08-16
Enable the iEEG extension (BEP 010)


choldgraf:
This is an issue to coordinate the steps we need to take before the iEEG extension can be "activated" in the bids-validator. Right now it is off by default, but can be turned on with a flag.

Was originally in https://github.com/INCF/bids-validator/pull/451 but we decided to merge that in order to make refactoring easier.

## To-do list for this PR to get merged

- [x] Agree on how to reference the coordinate systems in the main BIDS spec
- [x] Agree on nomenclature around "electrodes" vs "contacts" (e.g. for electrode shanks vs single electrodes)
- [x] Decide on a final list of file extensions to support.
  - Current extensions suggested: EDF, VHDR
  - Current extensions allowed in addition: MEF3, SET, NWB
  - In particular @jasmainak and @teonbrooks showed concerns with having EDF as a primary format we recommend people use.
- [x] Decide whether we want javascript readers to be a blocker on the specification (no)
- [x] Decide whether we want python readers to be a blocker on the specification (no)
- [x] Figure out prefixes for units
- [x] Relatively minor spot-checks in the google doc
- [x] Final pass
- [x] Merge!

## To test out this PR on your dataset:

I believe that the following will let you demo this repository on your local machine (cc @dorahermes )
1. Fork and clone the bids-validator repository:

       git clone https://github.com/bids-standard/bids-validator
2. Install the validator for local development:

       npm install -g 

   from now on when you run `bids-validator`, it should be using the code from this branch!

3. Run the BIDS validator with the BEP 010 flag, which activates the iEEG BEP extension:
   
       bids-validator --bep010 path/to/dataset

cc @dorahermes @sappelhoff @chrisfilo and @nellh as well, just FYI!

teonbrooks:
I think I have come around to the preferred EEG options especially in light of vhdr [writing](https://gitlab.com/palday/philistine/blob/master/philistine/mne/io.py) functionality and the its [validation](https://github.com/sappelhoff/brainvision-validator) efforts.

Since EDF is such a common format, writers already exist (although a compatible [one](https://github.com/mne-tools/mne-bids/issues/41) is needed for MNE). a js validator would be its next step.

sappelhoff:
currently, the validator only allows BrainVision and EDF, see: https://github.com/bids-standard/bids-validator/blob/cbcb0d31c5a945ef8be9e5c2b4e380b6d7a01a58/bids_validator/rules/file_level_rules.json#L179

[...]

PR #591, opened 2018-09-26
[WIP] updating ieeg spec with new info

choldgraf:
this PR updates the validator with changes in the ieeg spec since the latest push.

## To test out this PR on your dataset:

I believe that the following will let you demo this repository on your local machine (cc @dorahermes )
1. Fork and clone the bids-validator repository:

       git clone https://github.com/<your-username>/bids-validator
2. Change into its directory

       cd bids-validator

3. Add **my** repository as a remote:
  
       git remote add choldgraf https://github.com/choldgraf/bids-validator

4. Pull *this* branch from my repository:

       git fetch choldgraf ieeg_updates

5. Checkout the new branch

       git checkout ieeg_updtaes

   **NOTE THE TYPO IN UPDTAES....SORRY ABOUT THAT**
6. Install for local development:

       npm install -g 

from now on when you run `bids-validator`, it should be using the code from this branch!

If you want to update with newest changes, you should run

    git pull choldgraf ieeg_updtaes

## Todo

- [x] Add instructions for how to get the validator installed locally
- [ ] Add a warning if people use a "supported but not recommended" format
- [x] Incorporate updates from the latest email to the collaborators
- [x] Look into whether BIDS has a policy for a missing required metadata value
- [x] Add a string for DCOffset
- [ ] Check if we need (and if not remove) some extra file extensions we check for: https://github.com/bids-standard/bids-validator/issues/510#issuecomment-432531454

cc @dorahermes


validators/schemas/ieeg.json
```diff
@@ -2,6 +2,20 @@
...
+                    {"type": "object", "additionalProperties": {"type":"object"}},
+                    {"type": "string", "pattern": "^n/a$"}
+                  ]
+                },
+                "SoftwareFilters": {
+                  "anyOf": [
+                    {"type": "object", "additionalProperties": {"type":"object"}},
+                    {"type": "string", "pattern": "^n/a$"}
+                  ]
+                },
```
jasmainak: no tests to update for this?

sappelhoff (changes requested):
Please also address renaming `.dat` to `.eeg` in the regexps:

https://github.com/bids-standard/bids-validator/blob/3876694533738f2df73de14e9745a4f5f4ba3304/validators/bids/groupFileTypes.js#L53-L67

https://github.com/bids-standard/bids-validator/blob/cbcb0d31c5a945ef8be9e5c2b4e380b6d7a01a58/bids_validator/rules/file_level_rules.json#L179

... and perhaps in other spots. BrainVision has its binary data in `.eeg` files, nobody can remember where the `.dat` came from

choldgraf:
I'm working on implementing these changes, but it seems that the validator codebase has changed a bit since I last took a pass. A few questions:

1. @chrisfilo is there a place that describes the current organization of the codebase from a developer standpoint?
2. As @sappelhoff notes above, what's the difference between the file names specified in `groupFileTypes.js` vs. those in `file_level_rules.json`? Should they just be the same?
3. Where should something like "warn if the data is a subset of extensions" go?
4. Does anybody know if BIDS has a policy for what to do if a required field is missing? (e.g. we have hardwarefilters as required, but sometimes people simply don't have it)

chrisgorgo:
1. Unfortunately there isn't, but if you have specific questions perhaps @olgn can help
2. Not sure what that list is doing in `groupFileTypes.js` It probably shouldn't be there.
3. Hmm... I don't quite follow. What is a "subset of extension?
4. If a required field is missing this should be an error. Perhaps this field should be recommended.

choldgraf:
@chrisfilo I mean that when we check for data extensions (eg, EDF, fif, etc) a subset of those extensions are allowed, but not "officially recommended". We want to raise a warning if those are found, but still allow them to pass the validator.

choldgraf:
I believe that @olgn was the one that added the files in `groupFileTypes.js`...perhaps they can clarify what this section does?

choldgraf:
ok - I think that most of the changes are implemented that were in the google doc. Only a couple questions remain:

1. Not sure where (or if) we can raise a warning if a subset of extensions is used for data.
2. Not sure what's up with the `groupFileTypes.js` code (see comment above)

Other than that, I need to make sure that tests pass on the validator (can't do this now because I'm on a plane w/ crappy internet).

validators/tsv/tsv.js
```diff
@@ -241,11 +241,9 @@ const TSV = (file, contents, fileList, callback) => {
...
     checkheader('type', 1, file, 72)
     checkheader('units', 2, file, 72)
-    checkheader('sampling_frequency', 3, file, 72)
-    checkheader('low_cutoff', 4, file, 72)
-    checkheader('high_cutoff', 5, file, 72)
-    checkheader('notch', 6, file, 72)
-    checkheader('reference', 7, file, 72)
+    checkheader('low_cutoff', 3, file, 72)
+    checkheader('high_cutoff', 4, file, 72)
+    checkheader('reference', 5, file, 72)
```
sappelhoff: According to the [spec](https://docs.google.com/document/d/1qMUkoaXzRMlJuOcfTYNr3fTsrl4SewWjffjMD5Ew6GY/edit#heading=h.vntcs5d7jpb3), the reference column is recommended, not required ... I think this is with good reason, because it would only be of relevance in a bipolar recording scheme or with many special AUX channels that all have their own special reference. I suggest to remove this from the check:

```suggestion

```

tests/tsv.spec.js
```diff
@@ -295,18 +295,9 @@ describe('TSV', function() {
...
-      'value-one\tvalue-two\tvalue-three\tvalue-four\tvalue-five\tvalue-six\tvalue-seven\tvalue-eight'
-    validate.TSV.TSV(channelsFileIEEG, tsv, [], function(issues) {
-      assert(issues.length === 1 && issues[0].code === 72)
-    })
-  })
-
   it('should not allow iEEG channels.tsv files without low_cutoff column', function() {
     var tsv =
-      'name\ttype\tunits\tsampling_frequency\tcolumn_five\thigh_cutoff\tnotch\treference\n' +
+      'name\ttype\tunits\textra-column\thigh_cutoff\treference\n' +
```
sappelhoff: fix according to removing `reference` as a required field (see below)
```suggestion
      'name\ttype\tunits\textra-column\thigh_cutoff\n' +
```

tests/tsv.spec.js
```diff
@@ -295,18 +295,9 @@ describe('TSV', function() {
...
       'value-one\tvalue-two\tvalue-three\tvalue-four\tvalue-five\tvalue-six\tvalue-seven\tvalue-eight'
```
sappelhoff: too many columns in this row
```suggestion
      'value-name\tvalue-type\tvalue-units\tvalue-fake\tvalue-highcut'
```

tests/tsv.spec.js
```diff
@@ -315,16 +306,7 @@ describe('TSV', function() {
...
-      'value-one\tvalue-two\tvalue-three\tvalue-four\tvalue-five\tvalue-six\tvalue-seven\tvalue-eight'
-    validate.TSV.TSV(channelsFileIEEG, tsv, [], function(issues) {
-      assert(issues.length === 1 && issues[0].code === 72)
-    })
-  })
-
-  it('should not allow iEEG channels.tsv files without notch column', function() {
-    var tsv =
-      'name\ttype\tunits\tsampling_frequency\tlow_cutoff\thigh_cutoff\tcolumn_seven\treference\n' +
+      'name\ttype\tunits\tlow_cutoff\textra-column\treference\n' +
```
sappelhoff: again, remove reference (see below)
```suggestion
      'name\ttype\tunits\tlow_cutoff\textra-column\n' +
```

tests/tsv.spec.js
```diff
@@ -315,16 +306,7 @@ describe('TSV', function() {
...
       'value-one\tvalue-two\tvalue-three\tvalue-four\tvalue-five\tvalue-six\tvalue-seven\tvalue-eight'
```
sappelhoff: and too many values (same as above)
```suggestion
      'value-name\tvalue-type\tvalue-units\tvalue-lowcut\tvalue-fake'
```

tests/tsv.spec.js
```diff
@@ -333,7 +315,7 @@ describe('TSV', function() {
 
   it('should not allow iEEG channels.tsv files without reference column', function() {
```
sappelhoff: this test should be deleted, because the reference column is not required (needs to be changed below as well) ... I'd love to make a suggestion here as well, but Github does not allow multi-line suggestions so far :-O

tests/tsv.spec.js
```diff
@@ -342,8 +324,8 @@ describe('TSV', function() {
 
   it('correct columns should pass for iEEG channels.tsv file', function() {
     var tsv =
-      'name\ttype\tunits\tsampling_frequency\tlow_cutoff\thigh_cutoff\tnotch\treference\n' +
-      'value-one\tvalue-two\tvalue-three\tvalue-four\tvalue-five\tvalue-six\tvalue-seven\tvalue-eight'
+      'name\ttype\tunits\tlow_cutoff\thigh_cutoff\treference\n' +
```
sappelhoff:
```suggestion
      'name\ttype\tunits\tlow_cutoff\thigh_cutoff\n' +
```

tests/tsv.spec.js
```diff
@@ -342,8 +324,8 @@ describe('TSV', function() {
...
+      'value-one\tvalue-two\tvalue-three\tvalue-four\tvalue-five\tvalue-six\t'
```
sappelhoff:
```suggestion
      'value-name\tvalue-type\tvalue-units\tvalue-lowcut\tvalue-highcut'
```

sappelhoff (changes requested):
Looks good, I think the tests were failing because you edited the header columns in the `chanels.tsv` files, but forgot to also truncate the other columns. Other than that, the main suggestions are:

1. remove the required `reference`  column, it is only recommended in the spec
2. please change the `.dat` extensions to `.eeg` in the regexp sections for iEEG
3. potentially remove these unknown extensions tdix, tmet, ... - see [here](https://github.com/bids-standard/bids-validator/issues/510#issuecomment-431556824)

Finally, we also wanted to add the `value` and `sample` fields for the `events.tsv`. I got started with this for EEG already (see https://github.com/bids-standard/bids-validator/pull/599), but @chrisfilo wants it to be conditional on the bep flag, so we might want to wait with that in order not to complicate things.

I made some inline code suggestions that should be quick fixes :-)

validators/json/schemas/ieeg.json
```diff
@@ -1,45 +1,59 @@
...
+    "ElectricalStimulationParameters": {"type": "string"},
   },
   "required": [
     "TaskName",
-    "Manufacturer",
-    "PowerLineFrequency",
     "SamplingFrequency",
+    "PowerLineFrequency",
+    "SoftwareFilters"
     "iEEGReference"
```
sappelhoff, line 56: @choldgraf I think the error is related to these lines:

https://github.com/bids-standard/bids-validator/blob/551878e7dfb6af1e641676521d2066056192e41c/tests/json.spec.js#L108-L115

You need to update the test to reflect the new "required" key/value pairs that have changed here (see lines 51 to 55 right above this comment)

sappelhoff (approved):
+1 for merge after the remaining fix of the test:

https://github.com/bids-standard/bids-validator/blob/551878e7dfb6af1e641676521d2066056192e41c/tests/json.spec.js#L108-L115

choldgraf:
Hmm, I tried updating those fields in the test, and I'm still getting an error on that one. It seems like the required fields are all accounted for in the test:

required fields:

https://github.com/bids-standard/bids-validator/pull/591/files#diff-565f0182ad77f553142b7ff49357cd13R51

tested fields:

https://github.com/bids-standard/bids-validator/pull/591/files#diff-bacd38195c17182690f432a3e896cf20R109

Any idea what's going on?

tests/json.spec.js
```diff
@@ -108,10 +108,10 @@ describe('JSON', function() {
   it('*_ieeg.json sidecars should have required key/value pairs', function() {
     var jsonObj = {
       TaskName: 'Audiovis',
-      Manufacturer: 'TDT',
-      PowerLineFrequency: 50,
       SamplingFrequency: 10,
-      iEEGReference: 'reference',
+      PowerLineFrequency: 50,
+      SoftwareFilters: 'high-pass',
```
sappelhoff: @choldgraf I think the way the software filter is specified does not conform to what we require:
```json
    "SoftwareFilters": {
      "anyOf": [
        {"type": "object", "additionalProperties": {"type":"object"}},
        {"type": "string", "pattern": "^n/a$"}
      ]
    }
```

It should either be: a string `"na"`

... or it could be an object of objects: `{'HighPass': {'HalfAmplitudeCutOffHz': 1, 'RollOff: '6dB/Octave'}}`

```suggestion
      SoftwareFilters: {'HighPass': {'HalfAmplitudeCutOffHz': 1, 'RollOff: '6dB/Octave'}},
```

choldgraf:
can anybody suggest a way to launch an interactive debugger during an `npm` test? It's super frustrating that I can't figure out what exactly the error is in the JSON dictionary...there's gotta be a more efficient way to do this than me looking at the circle builds each time...

sappelhoff:
not a real answer to your problem, but instead of `npm test`, you can also run these other commands that are being run by CircleCI:

https://github.com/bids-standard/bids-validator/blob/aa58a07258f398673cea98c041cd028f2a18b3fd/circle.yml#L17-L28

Then at least you don't have to wait for Circle to build something 🤔

chrisgorgo:
WebStorm is a convenient way to debug tests. Works like any other ide
pretty similar to pycharm.

choldgraf:
hey all - apparently @sappelhoff is a wizard and I owe him multiple beers

tests are passing

sappelhoff:
ready for a merge @choldgraf? I think the remaining point "Add a warning if people use a "supported but not recommended" format" might very well be its own PR

choldgraf:
I'm +1 on merge - I just opened https://github.com/bids-standard/bids-validator/issues/647 to keep track of the warning stuff.

I wouldn't be surprised if we need to iterate a bit more once people start trying out the new validator on their data, but we can do that in subsequent PRs.

choldgraf:
cc @dorahermes , who mentioned she wants to test out the validator on her data!

dorahermes:
would be good to have this included, will test extensively

sappelhoff:
perfect, thanks @choldgraf  for the PR and happy testing @dorahermes

choldgraf:
wooo, thanks for all the help!

PR #594, opened 2018-10-02
gzipped fif files are not allowed

robertoostenveld:


robertoostenveld:
see also https://github.com/bids-standard/bids-validator/issues/491

The gz compression is not done by the manufacturer. It breaks compatibility of the file format with their and other processing software.

chrisgorgo:
Shouldn't the fif.gz extension be added to the spec first?

robertoostenveld:
It was agreed with wide consensus that it should NOT be part of the spec. See 
https://github.com/bids-standard/bids-validator/issues/491#issuecomment-407100002

chrisgorgo:
Sorry I misread the PR.

robertoostenveld:
thanks

PR #596, opened 2018-10-04
fix in -> of issues

olgn:
a regression was caused on openneuro because for (let item of iterable) can trigger inherited properties on array prototypes... sad!


nellh (approved):
Looks good, I'll make a PR with the Eslint rule to catch this.

PR #598, opened 2018-10-04
Enable eslint guard-for-in to catch for..in vs for..of errors

nellh:
This eslint rule will often catch `for ... in` vs `for ... of` errors by warning you if you didn't guard for prototype inherited properties in `for ... in` loops. I've added the guards needed for the intentional `for ... in` loops.


olgn (approved):
* lint yells at us when we write code that assumes untouched prototypes
* we can start moving unit tests closer to the source code - nice!
* maybe someday all the `vars` will grow up to become `const` or `let`

PR #599, opened 2018-10-04
[MRG] add non-custom events columns

sappelhoff:
This is adding two column names to events.tsv:

- sample ... describing the exact digitization point at which an event had its onset 
- value ... describing the value of a potential TTL trigger marking the event onset

These are non-custom column names that will be introduced with BEP006 and BEP010.

### DIFF
--- a/bids-validator/bids_validator/tsv/non_custom_columns.json
+++ b/bids-validator/bids_validator/tsv/non_custom_columns.json
@@ -34,7 +34,9 @@
     "onset",
     "trial_type",
     "response_time",
-    "stim_file"
+    "stim_file",
+    "sample",
+    "value"
   ],
   "misc": [],
   "participants": ["participant_id"],


chrisgorgo:
Thanks. Could you make this check conditional on the `--bep006` flag?

sappelhoff:
Yes, but it involved changes in at least three spots. I think it might be easier to wait for a merge of EEG and then make this simple fix.

This PR is not too crucial, because it only fixes warnings, not errors :-)

sappelhoff:
for reference, it would require:

1. passing self.options here:
https://github.com/bids-standard/bids-validator/blob/5bf12b4f24fe0fb9072e4aa3454e78ec02d40a50/validators/bids/fullTest.js#L166

2. then here:
https://github.com/bids-standard/bids-validator/blob/5bf12b4f24fe0fb9072e4aa3454e78ec02d40a50/validators/tsv/validateTsvColumns.js#L51

3. and finally here:
https://github.com/bids-standard/bids-validator/blob/5bf12b4f24fe0fb9072e4aa3454e78ec02d40a50/validators/tsv/validateTsvColumns.js#L34

to then do some overly complicated if/else action

sappelhoff:
https://github.com/bids-standard/bids-specification/pull/108 has been merged, this one is ready to go.

Issue #603, opened 2018-10-04
[release 1.0.1] TypeError: Cannot read property 'reason' of undefined


benoitberanger:
Hello,

I have a problem with the CLI version, but only with the current release.
With the current release [1.0.1](https://github.com/bids-standard/bids-validator/releases/tag/1.0.1) I have the error : 
```
(node:4340) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
(node:4340) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'reason' of undefined
    at headerFields (/home/benoit.beranger/bids-validator/validators/headerFields.js:41:64)
    at NIFTI.validate.then.niftiIssues (/home/benoit.beranger/bids-validator/validators/bids/fullTest.js:151:54)
    at process._tickCallback (internal/process/next_tick.js:68:7)
(node:4340) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:4340) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```
But **no problem** with the previous release [1.0.0](https://github.com/bids-standard/bids-validator/releases/tag/1.0.0), neither the previous one [0.27.10](https://github.com/bids-standard/bids-validator/releases/tag/0.27.10)

Ben

benoitberanger:
By the way, I always have this line, so I don't think it's relevant :
`[DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.`

nellh:
Thanks for the report!

Few things I'm seeing for us to fix here:
- [x] We should fix that Buffer usage, it is just a warning but future versions of Node will remove this usage.
- [x] It looks like we're lacking test coverage for issue 39 - "Not all subjects/sessions/runs have the same scanning parameters."
- [x] When issue 39 is found it throws this error as of 1.0.1

nellh:
This will be fixed in 1.0.2

benoitberanger:
Great, thank you !

PR #606, opened 2018-10-04
Fix 603

olgn:
fix for #603 

* fixes regression due to poor array index calls within headerFields.js
* separates the issue 39 collection logic into its own function
* implements unit tests on this function

### DIFF
--- a/validators/__tests__/headerFields.spec.js
+++ b/validators/__tests__/headerFields.spec.js
@@ -1,0 +1,65 @@
+const assert = require('chai').assert
+const { collect39Issues } = require('../headerFields')
+
+describe('headerFields', () => {
+  describe('collect39Issues()', () => {
+    it('should return an empty array if there are no files in allIssues39Dict', () => {
+      const allIssues39Dict = {}
+      const issues = collect39Issues(allIssues39Dict)
+      assert.isArray(issues)
+      assert.lengthOf(issues, 0)
+    })
+    it('should return one issue per file in the allIssues39Dict', () => {
+      const allIssues39Dict = {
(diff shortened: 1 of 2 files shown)


validators/__tests__/headerFields.spec.js
```diff
@@ -0,0 +1,62 @@
...
+            reason: 'reason1',
+          },
+          {
+            code: 39,
+            reason: 'reason2',
+          },
+        ],
+      }
+      const issues = collect39Issues(allIssues39Dict)
+      assert(issues.length == 1 && issues[0].reason == ' reason1 reason2')
```
nellh: This should be two assertions.

validators/__tests__/headerFields.spec.js
```diff
@@ -0,0 +1,62 @@
+const assert = require('chai').assert
+const { collect39Issues } = require('../headerFields')
+
+describe('headerFields', () => {
+  describe('collect39Issues()', () => {
+    it('should return an empty array if there are no files in allIssues39Dict', () => {
+      const allIssues39Dict = {}
+      const issues = collect39Issues(allIssues39Dict)
+      assert(issues.length == 0)
```
nellh: If this test fails it throws a hard to read error. Better would be `assert.lengthOf` or `assert.equal` so the issue is clearer.

It's also better to do `issues.length === 0` to avoid the possible type coercion (as rare as that would be in this case).

nellh (changes requested):

nellh (approved):

benoitberanger:
Great, thank you !

PR #607, opened 2018-10-04
Fix bug #600 and add a couple tests

### LINKED ISSUE #600: [BUG] sidecar without datafile ... although there is one
I encountered an error that looks like a bug:

example: for [ds001](https://github.com/bids-standard/bids-examples/tree/master/ds001), try to define a file: `ds001/sub-02/func/sub-02_task-balloonanalogrisktask_run-01_events.json` and run the validator.

--> This should not yield an error, because the corresponding data file exists (ds001/sub-02/func/sub-02_task-balloonanalogrisktask_run-01_events.tsv`)

bGrass:
Fixes #600

### DIFF
--- a/utils/__tests__/bids_files.spec.js
+++ b/utils/__tests__/bids_files.spec.js
@@ -1,0 +1,38 @@
+const assert = require('chai').assert
+const { checkSidecarForDatafiles } = require('../bids_files.js')
+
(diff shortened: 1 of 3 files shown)


nellh (approved):
Looks good and works as expected with an example dataset now. :tada: for unit tests.

PR #614, opened 2018-10-05
Switch to Jest for test runner

nellh:
* Provides different environments for browser and nodejs specific tests, less boilerplate for browser tests
* Watch mode reruns only relevant tests, speeding up development
* Simplifies some of the dynamic tests
* It takes better advantage of multiple cores - `real    0m3.754s` from Jest and `real    0m6.088s` with Mocha

Note that the drop from 267 to 258 tests is expected with the removal of eight dummy tests and one obsolete test.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -9,7 +9,7 @@
     "browser": true,
     "commonjs": true,
     "node": true,
-    "mocha": true
+    "jest": true
   },
   "globals": {
     "Uint8Array": false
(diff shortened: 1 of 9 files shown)


olgn (approved):
* once results have been cached, performance is increased through use of multiple cores
* npm test /validators/__tests__ = shorthand to JUST test one folder of tests. what a concept!
* all tests still working

PR #615, opened 2018-10-05
Validate remote files

### LINKED ISSUE #458: Support validating git-annex repos with remote-only files
To be able to validate a dataset stored in S3, we need to extend the validator to accept a tree of files that includes some files only accessible by URL. The bulk of dataset content for OpenNeuro will soon be stored in an S3 bucket and not a local filesystem.

See https://github.com/OpenNeuroOrg/openneuro/issues/331

The URL list should be an ordered array of local paths and URLs to try. A file being read should look at the local version first, if the local version does not exist, then access each URL in order. Something like this:

```JSON
{
  "localPath": "dataset_description.json",
  "remotePaths": [
    "https://openneuro.s3.amazonaws.com/ds000001/dataset_description.json?versionId=ox7Poh4aweeg8YieXohsh9", 
    "https://openneuro.org/datalad/dataset/ds000001/files/dataset_description.json"
  ]
}
```

Could be passed to BIDS.validate as an optional argument.

@chrisfilo Thoughts on this?

bGrass:
Fixes #458

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -6,6 +6,7 @@ var cliff = require('cliff')
 var pluralize = require('pluralize')
 var bytes = require('bytes')
 var fs = require('fs')
+const remoteFiles = require('./utils/files/remoteFiles')
 
 module.exports = function(dir, options) {
   if (fs.existsSync(dir)) {
--- a/utils/files/readNiftiHeader.js
+++ b/utils/files/readNiftiHeader.js
@@ -4,6 +4,7 @@ const fs = require('fs')
 const zlib = require('zlib')
 const testFile = require('./testFile')
 const Issue = require('../../utils/issues').Issue
+const remoteFiles = require('./remoteFiles')
 
 const isNode = typeof window == 'undefined'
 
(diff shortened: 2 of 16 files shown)


chrisgorgo:
Great to see this coming along. The final version would benefit from some mention in the README about support for git-annex repositories.

Whenever it's ready it would be worth pinging @yarikoptic - he might be interested in this feature. Context: this PR will enable the validator to work with git-annex folders that do not have all of their content downloaded locally.

yarikoptic:
well, to say the truth I am not quite sure on the benefit for me:  if files under annex and I do want to validate the content I would just "git annex get" them first.  What would be important is to not have autofetching enabled by default, so if I do run validator in some annex'ed repo, it should not start downloading possibly gigabytes of data to validate.  Or am I missing the point?

chrisgorgo:
We need this in the context of OpenNeuro. When a user modifies a dataset that has been previously exported to S3 we would like to perform the validation by downloading as little data as possible.

nellh:
The goal is to get an initial implementation here and then optimize how much data is transferred after, since in many cases we only need to read headers or subsets of the files to validate the dataset.

yarikoptic:
ah, with "export to S3" you do not retain those files on your server... 
@nellh indeed in many cases getting just the header should suffice if the tool supports loading it without demanding the full file.  `git-annex` indeed then would be suboptimal there since it can only provide a full file.

ok, got it.

cli.js
```diff
@@ -6,6 +6,7 @@ var cliff = require('cliff')
 var pluralize = require('pluralize')
 var bytes = require('bytes')
 var fs = require('fs')
+const remoteFiles = require('./utils/files/remoteFiles')
```
olgn, line 9: remoteFiles does not appear to be used in this file. reference should be removed

utils/files/readNiftiHeader.js
```diff
@@ -4,6 +4,7 @@ const fs = require('fs')
 const zlib = require('zlib')
 const testFile = require('./testFile')
 const Issue = require('../../utils/issues').Issue
+const remoteFiles = require('./remoteFiles')
```
olgn, line 7: remoteFiles is imported but ne'er be it used

utils/files/remoteFiles.js
```diff
@@ -0,0 +1,168 @@
...
+
+  extractGzipBuffer: function(buffer, callback) {
+    const decompressStream = zlib
+      .createGunzip()
+      .on('data', function(chunk) {
+        callback(null, null, chunk)
+        decompressStream.pause()
+      })
+      .on('error', function() {
+        callback(new Issue({ code: 28, file: config.file }), null, null)
```
olgn, line 98: `config` is not defined here.

utils/files/readNiftiHeader.js
```diff
@@ -13,28 +14,38 @@ const isNode = typeof window == 'undefined'
...
-      return
-    }
-    if (stats.size < 348) {
-      callback({ error: new Issue({ code: 36, file: file }) })
-      return
-    }
-
-    return extractNiftiFile(file, callback)
-  })
+function nodeNiftiTest(file, callback, annexed, dir) {
```
olgn, line 25: it is my opinion that callback should always be the last argument - @nellh do you think that is nit-picking?

olgn:
closed in favor of #640.

Issue #582, opened 2018-09-24
Uncaught exceptions are creating invalid issues


chrisgorgo:
An exception in a different part of the code led to an invalid issue passed down the line

```
C:\gdrive\workspace\bids-validator\utils\issues\index.js:74
        categorized[issue.code].files = []
                                      ^

TypeError: Cannot set property 'files' of undefined
    at Object.format (C:\gdrive\workspace\bids-validator\utils\issues\index.js:74:39)
    at C:\gdrive\workspace\bids-validator\validators\bids\fullTest.js:509:43
    at C:\gdrive\workspace\bids-validator\node_modules\async\dist\async.js:473:16
    at iterateeCallback (C:\gdrive\workspace\bids-validator\node_modules\async\dist\async.js:992:24)
    at C:\gdrive\workspace\bids-validator\node_modules\async\dist\async.js:969:16
    at process._tickCallback (internal/process/next_tick.js:61:11)
```

olgn:
@chrisfilo do you happen to have a reproduction case?

chrisgorgo:
This dataset https://drive.google.com/file/d/19RYauOI9Zqd97T_yKpJfzt3KrZTVsX1M/view?usp=sharing prior to this fix https://github.com/INCF/bids-validator/pull/581

PR #636, opened 2018-10-26
catch exceptions and reformat as Issue

david-nishi:
fix for #582 

Notable changes 
- addition of catch to fullTest
- exceptions now reformatted as issues and added to issue error list (code 0), along with the callstack and directions to (https://github.com/bids-standard/bids-validator/issues)
- redirects invalid issues in the bval and bvec validators to fullTest catch

should work with both web and cli


validators/bval/validate.js
```diff
@@ -15,15 +15,21 @@ const validate = (files, bContentsDict) => {
...
         })
     })
   })
 
-  return new Promise(resolve =>
-    Promise.all(bvalPromises).then(() => resolve(issues)),
+  return new Promise((resolve, reject) =>
+    Promise.all(bvalPromises)
+      .then(() => resolve(issues))
+      .catch(reject),
```
nellh, line 30: This pattern appears a few times and can be simplified to this.

```javascript
return Promise.all(bvalPromises).then(() => issues)
```

david-nishi:
reopened PR as #638

validators/bval/validate.js
david-nishi: Thanks Nell!

PR #638, opened 2018-10-29
catch exceptions and reformat as Issue

david-nishi:
fix for #582

Notable changes

addition of catch to fullTest
exceptions now reformatted as issues and added to issue error list (code 0), along with the callstack and directions to (https://github.com/bids-standard/bids-validator/issues)
redirects invalid issues in the bval and bvec validators to fullTest catch
should work with both web and cli

### DIFF
--- a/tests/utils/issues.spec.js
+++ b/tests/utils/issues.spec.js
@@ -53,7 +53,7 @@ describe('issues', () => {
   })
 
   describe('exception/issue redirect', () => {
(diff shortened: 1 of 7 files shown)


david-nishi:
reopened from PR #636

chrisgorgo:
@DaNish808 wasn't the intention of this PR to present issues such as #655 as validation error rather than infinite spinning wheel and error in the JS console?

PR #640, opened 2018-10-31
Remote files validation

### LINKED ISSUE #458: Support validating git-annex repos with remote-only files

olgn:
refactor + improvements on #615. fixes #458.

* implements git-annex remote file access
* unit tests for utils/files/remoteFiles.js

### DIFF
--- a/utils/files/__tests__/remoteFiles.spec.js
+++ b/utils/files/__tests__/remoteFiles.spec.js
@@ -3,7 +3,6 @@ const remoteFiles = require('../remoteFiles')
 const fs = require('fs')
 const child_process = require('child_process')
 const zlib = require('zlib')
-const env = Object.assign({}, process.env)
 const config = {
   s3Params: {
     Bucket: 'none',
@@ -19,11 +18,14 @@ describe('remoteFiles', () => {
     global.fetch = jest
       .fn()
       .mockImplementation(
-        () => new Promise(resolve => resolve({ buffer: () => 'buffer' })),
+        () =>
+          new Promise(resolve => resolve({ ok: true, buffer: () => 'buffer' })),
       )
   })
 
-  describe('getAnnexedFile', () => {})
+  beforeEach(() => {
+    delete process.env.AWS_ACCESS_KEY_ID
+  })
 
   describe('accessRemoteFile', () => {
     it('should return a promise', () => {
@@ -47,7 +49,6 @@ describe('remoteFiles', () => {
       process.env.AWS_ACCESS_KEY_ID = 12
       remoteFiles.accessRemoteFile(config).catch(err => {
         expect(err).toBeInstanceOf(Error)
-        process.env = env
         done()
       })
     })
@@ -75,7 +76,6 @@ describe('remoteFiles', () => {
       config.s3Params.Key = 'this_is_not_valid'
       const response = remoteFiles.constructAwsRequest(config).catch(e => {
         expect(e)
-        process.env = env
         done()
       })
       expect(response).toBeInstanceOf(Promise)
@@ -84,7 +84,6 @@ describe('remoteFiles', () => {
       process.env.AWS_ACCESS_KEY_ID = 12
       remoteFiles.constructAwsRequest(config).catch(e => {
         expect(e)
-        process.env = env
         done()
       })
     })
--- a/utils/files/__tests__/remoteFiles.spec.js
+++ b/utils/files/__tests__/remoteFiles.spec.js
@@ -245,20 +245,6 @@ describe('remoteFiles', () => {
     })
   })
 
-  describe('safeParse', () => {
-    let json = { field: 'value' }
-    it('parses a json string', () => {
-      const jsonString = JSON.stringify(json)
-      const parse = remoteFiles.safeParse(jsonString)
-      expect(parse.field).toEqual('value')
-    })
-    it('returns false when the input is not proper json', () => {
-      const jsonString = 'this is not json'
-      const parse = remoteFiles.safeParse(jsonString)
-      expect(parse).toEqual(false)
-    })
-  })
-
   describe('tryRemote', () => {
     it('should resolve with the results of accessRemoteFile', done => {
       remoteFiles.getSingleRemoteInfo = jest.fn()
--- a/utils/files/__tests__/remoteFiles.spec.js
+++ b/utils/files/__tests__/remoteFiles.spec.js
@@ -1,0 +1,289 @@
+// require('cross-fetch/polyfill')
+const assert = require('chai').assert
+// const expect = require('expect')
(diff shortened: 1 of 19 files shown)


olgn:
please ignore codeclimate telling me i have _too many tests_ -psh

utils/files/__tests__/remoteFiles.spec.js
```diff
@@ -0,0 +1,287 @@
...
+        expect(e).toHaveProperty('code')
+        expect(e.code).toEqual(28)
+      }
+    })
+  })
+
+  describe('callGitAnnex', () => {
+    it('should return the string result of execSync', () => {
+      child_process.execSync = jest.fn()
+      child_process.execSync.mockReturnValue(new Buffer('test'))
```
nellh: Modifying this global will break other tests which use it.

utils/files/remoteFiles.js
```diff
@@ -0,0 +1,212 @@
...
+
+  // Ask git-annex for more information about a file
+  getRemotesInfo: function(dir, file) {
+    // Remove leading slash from  relativePath
+    const relativePath =
+      file.relativePath && file.relativePath.startsWith('/')
+        ? file.relativePath.substring(1)
+        : file.relativePath
+    const getRemoteCmd = `cd ${dir}
+    git show git-annex:$(git-annex examinekey --json $(git-annex lookupkey ${relativePath}) | jq -r .hashdirlower)$(git-annex lookupkey ${relativePath}).log.rmet`
```
nellh: This should be split up so that it doesn't have external dependencies other than git-annex or depend on a specific shell, it was meant as an example of how to do this not the final implementation.

utils/files/remoteFiles.js
```diff
@@ -0,0 +1,212 @@
...
+      }
+    })
+    return remotesInfo
+  },
+  // Check if a local directory is a git-annex repo
+  isGitAnnex: function(path) {
+    return fs.existsSync(path + '/.git/annex')
+  },
+  // Parse JSON from string if it's valid, else return false
+  safeParse: function(string) {
```
nellh: This doesn't look like it's called anywhere? Maybe it can be eliminated.

utils/files/remoteFiles.js
```diff
@@ -0,0 +1,212 @@
...
+      ? Object.keys(process.env).indexOf('AWS_ACCESS_KEY_ID') > -1
+      : false
+    if (hasCreds) {
+      return s3
+        .getObject(config.s3Params)
+        .promise()
+        .then(data => data.Body)
+    } else {
+      let url = this.constructAwsUrl(config)
+      return fetch(url).then(resp => resp.buffer())
```
nellh: s3.getObject and fetch work differently for HTTP error handling here. getObject will reject the promise if an HTTP error is encountered but fetch will resolve the promise. The fetch version needs a guard like this:

```javascript
fetch(url).then(resp => {
  if (resp.ok) {
    return resp.buffer()
  } else {
    return Promise.reject(
      new Error(`HTTP response failed - ${resp.status} - ${resp.statusText}`),
    )
  }
})
```

This will make them equivalent and avoid handling this in any callers.

utils/files/remoteFiles.js
```diff
@@ -0,0 +1,212 @@
...
+    const config = {
+      file: file,
+      dir: dir,
+    }
+    if (limit) config.limit = limit
+    config.remotesInfo = this.getRemotesInfo(dir, file)
+
+    // try all the special git-annex remotes, and exit if there is an issue (reading / fetching files)
+    // if all remotes fail, throw issue code 97
+    Promise.all(
```
nellh: Nothing happens with this Promise.all call, so it's either not needed or not being returned properly.

utils/files/__tests__/remoteFiles.spec.js
```diff
@@ -0,0 +1,287 @@
...
+  beforeAll(() => {
+    // fetch mock
+    global.fetch = jest
+      .fn()
+      .mockImplementation(
+        () => new Promise(resolve => resolve({ buffer: () => 'buffer' })),
+      )
+  })
+
+  describe('getAnnexedFile', () => {})
```
nellh: This is extra.

utils/files/__tests__/remoteFiles.spec.js
```diff
@@ -0,0 +1,287 @@
...
+    it('should return the issue of extractGzipBuffer if unzip is unsuccessful', done => {
+      config.file.name = 'something.gz'
+      remoteFiles.accessRemoteFile(config).catch(issue => {
+        expect(issue).toHaveProperty('code')
+        config.file.name = 'something'
+        done()
+      })
+    })
+    it('should return the error of constructAwsRequest if unsuccessful', done => {
+      process.env.AWS_ACCESS_KEY_ID = 12
```
nellh, line 47: Might be good to reset proccess.env as well to avoid conflicts with other tests.

nellh: Oh, looks like it is done. But I think this should be handled with something like beforeEach and afterEach so that's clearly defined.

nellh (changes requested):

utils/files/__tests__/remoteFiles.spec.js
olgn: its also completely unnecessary -_-

utils/files/__tests__/remoteFiles.spec.js
```diff
@@ -19,11 +17,14 @@ describe('remoteFiles', () => {
     global.fetch = jest
       .fn()
       .mockImplementation(
-        () => new Promise(resolve => resolve({ buffer: () => 'buffer' })),
+        () =>
+          new Promise(resolve => resolve({ ok: true, buffer: () => 'buffer' })),
```
nellh: This is equivalent to:

```javascript
() => Promise.resolve({ok: true, buffer: () => 'buffer'});
```

nellh (approved):

Issue #454, opened 2018-04-12
Raise and error if empty files (0kb) are present


chrisgorgo:
Spotten in the wild here https://github.com/poldracklab/bids-core/pull/22#issuecomment-380975215

PR #645, opened 2018-11-05
Error for empty files

david-nishi:
fix for issue #454

Notable changes: 
- added new error code 98: EMPTY_FILE
- added code to push error in testFile() (already called on config, bval, bvec, JSON, and tsv files)
- added misc group to files in fullTest to collect everything not run through testFile()
- added validateMisc(), which runs misc files through testFile()


utils/files/testFile.js
```diff
@@ -5,34 +5,47 @@ const Issue = require('../../utils/issues').Issue
...
+          if (!accessErr) {
+            process.nextTick(function() {
+              if (stats.size === 0) {
+                resolve({
+                  issue: new Issue({
+                    code: 97,
+                    file: file,
+                    reason: `Empty files (${
+                      file.path
+                    }) not allowed; may corrupt archives.`,
```
nellh: The "may corrupt archives" part was due to a SciTran bug, not relevant to the validator.

nellh (changes requested):
This should use a smaller example dataset or the test data should go in bids-examples, this adds a lot of files to the repo for one test suite.

david-nishi:
working on merge conflict in testFile.js

utils/files/testFile.js
david-nishi: removed "may corrupt archives" and minimized test file

nellh (approved):

PR #649, opened 2018-11-12
update EMPTY_FILE error code where thrown to 99

david-nishi:


olgn:
@DaNish808 apparently the unit tests also referred to the incorrect error code. You will need to update these to get tests passing.

Issue #592, opened 2018-10-01
enable the eeg Extension (BEP006)


sappelhoff:
Similar to issue bids-standard/legacy-validator#510 concerning iEEG, we can list all the remaining To Do points for enabling the EEG validator by default (i.e., without passing the `--bep006` flag).

This list is work in progress. If you find more to dos, please comment and I can update this list. 
cc: @robertoostenveld @CPernet @arnodelorme @gllmflndn @ChristophePhillips

# To Do:
- [x] Go through [_eeg.json schema](https://github.com/bids-standard/bids-validator/blob/master/validators/json/schemas/eeg.json) and bring up to date ... see bids-standard/legacy-validator#661 
    - [x] remove field `DeviceSoftwareVersion`, which has been replaced by `SoftwareVersions`
    - [x] discuss the required format for `HardwareFilters` and `SoftwareFilters` ... also: make them required?
- [x] Go through code coverage report and make sure that all EEG related features are covered by tests
- [x] Discuss whether a non-custom column `event_types` should be included in BIDS-EEG. If yes, it should be added here: `bids_validator/tsv/non_custom_columns.json`. See [BEP spec google doc](https://docs.google.com/document/d/1ArMZ9Y_quTKXC-jNXZksnedK2VHHoKP3HCeO5HPcgLE/edit#heading=h.9rets2uzyf3s) comment for why it might be controversial ... see also:  bids-standard/legacy-validator#599
- [x] ~Discuss, whether the `name` column in `electrodes.tsv` and `channels.tsv` HAS to correspond and if yes, how to check it~ according to our definitions, they do not correspond... e.g., there can be two electrodes forming a single channel.
- [x] remove support for Neuroscan CNT
- [ ] think about introducing a check that parses the `channels.tsv` file and counts the number of EEG, EOG, EMG, etc. channels ... then compares this count that is given in the `*_eeg.json` ... check for validity - raise error if no assertion ... see bids-standard/bids-validator#80
- [x] solve bids-standard/legacy-validator#601 
- [ ] solve bids-standard/bids-validator#81 
- [x] solve bids-standard/legacy-validator#477 
- [ ] incorporate a check: if `electrodes.tsv` is included, there MUST be a `coordsystem.json` included as well. see bids-standard/legacy-validator#672 
[...]

PR #661, opened 2018-11-28
[MRG] bring eeg json schema up to date

sappelhoff:
related to #592 

@choldgraf I saw that in the [ieeg.json schema](https://github.com/bids-standard/bids-validator/blob/master/validators/json/schemas/ieeg.json) you have `"additionalProperties": true`, whereas it is false in `meg`, `eeg`, and the coordsystem definitions. Having it at false allows for a tight control during validation ... at the expense that users cannot define their own fields in the modality specific json file.

What is more desirable in this case?

cc @robertoostenveld

### DIFF
--- a/tests/json.spec.js
+++ b/tests/json.spec.js
@@ -81,10 +81,10 @@ describe('JSON', function() {
       TaskName: 'rest',
       SamplingFrequency: 1000,
       EEGChannelCount: 1,
-      EOGChannelCount: 2,
-      ECGChannelCount: 3,
-      EMGChannelCount: 4,
       EEGReference: 'Cz',
+      SoftwareFilters: {
+        HighPass: { HalfAmplitudeCutOffHz: 1, RollOff: '6dB/Octave' },
+      },
       PowerLineFrequency: 50,
     }
     jsonDict[eeg_file.relativePath] = jsonObj
--- a/validators/json/schemas/eeg.json
+++ b/validators/json/schemas/eeg.json
@@ -12,7 +12,6 @@
     "Manufacturer": { "type": "string", "minLength": 1 },
     "ManufacturersModelName": { "type": "string", "minLength": 1 },
     "DeviceSerialNumber": { "type": "string" },
-    "DeviceSoftwareVersion": { "type": "string" },
     "SoftwareVersions": { "type": "string" },
     "PowerLineFrequency": { "type": "number" },
     "SamplingFrequency": { "type": "number" },
@@ -27,6 +26,7 @@
     "EEGGround": { "type": "string" },
     "CapManufacturer": { "type": "string" },
     "CapManufacturersModelName": { "type": "string" },
+    "HeadCircumference": { "type": "number" },
     "HardwareFilters": {
       "anyOf": [
         { "type": "object", "additionalProperties": { "type": "object" } },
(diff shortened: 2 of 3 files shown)


choldgraf:
I do think that in our case, we set it to "true" intentionally, since it felt like it's more important that the required fields be there, than that we prohibit other fields from being there. However, I don't know that @dorahermes and I really thought it through in-depth. Is there a big downside to having extra metadata in these files?

sappelhoff:
>  Is there a big downside to having extra metadata in these files?

One downside I can think of is, that typos will be accepted, imagine the following:

```json
{
  "SoftwareFillters": "Lowpass"
}
```

There is a typo in the `"SoftwareFilters` key, thus the JSON schema does not recognize it and does not require the formatting according to the schema ...it is recognized as an "additionalProperty" and hence we end up with a mistake that our validator cannot catch. The mistake being here, that SoftwareFilters should be a dictionary of dictionaries or a string `"n/a"` ... but nothing else.

Note that this affects only fields that are not `REQUIRED`, because for a required field, you would get the message: "no key SoftwareFilters recognized, please add it" ... and then you would notice that your "SoftwareFilters" are faultily spelled with a double L (and then you correct it and the schema will recognize the key and complain that your formatting of this field is wrong).

As for the benefits of allowing "additionalProperties": Wouldn't users be able to simply define fields not covered in the spec in e.g., the README file?

dorahermes:
Does this only apply to .json files, or also to .tsv files? 

For example, in the _events.tsv files we commonly want to be able to specify other event attributes, and for the _electrodes.tsv file we commonly want to specify electrode labels from atlases that are not yet in the spec. Putting all these in the readme will not work. 

I agree that spelling issues can create problems. Could there perhaps be a warning for added fields (that are not recommended or optional), like 'note that you added additional properties not specified in BIDS:...'

sappelhoff:
> Does this only apply to .json files, or also to .tsv files?

It only applies to the modality specific sidecar JSON. The TSV files can have arbitrary columns added, for example "myDummY" in events.tsv ... however, then there will need to be a "myDummy" key in the accompanying events.json file --> so this means that for all JSON files EXCEPT the modality specific json files (eeg.json, ieeg.json, meg.json, ...) we do already accept "additionalProperties" --> it is just less obvious, because we do not have JSON schemas for these json files.

> Could there perhaps be a warning for added fields (that are not recommended or optional), like 'note that you added additional properties not specified in BIDS:...'

This is a very good idea ... with such a warning in place I would have no qualms for setting `"additionalProperties": True` for the EEG schema as well. But given what I said above, that we actually already accept additionalProperties in most of our JSON files, perhaps we can set it true everywhere already ... and improve the situation further by making a PR introducting the warning that Dora suggests.

dorahermes:
I see. For the _ieeg.json file I would be ok with having `additionalProperties:false` and storing them in the readme or an _scans.tsv file or so.

sappelhoff:
What is your opinion @choldgraf ? Should I set the ieeg `additionalProperties` to `false` in this PR as well? And then we can open a new issue to remind us that we implement Dora's suggested warning and once it's there change `additionalProperties` back to `true`

choldgraf:
I'm +1 to setting `additionalProperties` to false in this PR, and then beginning the process of discussing whether we need to make it allowed but raise a warning if this happens.

sappelhoff:
okay great, then this is ready to be merged! :-) Feel free to do so @choldgraf or @dorahermes

choldgraf:
I apparently don't have permissions to merge, but I would if I did! :-P

cc @chrisfilo :-)

chrisgorgo:
You do now!

PR #663, opened 2018-11-29
[MRG] QuickTest and regexp fixes

### LINKED ISSUE #477: Inheritance principle not taken into consideration by validator

### LINKED ISSUE #601: [BUG] sidecar not part of BIDS spec ... although it is
I encountered an error that looks like a bug:

example: for [eeg_matchingpennies](https://github.com/bids-standard/bids-examples/tree/bep006_eeg/eeg_matchingpennies), try to define a file: `eeg_matchingpennies/sub-05/eeg/sub-05_task-matchingpennies_events.json` and run the validator with the `--bep006` flag.

--> This should not yield an error, because it is perfectly fine to have a .json sidecar for a .tsv datafile at any directory level.

sappelhoff:
@chrisfilo based on your suggestion I went through the RegExps again and indeed: I found a pretty crucial typo (see https://github.com/bids-standard/bids-validator/pull/663/files#r237586297)

... through trial and error, this PR also ...

fixes #601 
fixes #477

### DIFF
--- a/bids_validator/rules/file_level_rules.json
+++ b/bids_validator/rules/file_level_rules.json
@@ -90,6 +90,7 @@
     "tokens": {
       "@@@_eeg_type_@@@": ["vhdr", "vmrk", "eeg", "edf", "bdf", "set", "fdt"],
       "@@@_eeg_ext_@@@": [
+        "_events.json",
         "_events.tsv",
         "_electrodes.tsv",
         "_channels.tsv",
@@ -178,6 +179,7 @@
         "mef"
       ],
       "@@@_ieeg_ext_@@@": [
+        "_events.json",
         "_events.tsv",
         "_electrodes.tsv",
         "_channels.tsv",
@@ -193,6 +195,7 @@
     "tokens": {
       "@@@_meg_type_@@@": [".fif", ".ds"],
       "@@@_meg_ext_@@@": [
+        "_events.json",
         "_events.tsv",
         "_channels.tsv",
         "_meg.json",
--- a/bids_validator/rules/session_level_rules.json
+++ b/bids_validator/rules/session_level_rules.json
@@ -65,7 +65,7 @@
   "eeg_ses": {
     "regexp": "^\\/(sub-[a-zA-Z0-9]+)\\/(?:(ses-[a-zA-Z0-9]+)\\/)?\\1(_\\2)?(?:_task-[a-zA-Z0-9]+)?(?:_acq-[a-zA-Z0-9]+)?(?:_proc-[a-zA-Z0-9]+)?(@@@_eeg_ses_type_@@@)$",
     "tokens": {
-      "@@@_anat_eeg_type_@@@": [
+      "@@@_eeg_ses_type_@@@": [
         "_events.tsv",
         "_channels.tsv",
         "_electrodes.tsv",
--- a/tests/type.spec.js
+++ b/tests/type.spec.js
@@ -219,8 +219,8 @@ describe('utils.type.file.isEEG', function() {
   ]
 
   goodFilenames.forEach(function(path) {
-    it("isEeg('" + path + "') === true", function(isdone) {
-      assert.equal(utils.type.file.isEeg(path), true)
+    it("isEEG('" + path + "') === true", function(isdone) {
+      assert.equal(utils.type.file.isEEG(path), true)
       isdone()
     })
   })
@@ -232,8 +232,8 @@ describe('utils.type.file.isEEG', function() {
   ]
 
   badFilenames.forEach(function(path) {
-    it("isEeg('" + path + "') === false", function(isdone) {
-      assert.equal(utils.type.file.isEeg(path), false)
+    it("isEEG('" + path + "') === false", function(isdone) {
+      assert.equal(utils.type.file.isEEG(path), false)
       isdone()
     })
   })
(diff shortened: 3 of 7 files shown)


bids_validator/rules/session_level_rules.json
```diff
@@ -65,7 +65,7 @@
   "eeg_ses": {
     "regexp": "^\\/(sub-[a-zA-Z0-9]+)\\/(?:(ses-[a-zA-Z0-9]+)\\/)?\\1(_\\2)?(?:_task-[a-zA-Z0-9]+)?(?:_acq-[a-zA-Z0-9]+)?(?:_proc-[a-zA-Z0-9]+)?(@@@_eeg_ses_type_@@@)$",
     "tokens": {
-      "@@@_anat_eeg_type_@@@": [
+      "@@@_eeg_ses_type_@@@": [
```
sappelhoff, line 68: note this typo ... I expected this fix to solve the other two issues named above - however, no effect. weird ...

tests/type.spec.js
```diff
@@ -84,6 +84,9 @@ describe('utils.type.file.isTopLevel', function() {
     '/T1w.json',
     '/acq-test_dwi.json',
     '/rec-test_physio.json',
+    '/task-testing_eeg.json',
+    '/task-testing_ieeg.json',
+    '/task-testing_meg.json',
```
sappelhoff, line 89: these three lines pass, so they are recognized as top level ... however defining only `/task-testing_eeg.json` in an EEG directory yields a failed QuickTest, suggesting that something is not really recognized.

sappelhoff:
trying to get closer to solving #477 

a failed QuickTest means that isBIDS fails:
https://github.com/bids-standard/bids-validator/blob/70062aa237e29047c81447f448f2e83723cf647a/utils/type.js#L67

... in the reported test cases, we get a failed QuickTest, if there is an eeg.json file only defined on the TopLevel ... if we define it at the subject level, everything is fine.

That could mean, that toplevel eeg.json files are not recognized as BIDS (as Chris suggested) ... however, I simply inserted these toplevel files into the tests, and they still pass - so I am at a loss again. see: 
https://github.com/bids-standard/bids-validator/pull/663/files#r237609244

sappelhoff:
I think I found the issue. It lies with :
https://github.com/bids-standard/bids-validator/blob/d7a00cce2920113d6021b1074271570e3b9a3227/utils/modalities.js#L51-L67


This will not return `True` for modality specific json files defined at the root directory ... because by being defined at the root directory, these files do not have a path that includes the modality such as `eeg`, `meg`, or `ieeg` ... and that is what we are testing for in the quicktest.

We can probably think of a better way of doing that.

tests/type.spec.js
sappelhoff: see: https://github.com/bids-standard/bids-validator/pull/663#issuecomment-442998486

sappelhoff:
#477 is now fixed ... this was done by tweaking the `isCorrectModality` a little bit:

Instead of only looking for json files and checking their path, we are now also looking for tsv files and check their path. If that path contains any modality (eeg, meg, ieeg) at any level, we assume it's BIDS:

`    } else if (['.json', '.tsv'].some(v => path[0].includes(v))) {`

This is not a perfect solution --> I imaging a resting state EEG dataset with multiple subjects and no `channels.tsv`, `electrodes.tsv`, or `events.tsv` defined at the subject/session level and only a single sidecar eeg.json defined at the root ... in that case, there would be the triggering of a QuickTestError.

But that case is arguably unlikely and the present solution is simple and straight forward. Most datasets will have *some* kind of tsv or json defined at a subject level.

I think we can improve this further once we get rid of the bep006 and bep010 flags. For now, I am ready to have this reviewed and potentially merged @chrisfilo

utils/modalities.js
```diff
@@ -48,7 +48,7 @@ module.exports = {
       ['anat', 'func', 'dwi'].indexOf(path[1]) != -1
     ) {
       isCorrectModality = true
-    } else if (path[0].includes('.json')) {
+    } else if (['.json', '.tsv'].some(v => path[0].includes(v))) {
```
sappelhoff, line 51: this fixes #477

bids_validator/rules/file_level_rules.json
```diff
@@ -90,6 +90,7 @@
     "tokens": {
       "@@@_eeg_type_@@@": ["vhdr", "vmrk", "eeg", "edf", "bdf", "set", "fdt"],
       "@@@_eeg_ext_@@@": [
+        "_events.json",
```
sappelhoff, line 93: this might actually be fixed with #654 as well ...

PR #669, opened 2018-12-08
BF: Checkout the release and adjust version BEFORE npm install and publish

### LINKED ISSUE #602: bleeding edge validator shows version 0.0.0
When using the bids-validator in the master repo and the /bin directory, it yields 0.0.0 as its version.

`bids-validator/bin/bids-validator --version`

Shouldn't this be 1.0.1.dev (whatever the current version is) or similar?

yarikoptic:
A semi-blind attempt to fix #602.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -52,12 +52,12 @@ jobs:
       - checkout
       - run: apk --no-cache add ca-certificates
       - run: apk --no-cache add git python build-base openssh
-      # Publish to NPM
-      - run: npm install
-      - run: ./node_modules/.bin/npm-cli-login -u $NPM_USERNAME -p $NPM_PASSWORD -e $NPM_EMAIL
       - run: git checkout -f $CIRCLE_TAG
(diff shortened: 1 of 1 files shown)


chrisgorgo:
The changes to `package.json` on circle are not pushed back to `master` so this will not work.

yarikoptic:
By this do you mean the PR itself or my changes?
FWIW, I have no authority over master or merging PRs

sappelhoff:
1. But wouldn't we be able to simply do a `git push origin master` like below?

https://github.com/bids-standard/bids-validator/blob/e48b55b001ce959bdaf3e2b021848df3948513c2/circle.yml#L74

2. we'd need to append a `-dev` or similar, because master is always ahead of the latest release

chrisgorgo:
@yarikoptic by "this" I meant your changes

@sappelhoff possibly, but please test any solution before sending a PR

PR #681, opened 2019-01-11
[MRG] remove optional arguments bep006 and bep010

### LINKED ISSUE #510: Enable the iEEG extension (BEP 010)

sappelhoff:
related to https://github.com/bids-standard/bids-specification/issues/107

closes #510 

this can be merged once https://github.com/bids-standard/bids-specification/issues/108 has been merged into the main BIDS specification.

### DIFF
--- a/bids-validator/bin/bids-validator
+++ b/bids-validator/bin/bids-validator
@@ -17,10 +17,6 @@ var argv = require('yargs')
     .describe('verbose', 'Log more extensive information about issues')
     .boolean('json')
     .describe('json', 'Output results as JSON')
-    .boolean('bep006')
-    .describe('bep006', 'Enable support for BIDS Extension Proposal 006')
-    .boolean('bep010')
-    .describe('bep010', 'Enable support for BIDS extension proposal 010')
     .option('config', {
         alias: 'c',
         describe: 'Optional configuration file. See https://github.com/bids-standard/bids-validator for more info'
--- a/bids-validator/tests/utils/issues.spec.js
+++ b/bids-validator/tests/utils/issues.spec.js
@@ -20,8 +20,6 @@ describe('issues', () => {
         ignoreWarnings: false,
         ignoreNiftiHeaders: false,
         verbose: false,
-        bep006: false,
-        bep010: false,
         config: {},
       }
       formattedIssues = utils.issues.exceptionHandler(
--- a/bids-validator/utils/modalities.js
+++ b/bids-validator/utils/modalities.js
@@ -40,7 +40,7 @@ module.exports = {
     return modalities
   },
 
-  isCorrectModality: (path, options) => {
+  isCorrectModality: path => {
     let isCorrectModality = false
     // MRI
     if (
@@ -57,11 +57,11 @@ module.exports = {
           break
         case 'eeg':
           // EEG
-          isCorrectModality = !!options.bep006
+          isCorrectModality = true
           break
         case 'ieeg':
           // iEEG
-          isCorrectModality = !!options.bep010
+          isCorrectModality = true
           break
         default:
           break
--- a/bids-validator/utils/options.js
+++ b/bids-validator/utils/options.js
@@ -11,8 +11,6 @@ module.exports = {
       ignoreWarnings: options.ignoreWarnings ? true : false,
       ignoreNiftiHeaders: options.ignoreNiftiHeaders ? true : false,
       verbose: options.verbose ? true : false,
-      bep006: options.bep006 ? true : false,
-      bep010: options.bep010 ? true : false,
       config: options.config ? options.config : {},
     }
     if (options.config && typeof options.config !== 'boolean') {
--- a/bids-validator/utils/summary/collectSessions.js
+++ b/bids-validator/utils/summary/collectSessions.js
@@ -6,7 +6,7 @@ const collectSessions = (fileList, options) => {
     const file = fileList[key]
     if (
       !type.file.isStimuliData(file.relativePath) &&
-      type.isBIDS(file.relativePath, options.bep006, options.bep010)
+      type.isBIDS(file.relativePath)
     ) {
       const pathValues = type.getPathValues(file.relativePath)
       const isEmptyRoom = pathValues.sub && pathValues.sub == 'emptyroom'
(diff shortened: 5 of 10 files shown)


chrisgorgo:
What's the situation for test for those beps in the validator? Do they exist? HAve they been running in CI already?

validators/bids/quickTest.js
```diff
@@ -8,7 +8,7 @@ const utils = require('../../utils')
  * otherwise it will throw a callback with a
  * generic error.
  */
-const quickTest = (fileList, options) => {
```
chrisgorgo: I think instead of ripping this out it would be better to refactor it to easily enable future "work in progress" validator options related to BEP. This surely be needed in the future.

chrisgorgo (changes requested):

sappelhoff:
> What's the situation for test for those beps in the validator? Do they exist? Have they been running in CI already?

We have a comprehensive set of tests for both EEG and iEEG, see for example:

- https://github.com/bids-standard/bids-validator/blob/bdec39565eb13dab5e0dd6381975fac85020ec2a/tests/type.spec.js#L207-L279
- https://github.com/bids-standard/bids-validator/blob/57ee2a8e2273d3266ef77e8ffbaa5b9c88072a9f/tests/tsv.spec.js#L256-L436
- https://github.com/bids-standard/bids-validator/blob/bdec39565eb13dab5e0dd6381975fac85020ec2a/tests/json.spec.js#L74-L171

yes, they have been running in CI. With the `--bep` flags we only prevented a validator run to start if only the bep modalities are present ... in the tests however, we call functions directely, so the `--bep` flags do not have an influence on whether a test gets run or not (mostly)

validators/bids/quickTest.js
sappelhoff: that is a good point. I'll make a suggestion that you can evaluate.

utils/type.js
```diff
@@ -74,8 +74,8 @@ module.exports = {
       this.file.isDWI(path) ||
       this.file.isFunc(path) ||
       this.file.isMeg(path) ||
-      (this.file.isIEEG(path) && bep010) ||
-      (this.file.isEEG(path) && bep006) ||
+      this.file.isIEEG(path) ||
+      this.file.isEEG(path) ||
```
sappelhoff: we'd have to insert some documentation that THIS is where the BEP flag should be, if somebody wants to add a new modality

utils/modalities.js
```diff
@@ -57,11 +57,15 @@ module.exports = {
...
+          isCorrectModality = true
           break
         case 'ieeg':
           // iEEG
-          isCorrectModality = !!options.bep010
+          isCorrectModality = true
+          break
+        // Add new potential modality under 'BEP_placeholder'
+        case 'BEP_placeholder':
+          isCorrectModality = !!options.bepWIP
```
sappelhoff: instead of this "placeholder", we can also write the instructions into some documentation (see also [comment below](https://github.com/bids-standard/bids-validator/pull/681/files#r247422081))

sappelhoff:
apart from the fact that `ES-lint` doesn't like my style (and neither does codeclimate) ( `67:26  error  'bepWIP' is defined but never used  no-unused-vars`) ... what do you think of this approach @chrisfilo ?

Do you have something different in mind?

sappelhoff:
looking back at this a few weeks later, it does not look nice at all. I am not sure about refactoring this structure of allowing BEP validation with the main validator.

I remember that we made this special case for EEG and iEEG because we had separate branches for them that were working nicely but then a major refactoring of the bids-validator which was supposed to happen at that time would have destroyed our work: that's why we integrated it into the main validator.

For future BEP validators I would recommend making a branch instead of this option feature that we have currently.

What do you think @chrisfilo ?

chrisgorgo:
Could you rebase/merge master and make sure all the test pass? I hope we could merge this soon.

sappelhoff:
before doing more work on this, could you comment on what I wrote here: https://github.com/bids-standard/bids-validator/pull/681#issuecomment-459954799 ?

chrisgorgo:
It's fine to remove optional validation functionality (that future beps could potentially benefit), if this is what you are asking. The valuator indeed went through a lot of changes.

sappelhoff:
I am not sure about that codeclimate issue, but these changes should suffice to make BEP006 and BEP010 part of the validator.

chrisgorgo (approved):

chrisgorgo:
Thanks! Could you check the web version https://2255-37161308-gh.circle-artifacts.com/0/root/web_version/index.html against an example EEG dataset?

sappelhoff:
> Thanks! Could you check the web version https://2255-37161308-gh.circle-artifacts.com/0/root/web_version/index.html against an example EEG dataset?

done and it's looking good :+1:

the issue from #704 is also visible with the EEG data ... and /stimuli files also get listed as modalities

sappelhoff:
https://github.com/bids-standard/bids-specification/pull/108 has been merged, so this PR is ready to be merged as well :-)

:shipit:

chrisgorgo:
will merge as soon as 1.2.0 is released - thank you for your hard work

PR #685, opened 2019-01-11
Remove isModality rule

chrisgorgo:
This rule did not describe valid BIDS files and turned out it was not necessary.


sappelhoff (approved):
yes, `isModality` seems superfluous. +1 to remove it 

We'll remain with the admissible file extensions only in the modality specific regexp groups

PR #686, opened 2019-01-11
Escape dots in regural expressions

chrisgorgo:
Dots mean "any character" if not escaped. This made the regular expression too flexible.


sappelhoff (approved):
good catch!

Issue #675, opened 2018-12-15
Error 44 on good files


ins0mniac2:
Following up on the thread https://github.com/bids-standard/bids-validator/issues/671 ,
version 1.1.1 continues to have this issue. Certain files that look good and is viewable in a viewer like ITK-SNAP (see attached) are flagged as

"We were unable to read this file. Make sure it contains data (fileSize > 0 kB) and is not corrupted, incorrectly named, or incorrectly symlinked."

chrisgorgo:
Could you share the files?

chrisgorgo:
Or at least one example

dorianps:
We can't share the original file, unfortunately. 

@sandhitsu can you empty the image (zero voxels) and check it again. If `bids-validator` still complains, we can share with @chrisfilo  the empty image.

dorianps:
If the empty file resolves the problem, than the problem is the original file. We can share the original file, but only privately with @chrisfilo (not online), and we need to remove the subject ID to some random number. We need to respect the data agreements to the letter, unfortunately.

ins0mniac2:
Zeroing out the image does not solve the problem. Here is the zero image.
[sub-XXXX_ses-01_run-01_T1w.nii.gz](https://github.com/bids-standard/bids-validator/files/2682945/sub-XXXX_ses-01_run-01_T1w.nii.gz)

chrisgorgo:
Could not replicate:
```
me@christop ~/Downloads $ node --version                                                                                             v11.4.0                                                                                                                              
me@christop ~/Downloads $ bids-validator --version                                                                                   1.1.1                                                                                                                               
me@christop ~/Downloads $ bids-validator test_ds                                                                                   
This dataset appears to be BIDS compatible.                                                                                                  
Summary:                Available Tasks:        Available Modalities:                                                                2 
Files, 51.94KB                                T1w                                                                                  1 - Subject                                                                                                                          
1 - Session                                                                                                                                                                                                                                                       
If you have any questions please post on https://neurostars.org/tags/bids
--
```
[test_ds.zip](https://github.com/bids-standard/bids-validator/files/2682956/test_ds.zip)

Perhaps a permission issues or something to do with a network drive?

PS Storing personal identifiable information in participant labels is a bad and dangerous practice. I edited your original post to remove the screenshot which included the participant label.

ins0mniac2:
[...]

PR #687, opened 2019-01-13
Prevent out-of-memory error

david-nishi:
Closes #677 

* [x] partial-load files for 0kb check
  * [x] in node
  * [x] in browser
* [x] limit concurrency in validators

PR #691, opened 2019-01-17
Close nifti files #675

david-nishi:
* [x] close all nifti files streams after getting header

### DIFF
--- a/utils/files/readNiftiHeader.js
+++ b/utils/files/readNiftiHeader.js
@@ -46,18 +46,18 @@ function extractNiftiFile(file, callback) {
   const bytesRead = 500
   const buffer = Buffer.alloc(bytesRead)
 
-  var decompressStream = zlib
-    .createGunzip()
-    .on('data', function(chunk) {
-      callback(parseNIfTIHeader(chunk, file))
-      decompressStream.pause()
-    })
-    .on('error', function() {
-      callback(handleGunzipError(buffer, file))
-    })
-
   fs.open(file.path, 'r', function(err, fd) {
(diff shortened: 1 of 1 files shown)


nellh:
I've been testing this with a hard limit of 64 or 128 files and I am not seeing a difference between this PR and master for reproducing the issue from #675. Does the structure of the dataset matter for testing this? I'm using a dataset with 147 nifti files.

david-nishi:
@nell i believe it only runs fs.open on annexed files, if that's related

nellh:
I think this isn't needed with the concurrency limiting introduced in #687

yarikoptic:
FWIW, I have ran

```
while true; do nl=$(ls -l /proc/59167/fd 2>/dev/null | wc -l ); if [ $nl -gt 100 ]; then echo "Got $nl files. I will list a sleep a little to rest"; ls -l /proc/59167/fd; sleep 10; fi; done 2>&1 | tee /tmp/toomany-open
```

and eventually I got "Got 4097 files. I will list a sleep a little to rest" and indeed it was lots of annex'ed .nii.gz files:

```
[bids@rolando tmp] > tail toomany-open
...
lr-x------ 1 bids bids 64 Jan 12 15:53 999 -> /usr/local/singularity_v2.5.1/var/singularity/.../.git/annex/objects/08/XJ/MD5E-s533081--e5095d4b869a241e06664d81bb8288e7.nii.gz/MD5E-s533081--e5095d4b869a241e06664d81bb8288e7.nii.gz
```

so may be this PR should be re-opened/re-considered?

PR #693, opened 2019-01-19
FIX: PyPI release

effigies:


circle.yml
```diff
@@ -71,24 +71,21 @@ jobs:
       - checkout
       - run: apk --no-cache add ca-certificates
       - run: apk --no-cache add git python build-base openssh
+      - run:
+          name: Install Python 3 and update setuptools
+          command: |
+            apk --no-cache add python3 python3-dev py3-cffi
+            python3 -m pip install --upgrade pip setuptools wheel twine
+      - run: git checkout -f $CIRCLE_TAG
+      - run: python3 setup.py sdist bdist_wheel
```
effigies, line 80: In order to properly build the package with the tag version number, it needs to be done from a clean directory, so I've moved this above the `npm` stuff. I've left the actual push to PyPI for after.

I can also move it to a separate deployment job, if that keeps things simpler.

setup.cfg
```diff
@@ -7,7 +7,7 @@ maintainer = BIDS Developers
 maintainer_email = bids-discussion@googlegroups.com
 description = Validator for the Brain Imaging Data Structure
 long_description = file:README.md
-long_description_content_type="text/markdown"
+long_description_content_type = text/markdown
```
effigies, line 10: In uploading manually I discovered a bug in the config.

PR #695, opened 2019-01-28
Fix nodejs only unhandled error crash for empty files

nellh:
Empty file error code 99 was being hidden by the generic error handler here, leading to a difference in behavior between the web and nodejs versions of the validator. If the exceptionHandler receives an Issue instead of an error this will report the correct issue instead of the internal error warning.


olgn (approved):
* tested the cli with a dataset with one empty file - before this branch, returned internal error. with this bugfix, returns error 44 (same as web validator)

PR #701, opened 2019-02-02
[MRG] add brainvision-validator to validate BrainVision data triplets

### LINKED ISSUE #475: Integrate validation of BrainVision data files

sappelhoff:
closes #475 

With the EEG extension soon being merged, I want to get started with validating the raw data files.

The BrainVision data format is one of the two core formats of the EEG extension, and using the [`brainvision-validator`](https://github.com/sappelhoff/brainvision-validator) npm module, this should be possible within the bids-validator.

I would greatly appreciate some guidance through the process, or reasons why we should take a different approach and close this PR instead :-)

For the beginning, I have added a snippet of code that would make sense IMO.

Questions:
1. the `validateBrainVision` currently returns an array of strings ... if the array is empty, the validation was successful ... else there are the strings that describe the problem --> How would I best transform these potential strings into the "issues" object used in BIDS-validator?
2. What else should I prepare to get this PR merged?

### DIFF
--- a/utils/issues/list.js
+++ b/utils/issues/list.js
@@ -558,6 +558,6 @@ module.exports = {
     key: 'BRAINVISION_LINKS_BROKEN',
     severity: 'error',
     reason:
-      'Internal file pointers in BrainVision file triplet (*.eeg, *.vhdr, and *.vmrk) broken or some files do not exist.',
+      'Internal file pointers in BrainVision file triplet (*.eeg, *.vhdr, and *.vmrk) are broken or some files do not exist.',
   },
 }
--- a/utils/issues/list.js
+++ b/utils/issues/list.js
@@ -558,6 +558,6 @@ module.exports = {
     key: 'BRAINVISION_LINKS_BROKEN',
     severity: 'error',
     reason:
-      'The internal file pointers between the present BrainVision file triplet (*.eeg, *.vhdr, and *.vmrk) appear to be broken or some files do not exist.',
+      'Internal file pointers in BrainVision file triplet (*.eeg, *.vhdr, and *.vmrk) broken or some files do not exist.',
   },
 }
--- a/utils/issues/list.js
+++ b/utils/issues/list.js
@@ -554,4 +554,10 @@ module.exports = {
     severity: 'error',
     reason: 'Empty files not allowed.',
   },
+  100: {
+    key: 'BRAINVISION_LINKS_BROKEN',
+    severity: 'error',
+    reason:
+      'The internal file pointers between the present BrainVision file triplet (*.eeg, *.vhdr, and *.vmrk) appear to be broken or some files do not exist.',
+  },
 }
--- a/validators/bids/fullTest.js
+++ b/validators/bids/fullTest.js
@@ -83,10 +83,20 @@ const fullTest = (fileList, options, annexed, dir, callback) => {
   var vhdrIssues = []
   vhdrFiles.forEach(function(vhdrFile) {
     var issues = validateBrainVision(vhdrFile.path)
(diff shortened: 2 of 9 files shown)


validators/bids/fullTest.js
```diff
@@ -77,6 +78,15 @@ const fullTest = (fileList, options, annexed, dir, callback) => {
...
   }
 
+  // Validate BrainVision EEG files (file triplets: .eeg, .vhdr, .vmrk)
+  var vhdrFiles = files.ephys.filter(file => file.name.endsWith('vhdr'))
+  var vhdrIssues = []
+  vhdrFiles.forEach(function(vhdrFile) {
+    var issues = validateBrainVision(vhdrFile.path)
+    vhdrIssues = vhdrIssues.concat(issues)
+  })
+  self.issues = self.issues.concat(vhdrIssues)
```
sappelhoff, line 103: currently this yields an error when testing locally, because I am concatenating an array of strings to `self.issues`, instead of an array of issues (see questions in OP)

chrisgorgo:
1. Why not return an array of Issues?
2. The tests need to pass. If you think the current failures are not related to your change try rebasing.

sappelhoff:
>     1. Why not return an array of Issues?

Would I have to create a new "issue type" and add it to the list in `issues/list.js` ?

e.g.,
```
  100: {
    key: 'BRAINVISION_NOT_LINKED',
    severity: 'error',
    reason: 'Internal links are broken for files (.eeg, .vhdr, .vmrk). Check DataFile and MarkerFile fields in .vhdr and .vmrk"',
  },
```

... and should I create several of these - one for each issue one can have with the BrainVision files?

chrisgorgo:
Yup!

sappelhoff:
@chrisfilo should be good to go now, if you have time for a review.

utils/issues/list.js
```diff
@@ -554,4 +554,10 @@ module.exports = {
     severity: 'error',
     reason: 'Empty files not allowed.',
   },
+  100: {
+    key: 'BRAINVISION_LINKS_BROKEN',
+    severity: 'error',
+    reason:
+      'Internal file pointers in BrainVision file triplet (*.eeg, *.vhdr, and *.vmrk) broken or some files do not exist.',
```
chrisgorgo:
```suggestion
      'Internal file pointers in BrainVision file triplet (*.eeg, *.vhdr, and *.vmrk) are broken or some files do not exist.',
```

chrisgorgo:
Thanks!

PR #710, opened 2019-02-11
Consolidate browser and node builds

### LINKED ISSUE #595: Only push build artifacts to GitHub pages
Right now the validator is built with browserify and grunt on releases by combining the web version in the gh-pages branch and importing the package from the master branch. If we made a few changes, we could do this build on the master branch and only push the built artifacts to gh-pages. There's a few advantages of this approach.

1. It would remove the need to split the web validator development into patches across two branches.
1. The browser version would be easier to test for a given change (no "did I get the right version into my github-pages build?") since you wouldn't need to mess around with two branches.
1. New features like #593 would be easier to implement if the libraries were maintained in the main branch and tests could be written for the web components along with the main test suite.
1. It is more discoverable for new contributors if it is on the main branch.

What needs to happen for this:

- [ ] Need to move the existing app to the master branch.
- [ ] Update the web version to current dependencies should happen, many of the web validator dependencies are very out of date and would need to be consolidated with the main branch.
- [ ] Switch to a more modern bundler (Rollup or Webpack 4). Might be optional but could simplify this setup.

olgn:
fixes #595

* the web browser demo now lives in the master branch (/bids-validator-web)
* the validator itself has been moved to a subdirectory in the master branch (/bids-validator)
* the web browser demo is updated to use the current version of react and is build using next.js
* development and testing for the web browser demo can now be implemented without switching branches
* developers can run `yarn web-start` to get the development server up and running on `localhost:3000`
* developers can run `yarn web-export` to see if there are any build errors locally before pushing
* in development mode, the web browser will utilize the bids-validator from the current working directory - no more package linking issues!
* deployment now consists of pushing build artifacts to gh-pages branch


nellh (commented):
I have one concern with this setup after testing it and that is it's pretty easy to end up with mismatched or confusing dependencies with two node_modules directories for the project. You can install different dependency versions in the web package and end up with different behavior in the web version and the command line version, negating a key advantage of making this change. The web version could not be separately published because it embeds the entire bids-validator source and you can't run one test suite at the top level to test the web package since it will load those modules with different dependencies (Next and React are missing when you try).

Can we reorganize this to use [workspaces](https://yarnpkg.com/lang/en/docs/workspaces/) instead? This does require moving all of the existing JavaScript into a subdirectory but workspaces solve these problems without necessarily requiring Lerna.

nellh:
We should also make sure to update the developer documentation since running tests will require you to use yarn instead of npm until npm@7 is released. End users can continue using npm without issue.

olgn:
@nellh that should be looking better - the codeclimate issues are the outstanding ones which now get triggered again (because we moved all the files)

Issue #703, opened 2019-02-07
Annexed files are counted twice in dataset summary


nellh:
I'm fairly certain the .git/annex copy used to be skipped but we're seeing file counts for datasets which include the git and git-annex metadata in the bids-validator summary for OpenNeuro datasets.

PR #714, opened 2019-02-15
Remove annexed file copies from summary

david-nishi:
#703


nellh:
This looks good to me but needs to be rebased.

Issue #715, opened 2019-02-16
Expose text color/formatting API for issues


nellh:
The validator has nice text formatting and color coding for summary and issues output but this is not currently exposed so that other JavaScript applications can reuse it. This would be really helpful for [openneuro-cli](https://www.npmjs.com/package/openneuro-cli) to provide comparably useful output when the validator is run as part of the upload process for users.

nellh:
This is done.

PR #716, opened 2019-02-16
expose colored/formatted console output in issues and summary

david-nishi:
#715


olgn:
works great - to clarify this change, the prettified console output is now stored as an additional field on the callback from `validate.BIDS`. `validate.BIDS` callsback with an `(issues, summary)` pair. 

* prettier issue formatting is stored in `issues.consoleOutput`
* prettier summary formatting is stored in `summary.consoleOutput`.

openneuro can now provide this pretty output by accessing these fields in the validation callback and feeding them to `console.log`  

@DaNish808 needs to be rebased but then it's ready to merge

nellh:
It would be more general to decouple the formatting from the validate function. This way you would have something like `validate.BIDS() => callback(issues, summary) => { formatIssues(issues), formatSummary(summary)}` which allows for those to be recomposed without changing validate.BIDS

olgn:
@nellh good point

nellh (approved):
Nice, thanks for making the decoupling changes, this should make it easier to use this in openneuro-cli.

PR #720, opened 2019-02-21
basic cbv support

TheChymera:
In reference to https://github.com/bids-standard/bids-specification/issues/149#issuecomment-465264324 and newly added CBV support in BIDS


chrisgorgo:
Thanks!

Issue #705, opened 2019-02-08
write a mock OpenNeuro (integration) test


ckrountree:

nellh:
This should replicate both entrypoints into the validator from OpenNeuro with a variety of test datasets, similar to the existing tests. The uploader embeds the validator and makes a call to BIDS.validate expecting a callback. Datalad-service runs the command line validator with the `--json` flag and should always get a stable valid JSON object (at least showing error 0 in case of non-zero exits).

PR #722, opened 2019-02-25
browser environment tests (jsdom) #705

david-nishi:
approaches #705 using separate jsdom environment
- runs bids.spec.js and utils/files.spec.js given browser's File and FileLists
- validate.BIDS entrypoint run with web-specific configs

### DIFF
--- a/bids-validator-web/tests/bids.spec.js
+++ b/bids-validator-web/tests/bids.spec.js
@@ -1,0 +1,218 @@
+/**
+ * eslint no-console: ["error", { allow: ["log"] }]
+ * @jest-environment jsdom
+ */
+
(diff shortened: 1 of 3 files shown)


nellh (changes requested):
Conceptually this looks great but we should find a way to avoid the duplication of the same tests, this will introduce maintenance effort whenever we make changes to those tests. This can be refactored to import the main test suite and use dependency injection or Jest's module mocking feature to run them with the jsdom specific functions.

nellh:
I think these tests might work better as part of the main module too, to avoid cross package imports.

nellh:
Replaced by #753

PR #735, opened 2019-03-01
[MRG] accept number instead of integer as type for sampling freq

sappelhoff:
fixes an error we encountered while going over the examples:

https://github.com/bids-standard/bids-examples/pull/144#issuecomment-468640345

This will allow float values in the "SamplingFrequency" field of the `ieeg.json` sidecar file.

cc @dorahermes


dorahermes (approved):
lgtm thanks!

PR #746, opened 2019-03-11
check for excessive commas in dataset_description.json Authors field

### LINKED ISSUE #629: Raise a warning if there is only one author and its name includes a comma
We want to catch situations like this
```
"Authors": [
        "Name1, O., Name2, J., Name3, D., & Name4, K.A."
    ]
```

olgn:
fixes #629.

* if any field of the Authors array in dataset_description.json contains more than one comma, throw a warning that gently suggests they make sure the field is properly formatted.

### DIFF
--- a/bids-validator/tests/utils/files.spec.js
+++ b/bids-validator/tests/utils/files.spec.js
@@ -48,9 +48,20 @@ describe('files utils in browsers', () => {
 
 describe('dataset_description.json', () => {
   it('throws warning if it does not exist in proper location', () => {
-    const fileList = {}
-    const issues = checkDatasetDescription(fileList)
+    const jsonFileContents = {}
+    const issues = checkDatasetDescription(jsonFileContents)
     assert(issues[0].key === 'DATASET_DESCRIPTION_JSON_MISSING')
+  })
+  it('throws a warning if the Authors field of the dataset description has too many commas', () => {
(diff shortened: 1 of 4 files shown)


nellh (changes requested):
This should be an error if there are multiple commas. One comma is a warning.

chrisgorgo:
> This should be an error if there are multiple commas. One comma is a warning.

To clarify, this is ok and should not raise a warning/error:

```
"Authors": [
        "Name1, O.",
        "Name2, J.",
        "Name3, D.", 
        "Name4, K.A."
    ]
```

This is also ok (although unusual for a dataset to have only one author - most likely the curator did not want to put the effort of listing everyone):

```
"Authors": [
        "Name1, O."
    ]
```

olgn:
@chrisfilo so there should not be a warning when there is a single author with a single comma? and just an error on multiples?

chrisgorgo:
I thought about this a bit more. Let's do:
- warning if there is a single author with one or less comma (to encourage curators listing everyone involved in the project)
- error if there is a single author with more than one comma (to enforce the right formatting on OpenNeuro)

olgn:
@nellh resolved based on feedback from @chrisfilo and yourself

nellh (approved):

PR #765, opened 2019-04-11
fix typo

DimitriPapadopoulos:
I hope there are no dependencies that prevent from changing these messages.


sappelhoff (approved):
thanks for catching the typo. I think the fix should be fine, and the CIs pass, so all good :+1:

PR #768, opened 2019-04-16
FIX: Re-include versioneer.py in sdist package

### LINKED ISSUE #757: from source installation of bids-validator fails due to missing versioneer.py
It seems like the problem reported in #707 and (supposedly) fixed in #708 has resurfaced.

Trying to install `bids-validator` 1.2.2 using the source tarballs from PyPI (https://pypi.org/project/bids-validator/1.2.2/#files) fails with:

```
$ tar xfvz bids-validator-1.2.2.tar.gz
$ cd bids-validator-1.2.2
$ pip install --prefix /tmp/example .
Processing /tmp/bids-validator-1.2.2
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/eb-0cjpGh/pip-req-build-aja1ythx/setup.py", line 3, in <module>
        import versioneer
    ModuleNotFoundError: No module named 'versioneer'
```

Indeed the `versioneer.py` file is not there in the `bids-validator-1.2.2.tar.gz` source tarball available from PyPI...

effigies:
Closes #757.


effigies:
#757 reproduced in test. Patching now.

effigies:
@rwblair This is good to go.

PR #774, opened 2019-04-25
bring back correct exit code on validation w/ errors

david-nishi:


rwblair:
Any chance we could add a test to check that the correct error code is thrown if errors are found?

rwblair:
Very strange when I run 'yarn coverage' while ssh'd into circle container all the tests pass.

david-nishi:
@rwblair looks like i was using a dataset in the git-ignored directory tests/data/bids-examples-1.1.1u1. It's pretty small (900 Kb) so I just added it to tests/data.

It worked in the container because bids-examples-1.1.1u1 is unzipped from data/examples.zip later in the tests

PR #779, opened 2019-05-10
HED validator integration

### LINKED ISSUE #537: [feature] HED tag validation
[HED](http://www.hedtags.org/) tags are an optional part in BIDS data  `.tsv` files. 

It would be nice to validate a potential HED column in a BIDS `.tsv` file while running the bids-validator. Currently, there need to be two steps:

1. Using the BIDS-validator
2. Separately using an HED validation tool, such as provided by the authors of HED: https://github.com/VisLab/HEDTools

Unfortunately, there is no JavaScript HED validation tool at the moment - I am filing this issue regardless to be aware of this opportunity in the future.

### LINKED ISSUE #764: Need error codes for HED validation errors
We are essentially ready to incorporate a first version of syntactic HED validation into the bids-validator.  However, we need to reserve error codes for the following syntax errors and warnings:

**HED syntax errors:**  
* Unsupported character (illegal characters in the HED string, e.g. square and curly brackets)
* Parentheses mismatch
* Missing comma after tag
* First word of tag not capitalized or CamelCase (warning)
* Duplicate tag at top level or in single group
* Too many (more than 2) tildes in a group  

The current version of HED validation does not address validation against a particular HED schema, which is the next step.  We anticipate the following errors for semantic validation, and perhaps this would be a good time to reserve codes for these as well:

**HED semantic errors:**
* commaValid: Previous tag contains a comma when it shouldn’t or tag isn’t valid.
* isNumeric: Invalid numeric tag.
* leafExtension: Extensions are only allowed for leaf tags.
* requireChild: Descendant tag required.
* unique: Multiple unique tags with prefix.
* unitClass: Invalid unit; valid units are…
* invalidTag: Invalid tag.
* required: (WARNING) Tag with prefix <> is required.
* unitClass: (WARNING) No unit specified. Using <> as the default.  

Please let us know how to proceed.

### LINKED ISSUE #773: Reorganize event validation
As part of the process for adding HED validation (#537), I've noticed that the HED code will take up the majority of the `validators/events.js` module (it's already almost half, and I haven't even added the issue handling yet). I propose that the event validation code be reorganized along the lines of `bvec` and `bval`. The HED validation could then be in a separate module within that folder.

happy5214:
closes #537 
closes #764 
closes #773 

This pull request integrates the external `hed-validator` package to validate HED strings in `_events.tsv` files. At this time, it includes only syntax validation, but we expect to add semantic validation in the future. This code includes new issue codes for each current validation issue returned by `hed-validator`, plus generic codes for currently unknown issues that may be added. We have added test cases for each new issue code, except for the generic ones. To improve code clarity, we split the previous `validators/events.js` file into a folder similar to other validators, with the HED code in its own module.

### DIFF
--- a/bids-validator/package.json
+++ b/bids-validator/package.json
@@ -26,6 +26,7 @@
     "colors": "^1.3.1",
     "cross-fetch": "^2.2.2",
     "date-fns": "^2.0.0-alpha.22",
+    "hed-validator": "^0.1.0",
     "ignore": "^4.0.2",
     "jshint": "^2.9.6",
     "minimatch": "3.0.4",
--- a/bids-validator/package.json
+++ b/bids-validator/package.json
@@ -26,7 +26,7 @@
     "colors": "^1.3.1",
     "cross-fetch": "^2.2.2",
     "date-fns": "^2.0.0-alpha.22",
-    "hed-validator": "^0.1.0",
+    "hed-validator": "^0.2.0",
     "ignore": "^4.0.2",
     "jshint": "^2.9.6",
     "minimatch": "3.0.4",
--- a/bids-validator/package.json
+++ b/bids-validator/package.json
@@ -26,7 +26,7 @@
     "colors": "^1.3.1",
     "cross-fetch": "^2.2.2",
     "date-fns": "^2.0.0-alpha.22",
-    "hed-validator": "^0.2.0",
+    "hed-validator": "^0.3.0",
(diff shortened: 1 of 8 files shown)


rwblair (approved):
@happy5214 Amazing work. Everything looks good, and thank you for adding tests. I'm waiting on another PR before releasing. Will also probably go test subset of openneuro datasets to see how many of them our out standard compliance with respect to HED.

PR #785, opened 2019-06-04
FIX - replaced old repository name (INCF) with new one

robertoostenveld:
The relevant change is on line 66, where the URL now points to https://github.com/bids-standard/bids-validator/tree/gh-pages instead of the old INCF location.

I did *not* change the quotation marks from ' into " on purpose (idem for the ; at the end of the line). My edit was in `vi`, so my only  guess is that the additional update of the quotation marks was done by one of the git hooks. That is something I cannot prevent.

### DIFF
--- a/bids-validator-web/components/Issues.jsx
+++ b/bids-validator-web/components/Issues.jsx
@@ -1,91 +1,92 @@
 // dependencies -------------------------------------------------------
 
-import React from 'react'
-import PropTypes from 'prop-types'
-import pluralize from 'pluralize'
-import Results from './results/Results'
-import Spinner from './Spinner.jsx'
-import ErrorLink from './ErrorLink.jsx'
-import Summary from '../components/Summary'
+import React from "react";
+import PropTypes from "prop-types";
+import pluralize from "pluralize";
+import Results from "./results/Results";
+import Spinner from "./Spinner.jsx";
+import ErrorLink from "./ErrorLink.jsx";
+import Summary from "../components/Summary";
 
 class Issues extends React.Component {
   constructor(props) {
-    super(props)
-    this._reset = this.props.reset.bind(this)
+    super(props);
+    this._reset = this.props.reset.bind(this);
   }
   // life cycle events --------------------------------------------------
 
   render() {
     // short references
-    let errors = this.props.errors
-    let warnings = this.props.warnings
-    let dirName = this.props.dirName
+    let errors = this.props.errors;
+    let warnings = this.props.warnings;
+    let dirName = this.props.dirName;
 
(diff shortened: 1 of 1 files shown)


bids-validator-web/components/Issues.jsx
```diff
@@ -1,91 +1,92 @@
...
-    )
+    );
     let sourcecode = (
       <h5>
-        The source code for the validator can be found{' '}
+        The source code for the validator can be found{" "}
         <a
-          href="https://github.com/INCF/bids-validator/tree/gh-pages"
-          target="_blank">
+          href="https://github.com/bids-standard/bids-validator/tree/gh-pages"
```
sappelhoff: looks good to me ... HOWEVER, why is this pointing to the `gh-pages` branch and not `master`?

robertoostenveld: I also wondered about that, but did not dare to touch it. For me it makes more sense to point to the top level of the <https://github.com/bids-standard/bids-validator> repository.

robertoostenveld: Now that you are at it: I also propose to change the link at https://github.com/bids-standard/bids-validator/blob/d93f054ff8567243982b438991c83911c72a54c3/bids-validator-web/components/Issues.jsx#L38 to https://bids-specification.readthedocs.io where the current spec is.

robertoostenveld: That link is under "Click to view details on BIDS specification" here.

Screen Shot 2019-06-04 at 20 19 46

sappelhoff: yes, sounds reasonable!

```suggestion
          href="https://github.com/bids-standard/bids-validator"
```

bids-validator-web/components/Issues.jsx
```diff
@@ -1,91 +1,92 @@
...
+      totalWarnings = warnings.length;
+      warningCount = totalWarnings + " " + pluralize("Warning", totalWarnings);
+      errorCount = totalErrors + " " + pluralize("Error", totalErrors);
     }
     // messages
     let specLink = (
       <h5>
-        Click to view details on{' '}
+        Click to view details on{" "}
         <a href="http://bids.neuroimaging.io" target="_blank">
```
sappelhoff:
```suggestion
        <a href="https://bids-specification.readthedocs.io/en/stable/" target="_blank">
```

sappelhoff (commented):
good to go once the two links are changed as you proposed  @robertoostenveld

sappelhoff:
thanks @robertoostenveld

Issue #593, opened 2018-10-02
select options in browser?


sappelhoff:
Is there a way to enable the selection of validator option the browser version (i.e., https://bids-standard.github.io/bids-validator/)?

E.g., by adding some radio buttons for `--ignoreNiftiHeaders` or `--bep006`

PR #793, opened 2019-07-06
add options

david-nishi:
#593 Add radio buttons to set options ignoreWarnings and ignoreNiftiHeaders.

Screen Shot 2019-07-05 at 18 48 49

### DIFF
--- a/bids-validator-web/components/Options.jsx
+++ b/bids-validator-web/components/Options.jsx
@@ -4,9 +4,9 @@ const Options = ({ setOption, options }) => (
   <>
     <form className="options" onClick={setOption}>
       <label><strong>Options:</strong></label>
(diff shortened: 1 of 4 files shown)


bids-validator-web/components/Options.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+
+const Options = ({ setOption, options }) => (
+  <>
+    <form className="options" onClick={setOption}>
+      <label><strong>Options:</strong></label>
+      <input name="ignoreWarnings" type="radio" checked={options.ignoreWarnings} readOnly />
```
nellh: I'd make these checkboxes, it was initially confusing to see radio selects even though they function as checkboxes :smile:

nellh (changes requested):

PR #795, opened 2019-07-09
no authors warning

david-nishi:
OpenNeuroOrg/openneuro#1155
throw warning when datasetDescription Authors field is empty.
```
113: {
    key: 'NO_AUTHORS',
    severity: 'warning',
    reason:
      'The Authors field of dataset_description.json should contain an array of fields - with one author per field. This was triggered because there are no authors, which could lead to DOI minting failures and will prevent snapshot creation on OpenNeuro.',
  }
```

### DIFF
--- a/bids-validator/utils/issues/list.js
+++ b/bids-validator/utils/issues/list.js
@@ -630,6 +630,6 @@ module.exports = {
     key: 'NO_AUTHORS',
     severity: 'warning',
     reason:
-      'The Authors field of dataset_description.json should contain an array of fields - with one author per field. This was triggered because there are no authors, which could lead to DOI minting failures and will prevent snapshot creation on OpenNeuro.',
+      'The Authors field of dataset_description.json should contain an array of fields - with one author per field. This was triggered because there are no authors, which will make DOI registration from dataset metadata impossible.',
   },
 }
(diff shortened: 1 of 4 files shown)


nellh (changes requested):
This looks good except it does break some tests which relied on empty authors fields. That needs to be fixed for merge.

bids-validator/utils/issues/list.js
```diff
@@ -626,4 +626,10 @@ module.exports = {
     reason:
       'The json sidecar does not contain this column value as a possible key to a HED string.',
   },
+  113: {
+    key: 'NO_AUTHORS',
+    severity: 'warning',
+    reason:
+      'The Authors field of dataset_description.json should contain an array of fields - with one author per field. This was triggered because there are no authors, which could lead to DOI minting failures and will prevent snapshot creation on OpenNeuro.',
```
effigies:
```suggestion
      'The Authors field of dataset_description.json should contain an array of fields - with one author per field. This was triggered because there are no authors, which will make DOI registration from dataset metadata impossible.',
```

effigies (commented):
We should probably not put OpenNeuro-specific text in this package, which is meant to be generic to anybody working with BIDS, regardless of whether they intend to upload to OpenNeuro. I suggested some alternative text that may help make the issue clear, though it will not specifically warn about snapshot failure.

PR #794, opened 2019-07-08
exit process with code 3 on unhandledRejection

david-nishi:
#658

PR #796, opened 2019-07-09
Print version number in standard out.

### LINKED ISSUE #770: Print version number in standard out.
Would be useful in situation where bids-validator is being embedded in reports:
https://github.com/poldracklab/fmriprep/pull/1583

david-nishi:
closes #770 
logs `bids-validator@x.x.x` at start

### DIFF
--- a/bids-validator/bin/bids-validator
+++ b/bids-validator/bin/bids-validator
@@ -1,30 +1,35 @@
 #!/usr/bin/env node
 
-process.title = 'bids-validator';
+process.title = 'bids-validator'
 
 var argv = require('yargs')
-    .usage('Usage: $0 <dataset_directory> [options]')
-    .help('help')
-    .alias('help', 'h')
-    .version(require('../package.json').version)
-    .alias('version', 'v')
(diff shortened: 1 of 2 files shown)


nellh:
This will break any JSON readers (OpenNeuro). It should only be printed when the --json flag is not present.

david-nishi:
Thanks for the catch!

sappelhoff:
@DaNish808 if this PR fixes #770, ...

then perhaps you could edit your original post to include the words `closes` or `fixes` right in front of the `#770` --> `closes #770` ... then [GitHub will automatically close that issue upon a merge](https://help.github.com/en/articles/closing-issues-using-keywords).

I have noticed that this could also have been done in https://github.com/bids-standard/bids-validator/pull/793 and https://github.com/bids-standard/bids-validator/pull/794

But then again: Perhaps you are already aware of it, and it's intentional that you don't close those issues - in that case, sorry for lecturing :-)

david-nishi:
@sappelhoff useful trick! Thanks for sharing :D

PR #801, opened 2019-07-18
upgrade packages github was complaining about

rwblair:


effigies (approved):

nellh:
The packages which are pulling these in should be updated or only the lockfile, but they are not direct dependencies of the validator, so including them in package.json adds dependencies which may not be needed in the future.

Issue #822, opened 2019-08-19
Unhandled rejection (reason: {}).


munoztd0:
Hello, I always get the same error 

Unhandled rejection (reason: {}).

(Running Ubunutu 18.04)

sappelhoff:
Hi @munoztd0 

- which version of the bids-validator are you running?
- how did you install bids-validator?
- is this a dataset specific error, or does it occur for all datasets you are trying to test?
- what is your nodejs version? We require 10.11, see the [README](https://github.com/bids-standard/bids-validator)

dmoracze:
Hello, I am also receiving this error with version 1.3.0, I did not receive it with 1.2.5.

I don't know about the OP, but I am running the validator on an HPC using a singularity build from docker. Containers were built using:
`singularity build bids-validator-1.3.0.simg docker://bids/validator:1.3.0`
`singularity build bids-validator-1.2.5.simg docker://bids/validator:1.2.5` 

The validator is run using:
`singularity run -B /path/to/data:/data bids-validator-1.3.0.simg /data`

This error occurs on every dataset using 1.3.0 but it works fine with 1.2.5.

munoztd0:
So, 
Yes, it was an issue with node.js for me.
The weird thing is that the sudo apt-get install nodejs installs an older version.

So I fixed it by uninstalling and installing the new version via the source file from the website.

Thanks.

nellh:
> Hello, I am also receiving this error with version 1.3.0, I did not receive it with 1.2.5.
> 
> I don't know about the OP, but I am running the validator on an HPC using a singularity build from docker. Containers were built using:
> `singularity build bids-validator-1.3.0.simg docker://bids/validator:1.3.0`
> `singularity build bids-validator-1.2.5.simg docker://bids/validator:1.2.5`
> 
> The validator is run using:
> `singularity run -B /path/to/data:/data bids-validator-1.3.0.simg /data`
> 
> This error occurs on every dataset using 1.3.0 but it works fine with 1.2.5.

@dmoracze Thanks for the bug report, I've submitted #823 to fix this issue for Docker and Singularity builds.

If you're running into this error with a standalone Node.js environment, you need to upgrade to a newer version of Node.js. The [README](https://github.com/bids-standard/bids-validator#quickstart) shows the current minimum supported version.

PR #824, opened 2019-08-20
CI: Build and test the Docker image

effigies:
Should prevent oversights like #822.


effigies:
Yeah, okay, I don't get why this is failing. The exact commands work for me locally.

effigies:
@rwblair Fixed.

PR #828, opened 2019-08-27
Bep018

rwblair:
~~- [ ] Add valid genetic_info.json to an example dataset.~~
Added valid genetic_info.json else where in tests.


CPernet (approved):

sappelhoff (commented):
Thanks @rwblair - I have given this a coarse look (not comparing directly with the specification draft yet).

@CPernet could you also have a look at the code and check if some keywords are missing or optional/mandatory keywords are treated appropriately? No specific JS skills needed here, it's very readable.

Another point: Have you tested this on an example dataset @rwblair? If yes, which one? Can we edit a dataset on https://github.com/bids-standard/bids-examples to include genetics information?

CPernet:
I'm finishing a UK biobank example, will push it soon as a PR = we can test the validator

rwblair:
New specification PR  bids-standard/bids-specification/pull/395 for discussion.

rwblair:
@sappelhoff  @CPernet has made an example dataset that can be found here:
https://github.com/CPernet/bids-examples/

I have a PR from my branch of his examples back to his master that passes using this validator:
https://github.com/rwblair/bids-examples/tree/cpernet

I figure we merge validator first and then examples? The validator uses the examples in its tests and the examples use the validators in its tests. One of them has to be merged first, the examples will never pass against a specified version of the validator without the necessary changes to the validator being merged.

effigies:
@rwblair Let's merge the validator first.

PR #835, opened 2019-09-17
Add functional phase data patterns to file_level_rules

### LINKED ISSUE #834: Support phase functional MRI data
Per https://github.com/bids-standard/bids-specification/pull/128 and https://github.com/bids-standard/bids-specification/pull/312, phase fMRI data are supported as a suffix (`_phase`) in the `func` datatype. It doesn't look like that change has been pushed to the BIDS validator (at least as of the current latest release on DockerHub last month).

tsalo:
Closes #834. Adds support for functional phase data (added to the BIDS specification in bids-standard/bids-specification#128 and bids-standard/bids-specification#312).


effigies:
I would clump the suffixes together. `bold`, `sbref`, `cbv` and `phase` are all clustered on the big table, so let's put them together here.

Beyond this, @rwblair @nellh are there further considerations for adding a suffix?

tsalo:
@effigies Done!

In reordering the tokens I noticed that `_defacemask` is included as a `func` token even though it's only included under `anat` in the [Entity table](https://github.com/bids-standard/bids-specification/blob/master/src/99-appendices/04-entity-table.md). I can open a related issue either here (if it should be under `anat`) or in the specification repository (if the Entity table is wrong), but I wasn't sure which way it should go.

rwblair:
I can't seem to find a reference to defacemasks for functional images in the standard even outside the entity table. If its removed then we can simplify the whole section by adding a '_func_ext_' like other sections have for matching .nii nii.gz and .json, defacemask is the only one missing .json extension.

effigies:
Perhaps check the `git blame` to see if there was a use case driving its inclusion?

rwblair:
Looks like its origins are in #233. Implies that at some point in the spec that func defacemasks were acceptable.

rwblair:
ds000113 has funcitonal defacemasks for every run.
ds001107 is using them as well

So I guess well keep them in the validator for now. I'll keep looking to see if there is any further discussion around this.

tsalo:
Okay, that sounds good.

Regarding supporting functional phase data, is there anything else that needs to be changed? It seems to be working locally on my own complex (phase+mag), multi-echo data.

rwblair:
@tsalo one last thing, could you add an example of a valid filename to [bids-validator/tests/type.spec.js](https://github.com/tsalo/bids-validator/blob/23d39dfda9d8a844951957e6e6ad47e68d06a280/bids-validator/tests/type.spec.js#L41) around line 41 to have it covered by some sort of test?

PR #838, opened 2019-09-19
818 remove file.size when moved to stats

david-nishi:


nellh (changes requested):
The basic fix here is good but to work with the existing schema we also need to exclude the 'id' and 'key' fields from the final object, as these are only present on the gitTreeMode version.

nellh:
> The basic fix here is good but to work with the existing schema we also need to exclude the 'id' and 'key' fields from the final object, as these are only present on the gitTreeMode version.

It is also strange to put this side effect in collectDirectorySize but I'm not sure what a good alternative is.

nellh (approved):
Matches the non-gitTreeMode output exactly now :+1:

PR #843, opened 2019-09-24
change subjectMetadata to array

david-nishi:
change summary.subjectMetadata to array and update tests


nellh (approved):
Ran through a set of test datasets and this works well.

PR #875, opened 2019-11-22
[WIP] Fix circle build for annotated tags.

rwblair:
Issue seems to be with checkout command:
https://discuss.circleci.com/t/checkout-fails-on-workflow-triggered-by-annotated-git-tag/27272


effigies:
This is weird, because every other project I know of that builds on CircleCI handles annotated tags fine.

Issue #873, opened 2019-11-22
Validation fails for subjects with labels with initial uppercase


effigies:
From [Neurostars #5472](https://neurostars.org/t/bids-validator-error-for-upper-case-subject-labels/5472):

> The bids specification allows for upper case labels for subjects:
> 
> BIDS uses custom user-defined labels in several situations (naming of participants, sessions, acquisition schemes, etc.) Labels are strings and MUST only consist of letters (lower or upper case) and/or numbers.
> 
> However when i use upper case e.g sub-X01, the bids validator (http://bids-standard.github.io/bids-validator/) says error 1 “NOT_INCLUDED” and if i use sub-x01 then it passes. I think in this regard, the bids validator is therefore inconsistent with bids specification…
> 
> Best wishes,
> 
> Ollie Hulme

effigies:
Another user encountering this in [Neurostars #5496](https://neurostars.org/t/dont-understand-why-validation-fails-when-naming-scheme-matches/5496).

Anybody know how to debug javascript?

sappelhoff:
I tried to replicate this on linux Ubuntu 18.04 with `master` validator and the `eeg_matchingpennies` example dataset.

1. Validate using `bids-validator eeg_matchingpennies/` --> all fine except for empty files
1. Rename `sub-05` in the dataset to `sub-X01`
1. Validate --> same as step 1, ... no additional errors

furthermore given the status in 

- https://neurostars.org/t/bids-validator-error-for-upper-case-subject-labels/5472 and
- https://neurostars.org/t/dont-understand-why-validation-fails-when-naming-scheme-matches/5496

I assume that this problem has been solved.

PR #880, opened 2019-12-02
FIX: Extra close-paren in regex

effigies:
Not sure if this will address #873. Looking for weird regexes with `sub-` involved.


sappelhoff (approved):
yes, seems right to me to remove that `)` from the regexp.

But I am not sure what that fixes, and why it was not caught by our tests.

sappelhoff:
merging, because it's a fix. but it's probably fixing something that's broken on a larger scale and also currently unused --> #961

Issue #882, opened 2019-12-03
Remove lerna?


rwblair:
Currently only bids-validator is having a version set by lerna. @nell what would the consequences of removing lerna and having npm pushes done in circle deployment workflow be? Is lerna responsible for anything besides setting version number/tag and pushing to npm?

Right now we get tags from lerna and then again from github publish, which had caused some problems with versioneer, and each tagging gets CI run against it. The other two solutions I can see would be to use circle filters to only run tests against one format of tag, and the other would be to have the github tag match the lerna style of tag, with the version being prepended with "bids-validator@".

nellh:
The other useful thing about Lerna was kind of supplanted by Yarn/NPM workspaces and NPM will support these soon, so it could be removed at that point if we don't want to use it to manage interdependencies between the web validator and core validator.

rwblair:
With #883 fixed this isn't an issue at the moment.

PR #883, opened 2019-12-03
Lerna fix: Use shared versioning for packages

nellh:
I saw #882 and I believe Lerna is misconfigured for the project. This switches to managing one version across multiple packages which was the original benefit of using Lerna in the first place. With this change, `lerna publish` will push one GitHub tag and manage the package version for both packages.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -3,5 +3,5 @@
     "bids-validator",
     "bids-validator-web"
   ],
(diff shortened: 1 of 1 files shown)


rwblair:
With this when we run lerna publish it will push "1.3.12" instead of "bids-validator@1.3.12" as the tag?

Issue #777, opened 2019-05-06
Validate units used in TSV files.


sappelhoff:
Several `.tsv` files can have a column called `units`.

We have a list of permissible units in our [appendix V](https://bids-specification.readthedocs.io/en/stable/99-appendices/05-units.html).

# The validator should ...

- check each tsv file for a "units" column
- check each unit in all units columns
- throw an error if a unit is not part of our list in appendix II

# Basic example of what this feature would fix

The `channels.tsv` file in https://github.com/bids-standard/bids-examples/pull/155#discussion_r281095667 contained a unit `uV` to indicate micro Volts. The validator accepted this, although according to our Appendix II, we would only accept `µV` or `microV` (...or `µvolt` or `microvolt`)

The validator could catch this case and throw a message "please fix this unit in that file, that column, that row" ... and link to our appendix.

cc @ftadel

sappelhoff:
What is the status on this issue @DaNish808 @ckrountree @rwblair ?

ckrountree:
@sappelhoff It was put back into the backlog for the moment.  We had some ON issues arise that needed more immediate attention.  That being said, we recognize that this is a higher priority BIDS issue!  We will address it asap.

ckrountree:
Released in v3.9

PR #885, opened 2019-12-06
validate units for tsv files

david-nishi:
Checks unit prefixes and suffixes against [bids-spec approved values](https://bids-specification.readthedocs.io/en/stable/99-appendices/05-units.html).

Creates errors specifying the line and unit prefix/root.


nellh (approved):
Looks good, I was going to complain about not allowing 'u' to sub 'µ' but it looks like this PR follows the BIDS spec more accurately than me, so :+1:

effigies:
Linking #777.

PR #891, opened 2019-12-18
Add default path for validator configuration

### LINKED ISSUE #741: Pickup config from a dataset at some standard location
ATM config should be specified in the command line. Since config quite often dataset specific, such configuration would be stored within the dataset.  To not breed various user-specific paths, would be nice to standardize location of such a config (e.g. `.bids-validator.cfg`) and pick it up by default.  It would then require an option to avoid loading it though (e.g. an empty config path in `-c`?)

Might interact with bids-standard/bids-validator#294

nellh:
This adds a default configuration path which is searched from the dataset root. You can add this configuration file to your dataset `dataset root/.bids-validator-config.json` and share any dataset specific config needed.

The web validator is updated to use this file and override it with the page controls. Other consumers of the bids-validator will pick up this behavior if they do not override the configuration already, but those that do override it will need to decide if they want to ignore this file or merge it with any other overrides (e.g. OpenNeuro will not automatically start parsing this file without changes but other usages may if they do not configure the validator already).

There is one breaking change, the config option path is now relative from the dataset root unless it is an absolute path. Previously it was relative to the working directory where the validator was executed.

Fixes #741

### DIFF
--- a/bids-validator-web/components/App.jsx
+++ b/bids-validator-web/components/App.jsx
@@ -23,7 +23,7 @@ const initState = () => ({
   options: {
     ignoreWarnings: false,
     ignoreNiftiHeaders: false,
-  }
+  },
 })
 
 export default class App extends React.Component {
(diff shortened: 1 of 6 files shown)


david-nishi (approved):
Very cool

Issue #820, opened 2019-08-14
Doesn't catch participants.tsv having multiple rows per participant


yarikoptic:
original investigation: https://github.com/bids-standard/pybids/issues/474#issuecomment-521240290

example dataset/file: https://github.com/OpenNeuroDatasets/ds001890/blob/master/participants.tsv

<details>
<summary>this participants.tsv is not BIDS compliant -- it has multiple entries for the participants with multiple sessions -- one row per session:</summary> 

```shell
$> head participants.tsv 
participant_id  session sex     genotype        Weight  SpO2    HR      Temperature     DOB     Experiment_Date Age
c1NT    1       M       3xTG    32.3    98      272     35.8    2016-11-22      2017-03-23      3
c1NT    2       M       3xTG    36.2    94      311     35.8    2016-11-22      2017-05-31      6
c2NT    1       M       3xTG    24.8    98      257     36.9    2016-11-29      2017-03-23      3
c2NT    2       M       3xTG    30.6    99      450     35.8    2016-11-29      2017-05-31      6
c2R1    1       M       3xTG    25      97      360     36.1    2016-11-29      2017-03-23      3
c4NT    1       M       C57BL/6 31.5    99      250     36      2016-12-10      2017-03-29      3
c4NT    2       M       C57BL/6 35      96      208     34.8    2016-12-10      2017-06-19      6
c5NT    1       M       C57BL/6 30.1    99      427     35.2    2016-12-10      2017-03-29      3
c5NT    2       M       C57BL/6 31.9    93      223     34.6    2016-12-03      2017-06-05      6
...

$> awk '{print $1;}' participants.tsv | sort | uniq -c | nl
     1        2 c11L
     2        2 c1NT
     3        1 c21L
     4        2 c2NT
     5        1 c2R1
     6        1 c31L
     7        2 c31R
     8        2 c32L
     9        2 c32R
    10        1 c3NT2
    11        2 c41L
    12        2 c41R
    13        2 c42L
    14        2 c42R
    15        2 c4NT
    16        2 c51L
    17        2 c5NT
    18        2 c61L
    19        2 c61R
    20        1 c62L
    21        2 c6NT
    22        2 c71L
    23        2 c71R
    24        2 c7NT
    25        2 c81L
    26        2 c81R
    27        2 c8NT
    28        2 c91L
    29        1 c91R
    30        1 c9NT
    31        1 participant_id

$> ls -ld sub-jgrADc* | nl | tail
[...]

PR #899, opened 2020-01-14
787 channels tsv status

exbotanical:


rwblair:
@MatthewZito I added a test for this PR, feel free to revert if we should be testing it in a different way.

exbotanical:
> @MatthewZito I added a test for this PR, feel free to revert if we should be testing it in a different way.

Thanks, Ross! Also, I'll be sure to resolve the issue number conflict before merging #820.

david-nishi (approved):

PR #903, opened 2020-01-15
820 participant rows bugfix

exbotanical:
1. Disallows duplicate `participant_id`s in `participants.tsv`.
2. Requires `sub-` prefix on all `participant_id`s in `participants.tsv`.

NOTE: Per Ross' comment, the most recent commit preemptively resolves the merge conflict with #787 by including issue id no.125 as a shared entry among both branches.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,10 +9,10 @@ jobs:
       - checkout
       - restore_cache:
           keys:
-            - npm-deps-{{ checksum "package.json" }}
+            - npm-deps-{{ checksum "package.json" }}-{{ checksum "yarn.lock" }}
       - run: yarn
       - save_cache:
-          key: npm-deps-{{ checksum "package.json" }}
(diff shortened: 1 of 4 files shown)


rwblair:
Issue code's might step on #787, its using 125 for the channel status column error.

david-nishi (dismissed):

david-nishi (changes requested):
Tests just need to be updated with new issue numbers. All else looks good!

exbotanical:
> Tests just need to be updated with new issue numbers. All else looks good!

Missed those. Thanks, just pushed up :-)

sappelhoff (changes requested):
meanwhile a part of this feature (`MISSING_SUB_PREFIX`) has been added in https://github.com/bids-standard/bids-validator/pull/1201

I think we don't have the `DUPLICATE_PARTICIPANT_ID` check from this PR in bids-validator yet, so that'd be a valuable addition.

However this PR needs to be rebased on `master`, and the current `MISSING_SUB_PREFIX` implementation needs to be removed.

PR #905, opened 2020-01-24
Small updates to bring in line with spec.

rwblair:
2 updates so far:
add age and sex to non_custom_column list for participants.tsv
extend top level func regex to include _ce and _dir options in file name

Last one addresses issue seen here https://neurostars.org/t/surprising-error-with-bids-validator/2347/4

I have not checked the other top level file regex's to see if there are any similar issues.


effigies:
Tests failing?

effigies:
I guess unrelated tests. This looks fine to me.

rwblair:
Yeah kind of frustrating, tests pass locally, and pass when I ssh into circle container, not sure what's up.

nellh:
The one mystery tests is failing because .bidsignore is not applied to the dataset tested but I was running into the same problem, it isn't reproducible except during an automated CircleCI run. The same containers run elsewhere work fine and ssh'ing into CircleCI seems to fix it as well.

PR #908, opened 2020-01-27
the tag prefix changed? 

rwblair:
New tag generation seems to use 'v' as a prefix instead of 'bids-validator@'


rwblair:
Versioneer from last tag push failed.

effigies:
You'll probably want to rerun versioneer install.

PR #920, opened 2020-02-27
DOC: Add Zenodo badge and initial .zenodo.json

### LINKED ISSUE #781: Zenodo entry for BIDS-Validator
Having a DOI for the validator would make referencing it a lot cleaner (and more amenable to scientific journals)

sappelhoff:
closes #781 

what's missing from `.zenodo.json` is the `creators` field to determine authors, their order, affiliations, and possibly ORCIDs.

We should probably use a script like this one in pybids: https://github.com/bids-standard/pybids/blob/master/tools/prep_zenodo.py

supplinyg a `.zenodo.json` without a `creators` field may result in either:
- zenodo's default creator parsing is preserved and displayed
- zenodo interprets this as "no creators"

I hope it's the first of these options, but I failed to find out for sure, because this is not very well documented anywhere.


effigies:
It's the former.

PR #923, opened 2020-02-28
prefix for micro can be either 'GREEK SMALL LETTER MU' or 'MICRO SIGN'

gpiantoni:
I'm getting an error from the validator regarding the SI units in a _channels.tsv file:
`Subunit (μV) of unit μV is invalid.`

The units in my channels.tsv file are in the from `μV` (`'GREEK SMALL LETTER MU'` in Unicode) while the BIDS validator only allows for `µV` (`'MICRO SIGN'`). However according to the [Unicode consortium](https://unicode.org/reports/tr25/) (and the [page linked](https://www.wikiwand.com/en/Micro-) on the bids-specification), the preferred sign is `μ` (`'GREEK SMALL LETTER MU'`) while `µ` (`'MICRO SIGN'`) should also be recognized.

This PR adds `μ` (`'GREEK SMALL LETTER MU'`) to the list of acceptable prefixes.

### DIFF
--- a/bids-validator/utils/unit.js
+++ b/bids-validator/utils/unit.js
@@ -89,6 +89,7 @@ const prefixes = [
   'milli',
   'm',
   'micro',
+  'μ',
   'µ',
   'nano',
   'n',


sappelhoff:
hi @gpiantoni, thanks a lot for the links. I wasn't aware that `GREEK SMALL LETTER MU` was officially preferred (recommended) over `MICRO SIGN`.

However, please see the current discussion on units here: https://github.com/bids-standard/bids-specification/pull/411

in the future, we will allow only ASCII representations of units (according to CMIXF-12) and three unicode characters for backward compatibility ("degree", "ohm", and "micro" signs, see the PR I link to in this comment)

sappelhoff:
@gpiantoni I am closing this issue now. But if this topic is interesting to you, please help us get everything right in https://github.com/bids-standard/bids-specification/pull/411

PR #938, opened 2020-04-02
allow 'n/a' in channels status column

rwblair:


bids-validator/validators/tsv/checkStatusCol.js
```diff
@@ -1,19 +1,24 @@
 const Issue = require('../../utils').issues.Issue
 
+/**
+ * Checks status column in a electroencephalography _channels.tsv file to
+ * ensure its values are only * 'good', 'bad', or 'n/a'
+ */
```
sappelhoff, line 12: cool to see a docstring! :-)

Is it also possible to document the inputs and the result (and/or output) of the function?

sappelhoff (approved):

PR #940, opened 2020-04-05
DOC: add TOC  + docker info

### LINKED ISSUE #931: improve the docs on how to use docker
I was playing around with docker today and wanted to use the bids-validator docker image.

The two steps described in the README are not sufficient to get started for newbies:

```
Docker
1. Install Docker
2. From a terminal run docker run -ti --rm -v /path/to/data:/data:ro bids/validator /data
```

questions / ideas about how the current state could/should be improved:

1. should link to how install docker / to docker website
1. what do the parameters `-ti` and  `--rm` mean
1. `-v` means mounting a volume I guess? but it's unclear what the `:` means and the `ro` at the end ... also it's unclear what part of that string should be replaced by a real path and what shouldn't
1. the `bids/validator` near the end is the image I guess? We should get some kind of badge from DockerHub or at least link to it
1. we should say how often the docker image is updated (if it's done by a CI, say so, and report status through badge etc.)
1.  what's the final `/data` in the command supposed to do?

sappelhoff:
closes #931 

This PR enhances the current documentation in several places.

- add a TOC to the README
- add documentation on Docker

preview of rendered README (see especially the TOC, and the new "Docker image" section):

https://github.com/bids-standard/bids-validator/blob/2152926af2326c6f69b810a23339f48ddfc2e3ce/bids-validator/README.md

### DIFF
--- a/bids-validator/README.md
+++ b/bids-validator/README.md
@@ -289,7 +289,7 @@ See here for a brief explanation of the commands:
 
 - `docker run` is the command to tell docker to run a certain docker image,
   usually taking the form `docker run <IMAGENAME> <COMMAND>`
-- the `--ti` flag means the inputs are accepted and outputs are printed to the
+- the `-ti` flag means the inputs are accepted and outputs are printed to the
   terminal
 - the `--rm` flag means that the state of the docker container is not saved
   after it has run
--- a/bids-validator/README.md
+++ b/bids-validator/README.md
@@ -17,6 +17,7 @@
       * [In the Browser](#in-the-browser)
       * [On the Server](#on-the-server)
       * [Through Command Line](#through-command-line)
+   * [Docker image](#docker-image)
    * [Python Library](#python-library)
       * [Example](#example)
    * [Development](#development)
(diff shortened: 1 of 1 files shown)


sappelhoff:
I did not really find where / how the docker image gets pushed to DockerHub ... is this done manually @rwblair ?

edit: Okay, got it -- apparently there is a webhook on Dockerhub and it builds whenever it recognizes a merged commit here

bids-validator/README.md
```diff
@@ -229,45 +250,77 @@ This style of use puts limits on what configuration you can require, so for
...
+  docker image
+- `docker run -ti --rm  bids/validator --help` to print the help
+- `docker run -ti --rm -v /path/to/data:/data:ro bids/validator /data`
+   to validate the dataset `/path/to/data` on your host machine
+
+See here for a brief explanation of the commands:
+
+- `docker run` is the command to tell docker to run a certain docker image,
+  usually taking the form `docker run <IMAGENAME> <COMMAND>`
+- the `--ti` flag means the inputs are accepted and outputs are printed to the
```
nellh: Should be `-ti` since it is two flags.

sappelhoff:
```suggestion
- the `-ti` flag means the inputs are accepted and outputs are printed to the
```

PR #950, opened 2020-04-29
Fix $CIRCLE_TAG not being properly read in gh-pages commit.

rwblair:


.circleci/config.yml
```diff
@@ -147,7 +147,7 @@ jobs:
       - run: git status
       - run:
           name: Commit updates
-          command: git add . && git commit --allow-empty --no-verify -m 'Circle CI - Generate validator demo with version $CIRCLE_TAG'
+          command: git add . && git commit --allow-empty --no-verify -m "Circle CI - Generate validator demo with version $(echo $CIRCLE_TAG)"
```
effigies: I think it's just the double vs single quotes. If you can `$(echo $x)` you can `$x`.

```suggestion
          command: git add . && git commit --allow-empty --no-verify -m "Circle CI - Generate validator demo with version $CIRCLE_TAG"
```

rwblair: You are correct.

rwblair: And for some reason I made a new commit instead of just accepting your suggestion.

effigies (approved):

PR #954, opened 2020-05-08
corrected ieeg filetype extension from .mef to .mefd

### LINKED ISSUE #953: Validator error with iEEG data in MEF3 format
I am preparing a BIDS-iEEG dataset to share on OpenNeuro. Some of my iEEG data are in MEF3 format, which should currently be allowed (but not supported) and maybe give a warning, but not an error. The validator does however not recognize the _ieeg.mefd file as BIDS-compatible and gives errors. 

@sappelhoff @maxvandenboom

dorahermes:
to resolve issue #953 

closes #953

### DIFF
--- a/bids-validator/bids_validator/rules/file_level_rules.json
+++ b/bids-validator/bids_validator/rules/file_level_rules.json
@@ -185,7 +185,7 @@
         "set",
         "fdt",
         "nwb",
-        "mef"
+        "mefd"
       ],
       "@@@_ieeg_ext_@@@": [
         "_events\\.json",


sappelhoff (commented):
From the [MEF 3 specification](https://msel.mayo.edu/files/codes/MEF%203%20Specification.pdf) I see that `.mefd` is the extension for a MEF3 "session directory" in which several kinds of data are saved (videos, channels). 

Is it fair to say that to use MEF3 in BIDS, a user would have to supply a `mefd` directory for each recording? I.e., use a `mefd` "file" similar to a NIfTI file? If yes, then this PR could be merged :+1: 

I think it'd also be good to clarify this in the [iEEG part of the BIDS specification](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/04-intracranial-electroencephalography.html#ieeg-recording-data), where we only mention `.mef`.

@dorahermes perhaps you want to take a stab at that, now that you have some experience with MEF3?

PS: I edited your PR post to include the phrase "closes #953" --> this phrase will ensure that the issue #953 is automatically closed once this PR is merged (just for PRs you may open in the future).

dorahermes:
Yes, it is correct that you indeed would supply a .mefd directory in case of mef3, this would be called `sub-<>_ses-<>_task-<>_run-<>_ieeg.mefd`  I can take a stab at the iEEG part of the BIDS-spec.

Would you also like me to add an example to the bids-examples?

sappelhoff:
> I can take a stab at the iEEG part of the BIDS-spec.

great!

> Would you also like me to add an example to the bids-examples?

that would be the full package! validator, spec, example :-)

Thanks a lot already for this PR.

PR #961, opened 2020-05-18
[PYTHON] -> get_path_values method and path.json rules BROKEN + UNUSED

sappelhoff:
The `get_path_values` method is not tested and not used anywhere throughout the code (or pybids):

https://github.com/bids-standard/bids-validator/blob/40ce6c45c2c9fd2e30284a0484769db915cd8382/bids-validator/bids_validator/bids_validator.py#L167-L178

Trying to test it locally also revealed that it's broken, because it expects `regexps` to be a dict, but it's actually a list, and all other validator python methods also expect it to be a list.

Digging further in, the `path.js` file that is used to store the regular expressions for `get_path_values` is also broken, in that the `get_regular_expressions` method cannot read it. This was never caught because it's not tested and not used anywhere (see also #880)

https://github.com/bids-standard/bids-validator/blob/40ce6c45c2c9fd2e30284a0484769db915cd8382/bids-validator/bids_validator/rules/path.json#L1-L6

Does somebody know where this came from? 
- Do we need to fix it?
- Can we remove it as garbage that we forgot to clean?


sappelhoff:
in fact, the method of the bids-validator using `path.json` is not used in pybids (or anywhere else I looked)

https://github.com/bids-standard/bids-validator/blob/1875a4399a0c71a1bcdfae66accb3d96881868db/bids-validator/bids_validator/bids_validator.py#L167-L178

sappelhoff:
@effigies @rwblair could you please double check this if PR you have 10 minutes?

ready to be merged from my side.

effigies (approved):
I think common.js itself is entirely unused...

No need to remove it here. Just was trying to see if the removed import might have a consequence.

Issue #959, opened 2020-05-17
Python implementation incompatible with windows paths


MaxvandenBoom:
Hi guys,

I really like the cross-platform compatibility of the bids-validator and the .js node implementation works perfect, but the python implementation breaks on windows filepaths (due to the backslashed instead of forward slashes).

It is possible to change the slashes in the filepaths in my own code (before feeding them to python's is_bids function), however, that feels more like a workaround rather than a fixing the issue and keeping the bids-validator cross-platform.

There are two relatively easy fixes. One quick and dirty, which is to add: 
`path = path .replace("\\", "/")`
inside `is_bids`. But this might have an influence if you have backslashes in the directoryname (which is not allowed in windows, but is allowed on mac).

Or, a more elegant solution, is to adjust the test regexes to allow both a forward or a backward slash as directory seperator. e.g. to change: 
`"^\\/(sub-[a-zA-Z0-9]+)\\/\\1(@@@_subject_level_ext_@@@)$"`
to:
`"^(\\\\|\\/)(sub-[a-zA-Z0-9]+)\\/\\2(@@@_subject_level_ext_@@@)$"`
I guess this is more invasive since the regexes are used in general. Then again, it might also be a good thing to have the (general) regexes to be able to handle windows paths for the future.

What do you think? I am willing help and provide a PR, but I would like to know what you think first.

Regards,
Max

MaxvandenBoom:
I am aware of issue https://github.com/bids-standard/bids-validator/issues/928, but I prefer not to settle for a workaround and am offering two solutions with a pull-request here

sappelhoff:
Hi @MaxvandenBoom thanks for suggesting some improvements!

> But this might have an influence if you have backslashes in the directoryname (which is not allowed in windows, but is allowed on mac)

as far as I know this is also not allowed in BIDS :thinking: but I may fail to think about some edge case here.

potential third option:

Would it be an option to parse a regexp and replace any kind of slashes in the regexps with `os.sep` in Python?

Provided that these slashes are not escaped or escape slashes themselves?

Finally, I think it'd be good to first replicate the issue in our CIs before we start making fixes. I got started with that here: https://github.com/bids-standard/bids-validator/pull/933 but it currently only runs in the CircleCI workflow, which does not cover Windows.

If we agree that this issue should be solved (and I do agree) then perhaps a good first PR would be to add Windows testing of the Python bids-validator. 

[...]

PR #966, opened 2020-05-23
[PYTHON] add cross platform testing and real test case

sappelhoff:
This PR makes use of GitHub Action to test the `bids_validator` package on all 3 major platforms.

I also add a "proper" test beyond a doctest, where I use datalad and git annex to get a test data skeleton and then validate the file names.

~Originally I expected this PR to fail on Windows (#959), so that we can start fixing the issue ... but it seems like it works fine?~

EDIT: the reason windows passed was that I let it validate an empty file list :man_facepalming:  ... fixed in https://github.com/bids-standard/bids-validator/pull/966/commits/f715b898fb2c6cc9d2f4248e1f2c6b95bcea97cd --> so now we do replicate the issue from #959 and once we merge this PR we can start fixing it :slightly_smiling_face: 

cc @MaxvandenBoom

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,13 +3,29 @@ version: 2
 jobs:
   test_python:
     docker:
-      - image: circleci/python:stretch
+      - image: continuumio/miniconda3
+    environment:
+      BASH_ENV: "/root/.bashrc"
     steps:
       - checkout
-      - run: pip install --user pytest flake8 pydocstyle
-      - run: pytest --doctest-modules bids-validator/bids_validator
-      - run: flake8 bids-validator/bids_validator
-      - run: pydocstyle bids-validator/bids_validator/bids_validator.py
+      - run:
+          name: setup conda
+          command: |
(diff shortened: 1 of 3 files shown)


MaxvandenBoom:
Nice!! Let me know if I can help

sappelhoff:
oookay, as discussed in the maintainers meeting some hours ago:

- circleci testing is back and now enhanced with a proper test (beyond doctest)
- it's good to have a python test there so that if it fails, the package will not be pushed to pypi
- we can test python in parallel on circleci (only linux) and github actions (win, mac, lin)

of course it'd be better to test windows and mac also on circleci ... but it's not so straight forward as it is with GH actions.

finally: moving the python publish job to GH actions is a potential idea, but it would tear apart the bids-validator release process for JS and Python, ... currently it's nicely aligned.

sappelhoff:
@MaxvandenBoom you can now help by submitting a PR to fix the windows CI for Python :slightly_smiling_face: 

as you suggested in #959

PR #973, opened 2020-06-02
Fix symlink test on Windows

### LINKED ISSUE #926: readDir does not behave as expected when following symbolic links on Windows
Test failure reported [here](https://gitlab.com/bids-standard/bids-validator/-/jobs/459698562). 

```
 FAIL bids-validator/utils/files/__tests__/readDir-examples.spec.js
  ● readDir.js - examples integration › readDir() › correctly follows symlinks for subjects with followSymbolicLink: true
    expect(received).toHaveLength(expected)
    Expected length: 12
    Received length: 6
    Received array:  ["0", "1", "2", "3", "4", "5"]
      20 |         ignoreSymlinks: false,
      21 |       }).then(files => {
    > 22 |         expect(Object.keys(files)).toHaveLength(12)
         |                                    ^
      23 |         const filenames = Object.values(files).map(f => f.name)
      24 |         filenames.sort()
      25 |         expect(filenames).toEqual([
      at toHaveLength (bids-validator/utils/files/__tests__/readDir-examples.spec.js:22:36)
```

nellh:
closes #926

@sappelhoff The reason this would fail hit me when I saw your comment in #926.

This fixes the test suite on Windows by working around GitLab CI's default of `core.symlinks=false` during checkout on Windows.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,5 +1,6 @@
 variables:
   GIT_STRATEGY: fetch
+  GIT_CHECKOUT: "false"
   GIT_SUBMODULE_STRATEGY: normal
   NODE_OPTIONS: --no-warnings
 stages:
@@ -14,5 +15,8 @@ tests:
     - choco install -y yarn
     - refreshenv
   script:
+    - git config core.symlinks true
+    - git checkout $CI_COMMIT_REF_NAME
+    - git submodule update --init --depth 1
     - yarn
     - yarn jest --testTimeout 30000


nellh:
GitLab delayed running GitHub forks for a release, so I ran this manually to [test it here](https://gitlab.com/bids-standard/bids-validator/-/jobs/578156469)

sappelhoff (commented):
oh, cool! Thanks a lot @nellh 

would a potential user have to configure anything on their windows machine? I am thinking about whether or not we should add a sentence to the install instructions for windows.

nellh:
core.symlinks=true is the default behavior for git, so most Windows users won't need to do anything. This also really only affects the test suite, since if you are running the validator on a system without symlink support, your datasets won't contain symlinks either.

sappelhoff (approved):
> if you are running the validator on a system without symlink support, your datasets won't contain symlinks either.

:smile: true.

Thanks, I think we can merge this!

Issue #969, opened 2020-05-26
overhaul units validation in bids-validator


sappelhoff:
we are currently in the final review round of a PR in bids-specification that will require CMIXF formatting of SI units: https://github.com/bids-standard/bids-specification/pull/411

Once that PR is merged, we should "switch off" units validation as it is currently and then start re-implementing / changing the way we validate to be in line with the specification again.

We also need to consider cases where SI units are impossible -> in these cases, users can specify their units in JSON sidecars. (This needs to be covered by the tests, i.e., IF invalid SI unit found, but unit specified in JSON, do not raise ... else send error "please format as CMIXF")

Lastly, the validator should alert users when the INVALID unicode  [`U+03BC`](https://codepoints.net/U+03BC) (μ) is used instead of the VALID  [`U+00B5`](https://codepoints.net/U+00B5) (µ). The warning should include something like "please take extra care ... these characters *look* the same, but are different"

sappelhoff:
@rwblair will your new package https://github.com/poldracklab/cmixf-js be easily integrated into bids-validator?

rwblair:
Should be. It's published it on npm. Don't have time at the moment but I'll make a PR for the validator to try and use it next week.

sappelhoff:
bids-standard/bids-specification#411 is now merged, we should "deactivate" current unit validation ASAP @rwblair

DimitriPapadopoulos:
Uncommenting the SI units validation code would fix this LGTM.com recommendation:
https://lgtm.com/projects/g/bids-standard/bids-validator/snapshot/18d01f82b84c4f67c48922e13928940176011ff4/files/bids-validator/validators/tsv/tsv.js?sort=name&dir=ASC&mode=heatmap#x247409e34f5d77de:1
https://github.com/bids-standard/bids-validator/blob/abb700c79b6d79683598a394aea5e0573b33412b/bids-validator/validators/tsv/tsv.js#L309-L336

DimitriPapadopoulos:
I have created a PR to uncomment the SI units validation code.

 It's been sitting unused for a long time. Does it need some overhauling or more testing?

sappelhoff:
yeah it needs both overhauling and testing.

We would want the unit checks to:

1. see if a unit anywhere in a BIDS metadata file is in CMIXF
2. if it is, all fine
3. else: **warn**, that CMIXF formatting is RECOMMENDED
[...]

PR #978, opened 2020-06-22
Disable old tsv unit checks.

rwblair:
comment out si unit checking call in tsv validation and in tsv tests until it is brought in line with specification. first part of bids-standard/bids-validator#71


sappelhoff (approved):
Can this be merged as the first step of #969 @rwblair?

or do you have further plans with it?

PR #1006, opened 2020-06-26
Fix error with repetitionTime being cast as a number. 

### LINKED ISSUE #674: Validator gets stuck when incorrectly passed a string when expecting a number.
Hello,

Found a small issue when a youngster in my lab tried to create our first BIDS dataset. 

When entering `RepetitionTime` into the .json files inside the `func` folder, the validator gets stuck on validating and becomes unresponsive when the number entered is mistakenly surrounded by quotation marks as a string, for example: `{"RepetitionTime: "2.0"}`. Took us a while to find this tiny mistake we made, and the validator worked fine after we removed the quotation marks.

Should the validator detect this incorrect syntax and spit out an error saying "RepetitionTime should not be a string"?

rwblair:
Cast to number in nifti header validation occurring after calling toFixed, which failed if repetitionTime was a string.

fixes #674

### DIFF
--- a/bids-validator/validators/nifti/nii.js
+++ b/bids-validator/validators/nifti/nii.js
@@ -265,8 +265,8 @@ export default function NIFTI(
             }),
           )
         } else {
-          const niftiTR = Number(repetitionTime.toFixed(3))
-          const jsonTR = Number(mergedDictionary.RepetitionTime.toFixed(3))
+          const niftiTR = Number(repetitionTime).toFixed(3)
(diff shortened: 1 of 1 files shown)


effigies:
Will this just silently permit the string? Or push that error to a different check?

rwblair:
The json schema check will still produce an error about incorrect type. Opened an issue clarify the specification: bids-standard/bids-specification#516

Issue #945, opened 2020-04-23
Improve validation of coordsystem.json files


sappelhoff:
issue identified in https://github.com/bids-standard/bids-specification/issues/420

We have `coordsystem.json` files for EEG, iEEG, and MEG.

These get validated against JSON schemas. See for example the [EEG schema](https://github.com/bids-standard/bids-validator/blob/master/bids-validator/validators/json/schemas/coordsystem_eeg.json).

In a given `coordsystem.json` file, we expect the `XXXCoordinateSystem` field (exchange `XXX` for EEG, iEEG, MEG).

This field MUST have a string from a list of *restricted keywords* as its value. See the [list of restricted keywords](https://bids-specification.readthedocs.io/en/stable/99-appendices/08-coordinate-systems.html).

Currently, the schemas accept any string

https://github.com/bids-standard/bids-validator/blob/cb0e79898020fc2910ef4904c29bd7ca9fd8b539/bids-validator/validators/json/schemas/coordsystem_eeg.json#L6

https://github.com/bids-standard/bids-validator/blob/cb0e79898020fc2910ef4904c29bd7ca9fd8b539/bids-validator/validators/json/schemas/coordsystem_ieeg.json#L4

https://github.com/bids-standard/bids-validator/blob/cb0e79898020fc2910ef4904c29bd7ca9fd8b539/bids-validator/validators/json/schemas/coordsystem_meg.json#L4

This should be changed. E.g., for MEG only the following strings should be accepted: `CTF`,  `ElektaNeuromag`, `4DBti`, `KitYokogawa`, `ChietiItab`, and `Other`

Note that `Other` is the acceptable "wildcard" [which exists for iEEG, EEG, and MEG](https://bids-specification.readthedocs.io/en/stable/99-appendices/08-coordinate-systems.html#coordinate-systems-applicable-to-meg-eeg-and-ieeg)

IF `Other` is specified as the `XXXCoordinateSystem`, the `bids-validator` SHOULD enforce, that `XXXCoordinateSystemDescription` is specified with a string of a certain length ... i.e., a free form text string that explains what `other` actually means. --> Of course, we cannot evaluate the quality of this free form description, but we can say that a reasonable description would have at least X characters (X could be e.g., 25 or something). and if the string is shorter than X characters, we warn that `XXXCoordinateSystemDescription` must contain an *exhaustive* description of the `Other` keyword.

---

practical hint: Something like this can be used:

[...]

PR #1009, opened 2020-06-30
add DataType to dataset_description.json schema

### LINKED ISSUE #1008: Support RECOMMENDED "DataType" key-value pair in dataset_description.json
was introduced with BIDS 1.4.0, see: https://github.com/bids-standard/bids-specification/pull/494

cc @effigies

rwblair:
fixes #1008

### DIFF
--- a/bids-validator/validators/json/schemas/dataset_description.json
+++ b/bids-validator/validators/json/schemas/dataset_description.json
@@ -44,6 +44,10 @@
     "DatasetDOI": {
       "type": "string"
     },
+    "DatasetType": {
+      "type": "string",
+      "enum": ["raw", "derivative"]
+    },
     "Genetics": {
       "type": "object",
       "properties": {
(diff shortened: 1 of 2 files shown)


effigies:
We should test that this doesn't raise an error if it's missing.

rwblair:
@effigies added. 

A note on recommended fields we don't issue warnings if they aren't present, we just verify that they are of the correct type if they are present.

bids-validator/validators/json/schemas/dataset_description.json
```diff
@@ -44,6 +44,10 @@
     "DatasetDOI": {
       "type": "string"
     },
+    "DatasetType": {
+      "type": "string",
+      "enum": ["raw", "derivative"]
```
sappelhoff, line 49: interesting, can this `enum` be used to only accept strings taken from a list of allowed keywords?

If so, this could be an easy fix for #945 @rwblair

rwblair: Correct, comes in pretty handy. Along similar lines more recent versions of json schema have a pattern key that can use regex if enum isn't enough in some cases:
https://json-schema.org/understanding-json-schema/reference/regular_expressions.html

PR #1013, opened 2020-07-06
Actually validate HED input

### LINKED ISSUE #967: Non-NIFTI event data not tested
The event validation does not appear to validate event data for EEG datasets, nor any non-NIFTI datasets for that matter. Lines 61 and 79 of `validators/events/events.js` in particular appear to assume NIFTI "bold" data. If NIFTI data is not in the dataset, it will not check event data associated with other filename templates.

This will need to be fixed in order to run EEG validation with the sidecar additions suggested in #942.

happy5214:
This fixes #967 by using the passed `events` parameter to source the list of event TSV files. It now properly validates HED input with a test dataset provided privately by @VisLab (slightly modified to actually use `HED` as the key for the codes).

I've also taken the liberty of removing the `issues` argument to the event validation call, as it was not listed in the parameter list for that function.

### DIFF
--- a/bids-validator/tests/events.spec.js
+++ b/bids-validator/tests/events.spec.js
@@ -22,7 +22,8 @@ describe('Events', function() {
       directory: [{ relativePath: '/stimuli/images/blue-square.jpg' }],
     }
     return validate.Events.validateEvents([], stimuli, [], {}).then(issues => {
-      assert(issues.length === 1 && issues[0].code === 77)
+      assert.strictEqual(issues.length, 1)
+      assert.strictEqual(issues[0].code, 77)
     })
   })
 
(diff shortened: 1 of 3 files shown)


rwblair:
Tried current master and this branch out against a quick events file I made and it worked branch worked well where master was silent.

We don't have an example HED dataset in bids-standard/bids-examples, @happy5214 is there a good candidate dataset we can model one off of? Master version tests against bids-examples (with its empty nifti files) doesn't appear to populate the headers variable so before this PR an example wouldn't of been viable. Also considering if we should put example broken data in bids-validator/bids-validator/tests/data/ to do a more comprehensive test.

VisLab:
I'll come up with one.  I have to look at the examples that you have.

rwblair:
@VisLab You can add a new events.tsv or update one in an existing dataset in [bids-standard/bids-examples](https://github.com/bids-standard/bids-examples/). Or add a new one if you have a dataset already put together; main thing about the example datasets is that they have the actual imaging data removed, proper nifti file names for empty files.

PR #1015, opened 2020-07-15
Allow datasets to specify HED versions or individual files as validation schema

### LINKED ISSUE #942: Need permission to update the HED validator component of BIDS for units
All,
A group of us has been working on improving the HED (https://github.com/hed-standard) event annotation schema and tools. My lab has contributed the HED validation component of the current BIDS validator.

One of the things that we have worked on is units associated with HED tags.  BIDS uses SI units (for the most part).  The HED schema had a limited list of allowed units.  We have created a new branch of the HED schema that allows all of the SI units specified by bids including multiples and sub-multiples.  We have also created a branch of the hed-validator that handles this.  We would like to merge this branch into our master and incorporate it into the bids-validator.  

The issue is that our new validator will no longer support previous versions of the HED schema. We think that this is an important step and should be done as soon as possible for the community. It should not affect datasets that have already been tagged with HED.  It mainly requires that they will need to validate against HED schemas 8.0.0 or later.   

We want to know if you foresee any issues or there are objections before we do a pull request to make this happen.

happy5214:
closes #942 

This implements the changes proposed in https://github.com/bids-standard/bids-specification/pull/527 to allow datasets to specify their HED schema version, or a specific schema file in sourcedata/, to run HED validation against. I have tested both ways with a sample dataset.


sappelhoff (changes requested):
Thanks @happy5214. I'll leave a full review of this to @rwblair ... and we should also probably wait with merging this until the PR https://github.com/bids-standard/bids-specification/pull/527 is merged.

But I'll leave a few comments:

- You need to update https://github.com/bids-standard/bids-validator/blob/master/bids-validator/validators/json/schemas/dataset_description.json so that it includes `HEDVersion`
- in https://github.com/bids-standard/bids-specification/pull/527 we are discussing that `HEDVersion` should be a *recommended* field. So perhaps it'd be nice to make BIDS-validator *warn* users if they are using HED tags ... but not supplying an `HEDVersion` field.
- I think we should have a *very basic* example dataset with HED tags in https://github.com/bids-standard/bids-examples/ ... it could be as simple as a single subject single session EEG ... with HED tags in the events.json. One idea could be to update https://github.com/bids-standard/bids-examples/tree/master/eeg_matchingpennies which is a very basic example.


edit: tagging @VisLab for attention

happy5214:
> Thanks @happy5214. I'll leave a full review of this to @rwblair ... and we should also probably wait with merging this until the PR [bids-standard/bids-specification#527](https://github.com/bids-standard/bids-specification/pull/527) is merged.
> 
> But I'll leave a few comments:
> 
> * You need to update https://github.com/bids-standard/bids-validator/blob/master/bids-validator/validators/json/schemas/dataset_description.json so that it includes `HEDVersion`
> 
> * in [bids-standard/bids-specification#527](https://github.com/bids-standard/bids-specification/pull/527) we are discussing that `HEDVersion` should be a _recommended_ field. So perhaps it'd be nice to make BIDS-validator _warn_ users if they are using HED tags ... but not supplying an `HEDVersion` field.
> 
> * I think we should have a _very basic_ example dataset with HED tags in https://github.com/bids-standard/bids-examples/ ... it could be as simple as a single subject single session EEG ... with HED tags in the events.json. One idea could be to update https://github.com/bids-standard/bids-examples/tree/master/eeg_matchingpennies which is a very basic example.
> 
> 
> edit: tagging @VisLab for attention

The issues listed have been addressed (or will be after https://github.com/bids-standard/bids-examples/pull/191 is merged). I left in the local schema code for now, but I can remove it if anyone has a strong opinion on that.

bids-validator/validators/events/hed.js
```diff
@@ -85,7 +87,26 @@ export default function checkHedStrings(events, headers, jsonContents) {
...
+          'sourcedata',
+          datasetDescription.HEDVersion,
+        )
+      }
+    } else {
+      issues.push(new Issue({ code: 130 }))
+    }
+
+    // run HED validator
+    return hedValidator.buildSchema(schemaDefinition).then(hedSchema => {
```
rwblair, line 109: I'm ok with leaving in the local schema ability, can we add a catch statement on this promise chain? If `HEDVersion` is not a semantic version and the file doesn't exist it produces an uncaught error. I think generating  a new issue here would be good. I figure most end users will have mistyped the semantic version rather than be trying to use their own schema.

I think the catch is also good in case the hed validator ever rejects its promise during build schema for whatever reason: https://github.com/hed-standard/hed-javascript/blob/f9674dfc6a18ec068260559b6d485b6ca61d0298/validators/schema.js#L374

happy5214:
Is this waiting on the example dataset? That will probably take a while to finish, so perhaps we should merge this first.

sappelhoff:
It'd be cool to give @rwblair some more time to review this - and I think bids-standard/bids-specification#527 should be merged first once discussion is done over there.

rwblair (changes requested):

bids-validator/validators/events/hed.js
rwblair: Sorry I didn't realize that I had left this unsubmitted @happy5214 I'm going to go ahead and fix the merge conflict and merge this, and create a new issue for my concern here.

PR #1019, opened 2020-07-23
ENH: Validate optional fractional seconds in AcquisitionTime in scans.tsv

### LINKED ISSUE #957: Allow fractional seconds in scans file datetimes
This stems from bids-standard/bids-specification#470, and is related to nipy/heudiconv#447 and nipy/heudiconv#451.

In bids-standard/bids-specification#470, I am proposing that fractional seconds be allowed in the accepted datetime format for the `acq_time` column in scans files. @sappelhoff pointed out that the validator will need to be updated in line with this change to the specification. The locations that need to be updated in the validator are below:

https://github.com/bids-standard/bids-validator/blob/ecacf0d30f9bc803393c1bb9833179d47a5d5451/bids-validator/validators/tsv/checkAcqTimeFormat.js#L4-L12

https://github.com/bids-standard/bids-validator/blob/920d1957c1788c96f3df5b0be3eb7c2be594f8d3/bids-validator/utils/issues/list.js#L465-L470

I can update `list.js` easily enough, but I don't know Javascript, so I'm not sure how to update `checkAcqTimeFormat.js`. Any help anyone can give would be appreciated.

sappelhoff:
closes #957 

code climate does not like my changes ... but they work. :-)

suggestions for improvements are welcome. cc @tsalo @rwblair

### DIFF
--- a/bids-validator/validators/tsv/checkAcqTimeFormat.js
+++ b/bids-validator/validators/tsv/checkAcqTimeFormat.js
@@ -1,42 +1,27 @@
 const Issue = require('../../utils').issues.Issue
-import { isValid as dateIsValid, parse } from 'date-fns'
+import { isValid as dateIsValid, parseISO } from 'date-fns'
 
 const checkAcqTimeFormat = function(rows, file, issues) {
+  const rfc3339ish = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(.\d+)?$/
   const header = rows[0]
   const acqTimeColumn = header.indexOf('acq_time')
   const testRows = rows.slice(1)
   testRows.map((line, i) => {
-    var format = "yyyy-MM-dd'T'HH:mm:ss"
     const lineValues = line
     const acqTime = lineValues[acqTimeColumn]
-    // Try different date formats: We accept any amount of optional fractional
-    // seconds: .S, .SS, .SSS, .SSSS, .SSSSS, .SSSSSS
-    var isValid = false
-    for (i = 0; i < 7; i++) {
+    let isValid = dateIsValid(parseISO(acqTime)) && rfc3339ish.test(acqTime)
 
-      isValid = dateIsValid(parse(acqTime, format, new Date()))
-
-      // exit early if date is valid or n/a
-      if (acqTime === 'n/a' | isValid) {
-        isValid = true
-        break
-      }
-
-      // If not valid, try again with parsing a new amount of fractional secs
-      if (i === 0) {
-        format += '.'
-      }
-      format += 'S'
+    if (acqTime === 'n/a') {
+      isValid = true
     }
 
-    // if the loop exits with !isValid, we need to raise an issue
     if (acqTime && !isValid) {
       issues.push(
         new Issue({
           file: file,
           evidence: acqTime,
           line: i + 2,
-          reason: 'acq_time is not in the format yyyy-MM-ddTHH:mm:ss ',
+          reason: 'acq_time is not in the format yyyy-MM-ddTHH:mm:ss[.000000]',
           code: 84,
         }),
       )
--- a/bids-validator/validators/tsv/checkAcqTimeFormat.js
+++ b/bids-validator/validators/tsv/checkAcqTimeFormat.js
@@ -2,17 +2,34 @@ const Issue = require('../../utils').issues.Issue
 import { isValid as dateIsValid, parse } from 'date-fns'
 
 const checkAcqTimeFormat = function(rows, file, issues) {
-  const format = "yyyy-MM-dd'T'HH:mm:ss"
   const header = rows[0]
   const acqTimeColumn = header.indexOf('acq_time')
   const testRows = rows.slice(1)
   testRows.map((line, i) => {
+    var format = "yyyy-MM-dd'T'HH:mm:ss"
     const lineValues = line
     const acqTime = lineValues[acqTimeColumn]
-    let isValid = dateIsValid(parse(acqTime, format, new Date()))
-    if (acqTime === 'n/a') {
-      isValid = true
+    // Try different date formats: We accept any amount of optional fractional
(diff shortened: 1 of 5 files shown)


adam2392 (approved):
A good clean simple improvement. Looks pretty good to me!

effigies:
Is there a reason that we're creating our own format string, rather than just using `parseISO`?

```Javascript
> var parseISO = require('date-fns/parseISO')
> parseISO('2017-05-03T06:45:45Z')
2017-05-03T06:45:45.000Z
> parseISO('2017-05-03T06:45:45.2Z')
2017-05-03T06:45:45.200Z
> parseISO('2017-05-03T06:45:45.22Z')
2017-05-03T06:45:45.220Z
> parseISO('2017-05-03T06:45:45.222Z')
2017-05-03T06:45:45.222Z
> parseISO('2017-05-03T06:45:45.2222Z')
2017-05-03T06:45:45.222Z
> parseISO('2017-05-03T06:45:45.22222Z')
2017-05-03T06:45:45.222Z
```

Are there values it accepts that we don't want to?

effigies:
Oh, I see. RFC3339 is a strict subset of ISO8601, and we're doing *almost* RFC3339. I think we can do this by checking that it's valid ISO8601 and then that it matches our format of RFC3339 - offset.:

```Javascript
const rfc3339ish = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(.\d+)?$/
var isValid = dateIsValid(parseISO(acq_time)) && rfc3339ish.test(acq_time)
```

Feels cleaner than a loop, IMO.

sappelhoff:
Thanks for your suggestion, I'll give it a try soon!

>  RFC3339 is a strict subset of ISO8601, and we're doing almost RFC3339

@effigies could you clarify what you mean by "almost RFC3339"? To my knowledge, we *are* following RFC3339 (at least that was the plan :thinking: )

If there is a deviation, we should consider fixing it (most recent changes with frac secs are not yet released), because IMO it's not a good idea to *almost* follow a standard (or in this case, subset of ISO).

- https://bids-specification.readthedocs.io/en/latest/02-common-principles.html#units

One (potentially not-fixable-because-breaking) divergence that I *can* see is with regards to the timezone. Let me quote from the BIDS spec:

> `YYYY-MM-DDThh:mm:ss[.000000]` This is equivalent to the RFC3339 "date-time" format, time zone is always assumed as local time).

apart from the superfluous `)`, RFC3339 actually assumes *UTC* and different timezones are supplied with an [offset](https://tools.ietf.org/html/rfc3339#section-4.2) ... whereas in BIDS we seem to assume "local time" --> and we have no way of knowing this local time in a dataset, because there is no BIDS field capturing the recording site on the earth.


---

PS: for (my own) later reference, this is a nice article: https://medium.com/easyread/understanding-about-rfc-3339-for-datetime-formatting-in-software-engineering-940aa5d5f68a

effigies:
Yeah, the "almost" refers to the fact that, in RFC3339, the time zone is required. It can be either `Z` or an explicit offset. We relax that and permit people to use local time. I think that's fine, as relative times is all a compliant BIDS app (accepting anonymized datasets) can rely on being valid, so it doesn't really matter.

bids-validator/validators/tsv/checkAcqTimeFormat.js
```diff
@@ -2,17 +2,34 @@ const Issue = require('../../utils').issues.Issue
 import { isValid as dateIsValid, parse } from 'date-fns'
 
 const checkAcqTimeFormat = function(rows, file, issues) {
-  const format = "yyyy-MM-dd'T'HH:mm:ss"
   const header = rows[0]
   const acqTimeColumn = header.indexOf('acq_time')
   const testRows = rows.slice(1)
   testRows.map((line, i) => {
+    var format = "yyyy-MM-dd'T'HH:mm:ss"
```
rwblair: Use `let` here instead of `var` if [possible](https://javascript.info/var).

bids-validator/validators/tsv/checkAcqTimeFormat.js
```diff
@@ -2,17 +2,34 @@ const Issue = require('../../utils').issues.Issue
...
     const acqTime = lineValues[acqTimeColumn]
-    let isValid = dateIsValid(parse(acqTime, format, new Date()))
-    if (acqTime === 'n/a') {
-      isValid = true
+    // Try different date formats: We accept any amount of optional fractional
+    // seconds: .S, .SS, .SSS, .SSSS, .SSSSS, .SSSSSS
+    var isValid = false
+    for (i = 0; i < 7; i++) {
+
+      isValid = dateIsValid(parse(acqTime, format, new Date()))
```
rwblair: A note if we ever run into performance concerns, parse with format "yyyy-MM-dd'T'HH:mm:ss.SSSSSSS" will work on a date with fractional seconds up to 7. It won't return the correct fraction but I believe its functionally equivalent for testing validity:

> dateIsValid(parse("2017-05-03T06:45:45.9", format, new Date()))
true
> parse("2017-05-03T06:45:45.9", format, new Date())
2017-05-03T11:45:45.000Z

rwblair (changes requested):
Just the 'var'->'let' update. Other wise looks good to me.

sappelhoff:
@rwblair are you recommending against @effigies's suggestion in https://github.com/bids-standard/bids-validator/pull/1019#issuecomment-664095375 ?

rwblair:
> @rwblair are you recommending against @effigies's suggestion in [#1019 (comment)](https://github.com/bids-standard/bids-validator/pull/1019#issuecomment-664095375) ?

@sappelhoff  Sorry for the confusion, no I am not. Was referring only to the changes in my code review. His suggestion seems reasonable if it works just as well as the current implementaiton.

sappelhoff:
The tests for TSV spec pass locally and @effigies solution is nice and elegant IMO.

A bunch of unrelated tests seem to fail however.

rwblair (approved):

effigies (approved):

PR #1035, opened 2020-08-03
[WIP] bep_001 updates.

rwblair:
First commit covers the new entity table in bep 001 spec.

Still need to add all the additional validation in qmri appendix and tests.


rwblair:
Out of date with respect to spec and regex changes.

PR #1038, opened 2020-08-04
Make beh and func file level rules consistent with spec.

### LINKED ISSUE #1036: Should acq and rec be supported for behavior-only data?
This stems from https://github.com/bids-standard/bids-specification/pull/556#issuecomment-667694066. As part of https://github.com/bids-standard/bids-specification/pull/556, we noticed that the validator supports `acq` and `rec` entities for the `beh` datatype, even though the specification does not.

Is it possible that this is an oversight on the validator side, or should the specification include support for `acq` and `rec`?

Here are the relevant rules, for reference:

https://github.com/bids-standard/bids-validator/blob/8317fb5638dfa6d9907116e1fcfdde7d35e96a7d/bids-validator/bids_validator/rules/file_level_rules.json#L58-L72

rwblair:
Attempt to fix #1036 

@effigies @tsalo Are these changes along what we were thinking?

### DIFF
--- a/bids-validator/bids_validator/rules/file_level_rules.json
+++ b/bids-validator/bids_validator/rules/file_level_rules.json
@@ -73,18 +73,6 @@
     }
   },
 
-  "cont": {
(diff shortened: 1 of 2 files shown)


effigies (approved):
This looks right to me.

tsalo (approved):
It's a little harder to read for a JS/regex novice like me, but the changes make sense to me.

sappelhoff (approved):

sappelhoff:
Thanks @rwblair

PR #1040, opened 2020-08-06
check continous recording datafiles have matching json.

### LINKED ISSUE #1024: physio JSON is required when physio TSV.GZ is supplied
see: https://github.com/bids-standard/bids-examples/issues/186

IMHO, physio JSON is REQUIRED when a TSV is supplied. But this is not enforced by the validator.

May also need a clarification about that in bids-specification. see: https://github.com/bids-standard/bids-specification/pull/542

rwblair:
Also add schema for physio and stim json files so that they will be validated with required fields during json validation.
Fixes #1024

Still working on tests.

### DIFF
--- a/bids-validator/validators/json/json.js
+++ b/bids-validator/validators/json/json.js
@@ -111,6 +111,10 @@ const selectSchema = file => {
         file.name.endsWith('genetic_info.json')
     ) {
       schema = require('./schemas/genetic_info.json')
+    } else if (
+        file.name.endsWith('physio.json')
+    ) {
+      schema = require('./schemas/physio.json')
     }
   }
   return schema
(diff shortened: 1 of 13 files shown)


sappelhoff:
great to see some progress here, thanks for taking the time @rwblair! 

This PR will probably ~break some of the bids-examples~ reveal that some bids-examples are broken: https://github.com/bids-standard/bids-examples/issues/186

bids-validator/validators/json/json.js
```diff
@@ -111,6 +111,10 @@ const selectSchema = file => {
         file.name.endsWith('genetic_info.json')
     ) {
       schema = require('./schemas/genetic_info.json')
+    } else if (
+        file.name.endsWith('physio.json')
```
effigies: Do we need to have `stim.json`, as well?

rwblair: Yes, good catch.

PR #1051, opened 2020-09-02
Update top level file regex to allow modality metadata files with no required entities.

### LINKED ISSUE #1039: Problem with top level coordsystem.json and events.json
Fix in top level rules and in json validator:
https://github.com/bids-standard/bids-specification/issues/528#issuecomment-669951872

rwblair:
fixes #1039

### DIFF
--- a/bids-validator/bids_validator/rules/top_level_rules.json
+++ b/bids-validator/bids_validator/rules/top_level_rules.json
@@ -17,18 +17,18 @@
       ]
     }
   },
-  
+
   "func_top": {
-    "regexp": "^[\\/\\\\](?:ses-[a-zA-Z0-9]+_)?(?:recording-[a-zA-Z0-9]+_)?task-[a-zA-Z0-9]+(?:_acq-[a-zA-Z0-9]+)?(?:_ce-[a-zA-Z0-9]+)?(?:_dir-[a-zA-Z0-9]+)?(?:_rec-[a-zA-Z0-9]+)?(?:_run-[0-9]+)?(?:_echo-[0-9]+)?(@@@_func_top_ext_@@@)$",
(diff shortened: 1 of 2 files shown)


rwblair:
@sappelhoff before this func/eeg/ieeg/meg metadata files needed a task name in the root of a dataset, now all entities are optional there. Is this what we want?

@tsalo  I don't think these top level metadata files aren't currently captured by the schema in the specification, you have any intuition on the best way to include them; New field in each of the "datatypes/*.yaml"s, or maybe something in top_level_files?

sappelhoff:
In the last maintainers call @effigies made the point that the spec is explicit about that a "task" label is pretty much always required.

I checked a bit, and it seems to be true at least for *some* files (e.g., `*_eeg.json`).

You also mentioned that the behavior in this PR could lead to problems e.g., with `coordsystem.json` -> https://github.com/bids-standard/bids-specification/issues/528#issuecomment-669959132

> Is this what we want?

I am beginning to think that we are shooting ourselves in the foot with these changes.

So I am now inclined to close my issues :grimacing: sorry for not communicating this to you earlier!

tsalo:
> @tsalo I don't think these top level metadata files aren't currently captured by the schema in the specification, you have any intuition on the best way to include them; New field in each of the "datatypes/*.yaml"s, or maybe something in top_level_files?

To be honest, I had hoped that the inheritance logic (which isn't represented anywhere in the schema) would naturally be able to handle files like this. Unfortunately, I don't know much about the details of the inheritance rule; I basically just know the broad strokes. Do the inheritable versions of the files need to be represented in the schema?

rwblair:
@tsalo You're right, I wasn't thinking clearly. For the script that generates the regex I'll update it to just output additional rules for top level files for any json, tsv it rules in finds in the yaml so we won't need any changes to the yaml.

PR #1053, opened 2020-09-03
Extend validator for crosstalk+finecalib for FIF

### LINKED ISSUE #1047: update validator to allow *_acq-crosstalk.dat files for MEG
The changes in the spec from https://github.com/bids-standard/bids-specification/pull/581 require validator adjustments:

1. `.dat` becomes a valid extension for MEG data ... but ONLY if `acq-crosstalk` is part of the file name
1. `*_acq-calibration_meg.fif` is valid ... but I think we do not need adjustments for this

crossref of original issue: https://github.com/mne-tools/mne-bids/issues/495

cc @hoechenberger

sappelhoff:
closes #1047 

Here I added some tests that must pass --> i.e., as long as they don't, we need to change the validator.

@hoechenberger please look at my suggested "good" and "bad" filenames --> do you think this is fine when reading the specification here https://github.com/bids-standard/bids-specification/pull/581 and when considering your experience?

I think it should be fair enough to allow people to store their crosstalk+finecalibration files:

1. at the root
1. for each subject / session

Can you think of other sensible places where to enable storage?

Also, can you think of other "bad filenames" (see code), if so, then we can add them to the tests to allow for better test driven development.

In any case, we should implement this part of the validator BEFORE the next release, so that if we find issues with the current spec for crosstalk+finecalib we can still adjust it without breakage issues.

cc @rwblair 

PS: Everybody please feel free to push to this PR

---

EDIT: new related PR: https://github.com/bids-standard/bids-specification/pull/598


sappelhoff:
Note, if the current spec for these two files turns out to be a total pain, we could still make some adjustments to the spec -> https://github.com/bids-standard/bids-specification/pull/581/files

That is: we could constrain a bit more on how to store the files, if that makes our lives easier for the validation!

(For example, require that the two files MUST be stored on the subject/session/meg level)

bids-validator/tests/type.spec.js
```diff
@@ -206,6 +206,11 @@ describe('utils.type.file.isMEG', function() {
     // NO father dir: fif data
     '/sub-01/ses-001/meg/sub-01_ses-001_task-rest_run-01_split-01_meg.fif',
     '/sub-01/ses-001/meg/sub-01_ses-001_task-rest_acq-TEST_run-01_split-01_meg.fif',
+    // cross-talk and fine-calibration files for Neuromag/Elekta/MEGIN data (.fif)
+    '/acq-calibration_meg.fif',
+    '/acq-crosstalk_meg.dat',
+    '/sub-01/ses-001/meg/sub-01_ses-001_acq-calibration_meg.fif',
+    '/sub-01/ses-001/meg/sub-01_ses-001_acq-crosstalk_meg.dat',
```
hoechenberger:
```suggestion
    // cross-talk and fine-calibration files for Neuromag/Elekta/MEGIN data (.fif)
    '/acq-calibration_meg.fif',
    '/acq-crosstalk_meg.dat',
    '/sub-01/sub-01_acq-calibration_meg.fif',
    '/sub-01/sub-01_acq-crosstalk_meg.dat',
    '/sub-01/ses-001/meg/sub-01_ses-001_acq-calibration_meg.fif',
    '/sub-01/ses-001/meg/sub-01_ses-001_acq-crosstalk_meg.dat',
```

hoechenberger: Just adding here what is acceptable according to the specification (as I understand it). In the next iteration, I will reconsider which of these actually make sense for us :)

hoechenberger:
> I think it should be fair enough to allow people to store their crosstalk+finecalibration files:
> 
> 1. at the root
> 2. for each subject / session
> 
> Can you think of other sensible places where to enable storage?

I think your list covers all reasonable storage locations

bids-validator/tests/type.spec.js
sappelhoff: ah thanks -> but I think you are missing the `meg` folder in 
```diff
-    '/sub-01/sub-01_acq-calibration_meg.fif',
-    '/sub-01/sub-01_acq-crosstalk_meg.dat',
+    '/sub-01/meg/sub-01_acq-calibration_meg.fif',
+    '/sub-01/meg/sub-01_acq-crosstalk_meg.dat',
```

right?

hoechenberger: You see, this is what I was slightly confused about. Thanks for bringing this up.

I can only have the `meg` folder if I also supply a `session` folder (in case we have a `ses` entity), right? But what if I want to have the same cross-talk and fine-calibration files for _all_ sessions within an individual participant? This is what I was trying to achieve here.

sappelhoff: > I can only have the meg folder if I also supply a session folder 

no you can also completely drop "session" (e.g., if there is only a single session)

> what if I want to have the same cross-talk and fine-calibration files for all sessions within an individual participant?

yeah, so THAT's what we would actually want :thinking:

hoechenberger: > > I can only have the meg folder if I also supply a session folder 
> 
> 
> 
> no you can also completely drop "session" (e.g., if there is only a single session)
> 

Yes that's what I meant, sorry for bad suboptimal phrasing :)

> 
> 
> > what if I want to have the same cross-talk and fine-calibration files for all sessions within an individual participant?
> 
> 
> 
> yeah, so THAT's what we would actually want :thinking: ￼

Well, I haven't made up my mind if we actually WANT that, but it's something the specs currently allow for.

I would just say, let's try to get the validation to pass, and if we discover that it's a pain, we change the specs, as you suggested?

hoechenberger: @sappelhoff I'm currently looking into this.

sappelhoff: > I would just say, let's try to get the validation to pass, and if we discover that it's a pain, we change the specs, as you suggested?

:+1: 

> @sappelhoff I'm currently looking into this.

good luck and check out:

- https://github.com/bids-standard/bids-validator/blob/master/CONTRIBUTING.md#using-the-development-version-of-bids-validator
- https://github.com/bids-standard/bids-validator#testing

:-)

hoechenberger:
@sappelhoff I cannot push to your fork, could you grant me permission please?
You should see a checkbox on the right that looks sth like this:

Screenshot 2020-09-04 at 20 07 48

sappelhoff:
> could you grant me permission please?

done, I gave you (temporary) maintainer permissions :-)

hoechenberger:
> done, I gave you (temporary) maintainer permissions :-)


Neat 🥳
Just pushed

Instead of extending the existing regexp and making it even harder to parse for a human brain, I simply added two new regexps for fine-calibration and cross-talk files

hoechenberger:
Say… I just realized that modality-specific files like `*_meg.*` **have** to be placed in `/sub-<label>/[ses-<label>/]meg/`, according to https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/02-magnetoencephalography.html#meg-recording-data, is that correct? In that case, I can drop half of the changes I've made here…

sappelhoff:
> I just realized that modality-specific files like *_meg.* have to be placed in /sub-<label>/[ses-<label>/]meg/

yes, I think that's correct. @robertoostenveld can you weigh in with some advice on how you would store crosstalk+calibration files for FIF, please?

This looks a bit odd to me:

> /sub-12/sub-12_acq-calibration_meg.fif

What I think we should make valid (just finecalib examples here, crosstalk should be the same):

- `acq-calibration_meg.fif` ... define at the root ... valid for everybody and all sessions (--> how common is this case?! should we drop this?)
- `/sub-12/meg/sub-12_acq-calibration_meg.fif` ... if no session folders are present (=there is only one session ... --> this boils down to 1xcalib and 1xcrosstalk per subj ... probably leads to the same files being duplicated at different subj levels, e.g., because subjs have been measured on the same day .. or even the same week/month/...?)
- `/sub-12/ses-01/meg/sub-12_ses-01_acq-calibration_meg.fif` ... define separately for each session (--> e.g., if one subject has been measured one at time X and once at time X+6months)

I think what's missing from this and what you @hoechenberger wanted is 

- `/sub-12/sub-12_acq-calibration_meg.fif` ... store at subject level --> valid for all sessions of that subject

but that collides with the "meg" stuff should be stored in `.../meg...` (although arguably the `acq-calibration_meg.fif` defined at the root collides with that rule as well)

hoechenberger:
> * `acq-calibration_meg.fif` ... define at the root ... valid for everybody and all sessions (--> how common is this case?! should we drop this?)

This is actually **quite** common, as these files typically only change with system service every 6 months or so. So **if** we drop this, we should do it for other reasons than "it's not a common use case" 😅



> * `/sub-12/meg/sub-12_acq-calibration_meg.fif` ... if no session folders are present (=there is only one session ... --> this boils down to 1xcalib and 1xcrosstalk per subj ... probably leads to the same files being duplicated at different subj levels, e.g., because subjs have been measured on the same day .. or even the same week/month/...?)

I don't understand this, actually. Assume I have a long-running study with only a single session for each participant, but with a large number of participants. In that case, it's possible that the files **do** change between participants.



> * `/sub-12/ses-01/meg/sub-12_ses-01_acq-calibration_meg.fif` ... define separately for each session (--> e.g., if one subject has been measured one at time X and once at time X+6months)

Yep, so this one is clear and crisp 😁



> I think what's missing from this and what you @hoechenberger wanted is
> 
> * `/sub-12/sub-12_acq-calibration_meg.fif` ... store at subject level --> valid for all sessions of that subject

Yeah, this is the one you pointed out as looking a little odd.



> but that collides with the "meg" stuff should be stored in `.../meg...` (although arguably the `acq-calibration_meg.fif` defined at the root collides with that rule as well)

Yep. 

I mean. To make our lives super easy, we could decide to only allow 
- `/sub-12/meg/sub-12_acq-calibration_meg.fif`
- `/sub-12/ses-01/meg/sub-12_ses-01_acq-calibration_meg.fif`

But let's see what @robertoostenveld thinks :)

sappelhoff:
> > `acq_calibration_meg.fif` (at root level)

> This is actually quite common, as these files typically only change with system service every 6 months or so

I'd be fine to allow it :)

> > * `/sub-12/meg/sub-12_acq-calibration_meg.fif` ... if no session folders are present (=there is only one session ... --> this boils down to 1xcalib and 1xcrosstalk per subj ... probably leads to the same files being duplicated at different subj levels, e.g., because subjs have been measured on the same day .. or even the same week/month/...?)

> I don't understand this, actually. Assume I have a long-running study with only a single session for each participant, but with a large number of participants. In that case, it's possible that the files **do** change between participants.

this kind of syntax would also cover studies where the calib/crosstalk changes mid study:

1. `/sub-12/meg/sub-12_acq-calibration_meg.fif`
2. crosstalk/calib are changed through maintenance
3. `/sub-13/meg/sub-13_acq-calibration_meg.fif` ... the contents of this file are different from that of sub-12


> To make our lives super easy, we could decide to only allow
>  /sub-12/meg/sub-12_acq-calibration_meg.fif
>  /sub-12/ses-01/meg/sub-12_ses-01_acq-calibration_meg.fif

True ... that's tempting as well.

However we decide, I am +1 on adjusting the spec from https://github.com/bids-standard/bids-specification/pull/581 to include a more complete list of examples on how we want users to specify crosstalk + finecalibration

hoechenberger:
> However we decide, I am +1 on adjusting the spec from [bids-standard/bids-specification#581](https://github.com/bids-standard/bids-specification/pull/581) to include a more complete list of examples on how we want users to specify crosstalk + finecalibration

So now I gotta ask: what's the "BIDS philosophy" when addressing questions like these? I, personally, would be totally fine requiring these paths:

```
/sub-12/meg/sub-12_acq-calibration_meg.fif
/sub-12/ses-01/meg/sub-12_ses-01_acq-calibration_meg.fif
```
and not allowing anything else. Super clear, super restrictive. Would lead to some redundancy in most studies (storing the exact same files a bunch of times, for each session / participant)

But that's not how the BIDS specification is usually designed, right?

sappelhoff:
> Super clear, super restrictive.

would be fine with me in principle

> Would lead to some redundancy in most studies (storing the exact same files a bunch of times, for each session / participant)

you see redundancy in a lot of places in BIDS ... and yet in other places you find the inheritance principle to reduce that redundancy.

In our case I would consider this kind of redundancy fine. 

> But that's not how the BIDS specification is usually designed, right?

Just a side note: BIDS is continuously evolving and growing, so its initial design (whatever it was) has long been overhauled, changed, extended --> that's one of BIDS's drawbacks and advantages: We can accommodate a lot and we can work with the community and adapt ... at the expense of sometimes being clunky and not as strict as other standards are.

hoechenberger:
Ok. I've decided that I think the following would be good to have:
```
/acq-calibration_meg.fif
/sub-12/meg/sub-12_acq-calibration_meg.fif
/sub-12/ses-01/meg/sub-12_ses-01_acq-calibration_meg.fif
```
Will adjust the code tonight, and then we'll also have to revise the specification

hoechenberger:
> * `/sub-12/sub-12_acq-calibration_meg.fif` ... store at subject level --> valid for all sessions of that subject
> 
> but that collides with the "meg" stuff should be stored in `.../meg...` (although arguably the `acq-calibration_meg.fif` defined at the root collides with that rule as well)

This dataset:
https://github.com/OpenNeuroDatasets/ds000248

also has a bunch of `acq-` files related to MRI recordings in the root, so I guess it would make sense to support this for MEG too.

hoechenberger:
… and considering all of that. Maybe we can just leave things (in this PR) as they are, and only need to adjust the specs? To recap, the supported paths are:

```
/acq-calibration_meg.fif
/ses-01_acq-calibration_meg.fif
/sub-12/sub-12_acq-calibration_meg.fif <-- this one "feels" a little odd, but would apply to all sessions w/in a subject
/sub-12/meg/sub-12_acq-calibration_meg.fif
/sub-12/ses-01/meg/sub-12_ses-01_acq-calibration_meg.fif
```

sappelhoff:
but they are `.bidsignore`..d --> https://github.com/OpenNeuroDatasets/ds000248/blob/a42fc3f4759ef6bda8253aea4e9403a0cc3a066d/.bidsignore#L1

😬 

Over the course of this discussion and sleeping over it, I am now tending towards making our lives easy at first, and only allow:

```
/sub-12/meg/sub-12_acq-calibration_meg.fif
/sub-12/ses-01/meg/sub-12_ses-01_acq-calibration_meg.fif
```

Pros:
- this "style" of saving MEG files is widely used and spec conform
- crosstalk+finecalib can be stored for each participant/session combination --> there is no case where we cannot store the data
- no need to have a new, far-reaching discussion with all kinds of BIDS stakeholders

Cons:
- leads to duplication of files

Action items:
- adjust this  PR
- adjust the spec to be more restrictive

Sunday morning thought:
- Implementing the specification in the validator really *ALWAYS* lets you look upon the situation from a new angle ... and often reveals flaws ... that's a good lesson to avoid changes that you are not sure how to validate.

PS:
- doing this "easy" way does not prevent us from extending it in the future in a backwards compatible way ... however, doing "too much" now may prove harmful.

hoechenberger:
@sappelhoff I agree with everything you said

I will do the adjustments to this PR later today, and propose adjustments to the specs where needed.

We can always extend the specs later to include more storage places should it turn out to be required. Until then, maybe it's super clean to store all MEG stuff under `meg/`.

hoechenberger:
@sappelhoff 
Done, we're back to only allowing "file-level" paths, i.e.,
```
/sub-12/meg/sub-12_acq-calibration_meg.fif
/sub-12/ses-01/meg/sub-12_ses-01_acq-calibration_meg.fif
```
as discussed.

sappelhoff (commented):
awesome, I can't approve technically because GitHub counts this as my own PR, but I consider this +1

Thanks a bunch! ... but before merging, I want @rwblair to have a look at this and approve or suggest changes.

bids-validator/bids_validator/rules/file_level_rules.json
```diff
@@ -248,6 +248,14 @@
     }
   },
 
+  "meg_calbibration": {
+    "regexp": "^[\\/\\\\](sub-[a-zA-Z0-9]+)[\\/\\\\](?:(ses-[a-zA-Z0-9]+)?)[\\/]?meg[\\/]\\1(_\\2)_acq-calibration_meg\\.fif$"
```
rwblair: Semantics of the session matching group are interesting. In other regex we have `(?:(ses-[a-zA-Z0-9]+)[\\/\\\\])?` where the whole session group and then the subsequent regerence to the match is listed as optional like `(_\\2)?`

This regex makes the first session group optional inside of a non capturing group, and then the reference to the second match is non optional, and is a capturing group. It works fine if session is present in the filename, but I haven't been able to get it to work locally on ones without a session:
```
> re = "[\\/\\\\](sub-[a-zA-Z0-9]+)[\\/\\\\](?:(ses-[a-zA-Z0-9]+)?)[\\/]?meg[\\/]\\1(_\\2)_acq-calibration_meg\\.fif"
'[\\/\\\\](sub-[a-zA-Z0-9]+)[\\/\\\\](?:(ses-[a-zA-Z0-9]+)?)[\\/]?meg[\\/]\\1(_\\2)_acq-calibration_meg\\.fif'
> '/sub-01/meg/sub-01_acq-calibration_meg.fif'.match(re)
null
> 
```

If we make the regex look similar to other session matches it does go off:

```

> re = "[\\/\\\\](sub-[a-zA-Z0-9]+)[\\/\\\\](?:ses-[a-zA-Z0-9]+)?[\\/]?meg[\\/]\\1(_\\2)?_acq-calibration_meg\\.fif"
'[\\/\\\\](sub-[a-zA-Z0-9]+)[\\/\\\\](?:ses-[a-zA-Z0-9]+)?[\\/]?meg[\\/]\\1(_\\2)?_acq-calibration_meg\\.fif'
> '/sub-01/meg/sub-01_acq-calibration_meg.fif'.match(re)
[
  '/sub-01/meg/sub-01_acq-calibration_meg.fif',
  'sub-01',
  undefined,
  index: 0,
  input: '/sub-01/meg/sub-01_acq-calibration_meg.fif',
  groups: undefined
]
> 
```

Good examples only test against files with session present so they were able to pass as expected.

rwblair (changes requested):

bids-validator/bids_validator/rules/file_level_rules.json
hoechenberger: Nice catch @rwblair, will adjust the PR

hoechenberger: @rwblair I have addressed this issue in acf479b. There was still a small bug in your proposed regexp (we **do** need to capture the session entity to re-use it later), other than that it was all good.

rwblair (approved):

hoechenberger:
Awesome, thanks @rwblair and @sappelhoff!

sappelhoff:
And thanks @hoechenberger as well :-)

robertoostenveld:
I did not have time earlier, but just read up on this. The proposed solution under "meg"  (for now) looks good to me and can always be improved/extended in the future. 

Perhaps for future consideration is that this `acq-calibration_meg.fif` file might be compared to the bvec and bval files for dwi. Internal consistency is a general aspect of the  "BIDS philosophy", so if a solution is defined for case 1, and case 2 is similar, the same solution should be adopted for 2 as well.

PR #1054, opened 2020-09-03
allow string n/a for PowerLineFrequency in EEG,MEG,iEEG

### LINKED ISSUE #1046: Modality-specific electrophysiology json sidecar files PowerLineFrequency required to be an integer
After testing the data used for testing (20 seconds) compared to random data, the difference between the 50 Hz and 60 Hz power was not enough not to classify random data incorrectly a significant portion of the time. This means that the powerline frequency cannot be determined reliably from the data https://github.com/mne-tools/mne-bids/pull/505, and so should be given as 'n/a'. 'n/a' is not accepted by the validator, only an integer (-1 for now). The major downside to requiring the powerline frequency be not missing, as discussed, is that if it is automatically selected and wrong there is the potential to lead to spurious results, whereas if it is not specified than it is at the discretion of the person conducting the analysis to determine where to filter to remove line noise, and if there is not an obvious peak at 50 or 60 Hz then they would need to follow up or take into account that their analysis shouldn't include these frequencies.

sappelhoff:
closes  #1046

PowerLineFrequency must now be an "integer" ... not a "number" --> i.e., funky things like `-3.14` are no longer allowed.

We expect this to be 50(Hz) or 60(Hz) in almost all cases.

In some cases, the PowerLineFrequency may not be known anymore and may be hard/noisy to estimate --> for these cases it is now permitted to specify `n/a` as a string.

To preempt a comment that we should not allow `n/a` as a value for a REQUIRED field (which PowerLineFrequency is) --> maybe you are right, but the current state of affairs is that all string fields also permit `n/a` ... PowerLineFrequency was only an "unlucky" field, because it's a number / integer, not a string.

cc @alexrockhill

### DIFF
--- a/bids-validator/tests/json.spec.js
+++ b/bids-validator/tests/json.spec.js
@@ -84,7 +84,7 @@ describe('JSON', function() {
       SoftwareFilters: {
         HighPass: { HalfAmplitudeCutOffHz: 1, RollOff: '6dB/Octave' },
       },
-      PowerLineFrequency: 50,
+      PowerLineFrequency: "n/a",
     }
     jsonDict[eeg_file.relativePath] = jsonObj
     validate.JSON(eeg_file, jsonDict, function(issues) {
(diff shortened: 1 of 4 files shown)


alexrockhill:
Thanks @sappelhoff! Good thing we're not using -1 anymore ;)

Issue #1072, opened 2020-09-16
consider improving all schemas: fields that require a URI can be validated accordingly


sappelhoff:
We have some fields in the schemas like `CogPOID` that require a URL or a URI.

While the JSON schema does not directly allow for validating URLs, it seems like there is some functionality to check general uris --> https://json-schema.org/understanding-json-schema/reference/string.html?highlight=uri#uri-template

This might be worth looking into for stronger validation of several fields

ref: https://bids-specification.readthedocs.io/en/latest/04-modality-specific-files/03-electroencephalography.html#sidecar-json-_eegjson

list to be done:

- [x] CogAtlasID
- [x] CogPOID
- [x] Genetics.Dataset
- [x] Genetics.Database

in the future:

- Also lots of stuff in `GeneratedBy` object from derivatives, cc @effigies 

also cc @Remi-Gau to show him the benefits we now gain due to his recent work :+1:

PR #1076, opened 2020-09-18
Add JSON schema for events.json files

### LINKED ISSUE #1032: events.json does not have JSON schema
A schema should probably be added to assure proper validation.

happy5214:
Most of the TSV-related data was copied from `data_dictionary.json`.

This does not include the changes currently in progress to add value HED tags to complement the categorical tags (currently in https://github.com/VisLab/bids-specification/blob/hed-update/src/99-appendices/03-hed.md).

Fixes #1032

### DIFF
--- a/bids-validator/validators/json/schemas/events.json
+++ b/bids-validator/validators/json/schemas/events.json
@@ -61,7 +61,8 @@
           "type": "string"
         },
         "Code": {
-          "type": "string"
+          "type": "string",
+          "format": "uri"
         }
       }
     }
--- a/bids-validator/validators/json/schemas/events.json
+++ b/bids-validator/validators/json/schemas/events.json
@@ -64,7 +64,8 @@
           "type": "string"
         },
         "SoftwareRRID": {
-          "type": "string"
+          "type": "string",
+          "pattern": ".+_.+"
         },
         "SoftwareVersion": {
           "type": "string"
--- a/bids-validator/validators/json/schemas/events.json
+++ b/bids-validator/validators/json/schemas/events.json
@@ -32,14 +32,23 @@
           }
         },
         "HED": {
-          "type": "object",
-          "title": "HED",
-          "description": "For categorical variables: a dictionary of possible values (keys) and their HED strings (values).",
-          "patternProperties": {
-            "^.+$": {
-              "type": "string"
+          "anyOf": [
+            {
+              "type": "object",
+              "title": "HED",
+              "description": "For categorical variables: a dictionary of possible values (keys) and their HED strings (values).",
+              "patternProperties": {
+                "^.+$": {
+                  "type": "string"
+                }
+              }
+            },
+            {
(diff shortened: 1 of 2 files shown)


happy5214:
I mention the value HED tag additions because the spec version currently given in that draft is not what I had in mind. I had expected it to be a regular string instead of the objects used for the categorical columns, hence needed changes to the new schema.

happy5214:
The draft spec over on the VisLab fork has been fixed. The required change will be to allow a string value for the HED property in addition to an object.

happy5214:
I'm converting this to a draft until #1078 is merged, at which time I'll add the necessary type changes to the `HED` field in this schema.

bids-validator/validators/json/schemas/events.json
```diff
@@ -0,0 +1,69 @@
...
+      "type": "object",
+      "properties": {
+        "OperatingSystem": {
+          "type": "string"
+        },
+        "SoftwareName": {
+          "type": "string"
+        },
+        "SoftwareRRID": {
+          "type": "string"
```
sappelhoff: this can be validated in a stronger way --> it MUST be an RRID, so we can validate a specific format instead of only a type "string"

using e.g., https://json-schema.org/understanding-json-schema/reference/regular_expressions.html

I don't know exactly about *what* format RRIDs have, but perhaps they always look like this: "SCR_002881" ???

i.e.,
- leading SCR
- then underscore
- then 6 digits
?

bids-validator/validators/json/schemas/events.json
```diff
@@ -0,0 +1,69 @@
...
+        },
+        "SoftwareVersion": {
+          "type": "string"
+        },
+        "Code": {
+          "type": "string"
```
sappelhoff: this field can also be stronger validated ... it must be specified as a URI, see https://json-schema.org/understanding-json-schema/reference/string.html?highlight=uri#uri-template

also discussed in #1072

bids-validator/validators/json/schemas/events.json
happy5214: RRIDs are listed at https://scicrunch.org/resources. While that is not the general format of RRIDs, all of the tools in the portal happen to be from the SciCrunch registry, and thus they have that format. Whether that's enough to enforce that format for the regex is up to you, but the only absolutely general rule is the underscore separation.

bids-validator/validators/json/schemas/events.json
happy5214: Done.

bids-validator/validators/json/schemas/events.json
sappelhoff: > but the only absolutely general rule is the underscore separation

then let's go with the only absolutely general rule ... so probably something like `.+_.+`

happy5214: Done.

Issue #1070, opened 2020-09-16
FIX ephys schema: "epoch length" should only accept positive number, not string


sappelhoff:
evidence:

https://github.com/bids-standard/bids-validator/blob/4354d0338b6bdae40b7a038f444b5643038f6e15/bids-validator/validators/json/schemas/eeg.json#L49

only a positive number should be accepted

ref: https://bids-specification.readthedocs.io/en/latest/04-modality-specific-files/03-electroencephalography.html#sidecar-json-_eegjson

Issue #1068, opened 2020-09-16
ENH ephys schemas --> all "COUNT" keys should validate that number >= 0


sappelhoff:
The meg, eeg, ieeg schemas in https://github.com/bids-standard/bids-validator/tree/master/bids-validator/validators/json/schemas

should get an update with the `"minimum": 0` schema syntax, see [docs](https://json-schema.org/understanding-json-schema/reference/numeric.html#range) for the fields where it applies. These are namely all fields that end on `Count`:

For example 

https://github.com/bids-standard/bids-validator/blob/4354d0338b6bdae40b7a038f444b5643038f6e15/bids-validator/validators/json/schemas/eeg.json#L23-L28

Issue #1066, opened 2020-09-15
RecordingType in ephys schemas needs to only accept 3 str keywords, but accepts all strs


sappelhoff:
in ieeg

https://github.com/bids-standard/bids-validator/blob/f8dc8c75d0357a3964d4d82ad4ef57337e574b70/bids-validator/validators/json/schemas/ieeg.json#L43

also in meg

https://github.com/bids-standard/bids-validator/blob/f8dc8c75d0357a3964d4d82ad4ef57337e574b70/bids-validator/validators/json/schemas/meg.json#L54

and in eeg

https://github.com/bids-standard/bids-validator/blob/f8dc8c75d0357a3964d4d82ad4ef57337e574b70/bids-validator/validators/json/schemas/eeg.json#L48

HOWEVER, only "continuous", "epoched" and "discontinuous" should be accepted.

see spec:

> RECOMMENDED. Defines whether the recording is continuous, discontinuous or epoched

however, we'll also have to mention explicitly in the MEG spec that "discontinuous" is also allowed

link to spec: https://bids-specification.readthedocs.io/en/latest/04-modality-specific-files/03-electroencephalography.html#sidecar-json-_eegjson

Issue #1065, opened 2020-09-15
`GeneticLevel` in schema expects array of str, but should ALSO accept a string


sappelhoff:
EDIT: see discussion below --> this issue should be solved by allowing a string AND an array of strings

evidence in code:

https://github.com/bids-standard/bids-validator/blob/f8dc8c75d0357a3964d4d82ad4ef57337e574b70/bids-validator/validators/json/schemas/genetic_info.json#L4-L17

however, the spec says: 

> REQUIRED. String. Describes the level of analysis. Values MUST be one of Genetic, Genomic, Epigenomic, Transcriptomic, Metabolomic, or Proteomic.

see: https://bids-specification.readthedocs.io/en/latest/04-modality-specific-files/08-genetic-descriptor.html#genetic-information

Is this an error in the spec or in the validator @CPernet ? Would you want people to specify something like `["Genetic", "Genomic"]` ... or should it be either `"Genetic"` **or** `"Genomic"`?

would need to be fixed here: https://github.com/bids-standard/bids-examples/tree/master/genetics_ukbb

effigies:
This was set in a reformatting: https://github.com/bids-standard/bids-specification/commit/db8cb3747606fbe41b9ab64edab51c695ed8141f

Hopefully the semantic change was reviewed. The example only shows a string, not an array.

CPernet:
but an array e.g. ["Genetic", "Genomic"]  would be better - several analytic levels can be performed on the same data

CPernet:
- an array makes more sense
- ukbb example is fine nothing to fix
- I guess the previous iteration bids-standard/bids-specification@db8cb37 without the 'one of' was better allowing array

effigies:
Thanks for the clarification, @CPernet. @sappelhoff I think for backwards compatibility, we should permit a string or array of strings.

sappelhoff:
okay, sounds good to me - practically we'd then adjust the schema to also accept str ... and I will clarify the spec in https://github.com/bids-standard/bids-specification/pull/605

CPernet:
we also need to add in the spec the new reference :-)

`Clara Moreau, Martineau Jean-Louis; Ross Blair; Christopher Markiewicz, Jessica Turner, Vince Calhoun, Thomas Nichols, Cyril Pernet (2020) The genetics-BIDS extension: Easing the search for genetic data associated with human brain imaging. GigaScience `

don't have the issue number or page yet

PR #1077, opened 2020-09-21
[ENH] introduce common schema for shared JSON fields, make schemas stricter

### LINKED ISSUE #1069: ENH ephys schemas --> all "Unit" keys should validate a restricted vocabulary
The meg, eeg, ieeg schemas in https://github.com/bids-standard/bids-validator/tree/master/bids-validator/validators/json/schemas

should get an update such that they only accept a restricted vocabulary as a string for all fields ending on "UNIT".

For example 

https://github.com/bids-standard/bids-validator/blob/4354d0338b6bdae40b7a038f444b5643038f6e15/bids-validator/validators/json/schemas/coordsystem_eeg.json#L7

according to the specification, the only permitted values in this field are "m", "mm", "cm"

ref: https://bids-specification.readthedocs.io/en/latest/04-modality-specific-files/03-electroencephalography.html#coordinate-system-json-_coordsystemjson

### LINKED ISSUE #1075: Enh schemas --> validate array length for fields that want "object of arrays"
By using this part of the schema -> [array length](https://json-schema.org/understanding-json-schema/reference/array.html#length), we can ensure that fields like `FiducialsCoordinates` contain an object of arrays that are each of length 3 (corresponding to a 3D point in space)

there are multiple of such `...Coordinates` fields scattered across several JSON files, but here is the crossref to the EEG part of the spec:

https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/03-electroencephalography.html#coordinate-system-json-_coordsystemjson

rwblair:
fixes #1069
closes #1075 

Should help make json schema validation a little bit stricter. Added a common definitions file so that fields that appear in multiple schemas can be maintained in a single location.


rwblair:
47a7e81 commit message should read fixes #1069

rwblair:
@sappelhoff Can you double check my work hit a bunch of the low hanging schema fixes, #1065, #1066, #1068, #1069, #1070 
Once these get merged I'll cut a new release.

bids-validator/validators/json/schemas/coordsystem_eeg.json
```diff
@@ -3,15 +3,27 @@
...
     "FiducialsDescription": { "type": "string", "minLength": 1 },
+    "FiducialsCoordinates": {
+      "type": "object",
+      "additionalProperties": {
+        "type": "array",
+        "items": {
+          "type": "number"
+        }
+      }
+    },
```
sappelhoff: the expected length of each array passed as part of this object MUST be 3 (x, y, z coordinates) --> Can we validate that as well, please?

bids-validator/tests/json.spec.js
```diff
@@ -144,11 +144,12 @@ describe('JSON', function() {
         NAS: [0, 1, 0],
       },
       AnatomicalLandmarkCoordinateSystem: '...',
-      AnatomicalLandmarkCoordinateUnits: '...',
+      AnatomicalLandmarkCoordinateUnits: 'mm',
       AnatomicalLandmarkCoordinateSystemDescription: '...',
     }
     jsonDict[eeg_coordsystem_file.relativePath] = jsonObj
     validate.JSON(eeg_coordsystem_file, jsonDict, function(issues) {
+      console.log(issues)
```
sappelhoff: is still a remainder from debugging? or do you want to commit this?

bids-validator/validators/json/schemas/coordsystem_eeg.json
```diff
@@ -3,15 +3,27 @@
...
+    "FiducialsCoordinateSystem": { "type": "string", "minLength": 1 },
+    "FiducialsCoordinateSystemDescription": { "type": "string" },
+    "FiducialsCoordinateUnits": { "type": "string", "enum": ["m", "mm", "cm"] },
     "EEGCoordinateSystem": { "type": "string", "minLength": 1 },
-    "EEGCoordinateUnits": { "type": "string", "minLength": 1 },
+    "EEGCoordinateUnits": { "type": "string", "enum": ["m", "mm", "cm"] },
```
sappelhoff: Most of the "*Units" fields across schemas are "required" --> thus I think we should also allow the pattern `^n\/a$` (or are `^` and `$` perhaps not even required? --- we should allow for specifying the string `n/a` with an appropriate pattern)

bids-validator/validators/json/schemas/coordsystem_eeg.json
```diff
@@ -3,15 +3,27 @@
...
+    "FiducialsCoordinates": {
+      "type": "object",
+      "additionalProperties": {
+        "type": "array",
+        "items": {
+          "type": "number"
+        }
+      }
+    },
+    "FiducialsCoordinateSystem": { "type": "string", "minLength": 1 },
```
sappelhoff, line 7: why do some strings have a "minLength" of 1, and others do not have a minLength?

I guess this is to prevent users from filling the field with an empty string like `""` ... but in that case it'd perhaps be good to be consistent? Or was there another reason?

sappelhoff (commented):
nice, thanks a lot @rwblair this looks great - I left a few comments.

bids-validator/tests/json.spec.js
sappelhoff: I see that you already removed it, resolving this issue.

bids-validator/validators/json/schemas/coordsystem_eeg.json
sappelhoff: this would be the case for all `object of arrays` type of fields (typically they end with `*Coordinates`)

bids-validator/validators/json/schemas/common_definitions.json
```diff
@@ -0,0 +1,50 @@
...
+    "DeviceSerialNumber": { "type": "string" },
+    "ECGChannelCount": { "type": "integer", "minimum": 0 },
+    "EEGChannelCount": { "type": "integer", "minimum": 0 },
+    "EMGChannelCount": { "type": "integer", "minimum": 0 },
+    "EOGChannelCount": { "type": "integer", "minimum": 0 },
+    "EpochLength": { "type": "number", "minimum": 0 },
+    "HardwareFilters": {
+      "anyOf": [
+        { "type": "object", "additionalProperties": { "type": "object" } },
+        { "type": "string", "pattern": "^n/a$" }
```
sappelhoff, line 16: do we actually have to escape the `/` in the "pattern" here? Would the current `^n/a$` even match `n/a`?

bids-validator/validators/json/schemas/coordsystem_eeg.json
sappelhoff: note that this fix would close https://github.com/bids-standard/bids-validator/issues/1075

bids-validator/validators/json/schemas/coordsystem_eeg.json
sappelhoff: Update: The `^` and `$` are a good idea, according to https://json-schema.org/understanding-json-schema/reference/string.html?highlight=uri#regular-expressions.

> When defining the regular expressions, it’s important to note that the string is considered valid if the expression matches anywhere within the string. For example, the regular expression "p" will match any string with a p in it, such as "apple" not just a string that is simply "p". Therefore, it is usually less confusing, as a matter of course, to surround the regular expression in ^...$, for example, "^p$", unless there is a good reason not to do so.

bids-validator/validators/json/schemas/eeg.json
```diff
@@ -1,52 +1,37 @@
...
+    "ECGChannelCount": { "$ref": "common_definitions.json#/definitions/ECGChannelCount" },
+    "EMGChannelCount": { "$ref": "common_definitions.json#/definitions/EMGChannelCount" },
+    "MiscChannelCount": { "$ref": "common_definitions.json#/definitions/MiscChannelCount" },
+    "TriggerChannelCount": { "$ref": "common_definitions.json#/definitions/TriggerChannelCount" },
     "EEGPlacementScheme": { "type": "string" },
     "EEGReference": { "type": "string" },
     "EEGGround": { "type": "string" },
     "CapManufacturer": { "type": "string" },
     "CapManufacturersModelName": { "type": "string" },
     "HeadCircumference": { "type": "number" },
```
sappelhoff: we should validate that `HeadCircumference` MUST be a positive number

sappelhoff:
@rwblair can we perhaps also tackle https://github.com/bids-standard/bids-validator/issues/1072 in this PR? I just tried it out, and specifying "URI" as the "format" for a "string" type in the schema seems to work straight forwardly.

For all fields that require a URL or a DOI or a general URI.

If you'd rather have somebody else do it (or in another PR), that'd be completely fine

bids-validator/validators/json/schemas/coordsystem_eeg.json
rwblair: Ended up including "n/a" in the enum listing which is functionally equivalent to the "^n\/a$" regex

bids-validator/validators/json/schemas/common_definitions.json
rwblair: From playing with a toy schema looks like we don't have to escape the '/' in '^n/a$'.

bids-validator/validators/json/schemas/coordsystem_eeg.json
rwblair: Ok I made a common definition called LandmarkCoordinates for all the coordsytem keys ending in 'Coordinates' If you can think of a better name for the  common definition I can change it.

bids-validator/validators/json/schemas/eeg.json
rwblair: Made `HeadCircumfrance` have a `ExclusiveMinimum` of 0.

bids-validator/validators/json/schemas/coordsystem_eeg.json
rwblair: `minLength` of 1 does prevent the empty string as a value. I can't figure a rhyme or reason for some having it and others not. For a lot of them a single character would seem to be about as informative as an empty string so not sure how to resolve this.

rwblair:
@sappelhoff I'll punt #1072 to another PR. I've let this one sit for a bit too long and would like to get a release out.

bids-validator/validators/json/schemas/coordsystem_eeg.json
sappelhoff: yeah, I am not sure either, let's keep it for now :slightly_smiling_face:

sappelhoff (approved):
LGTM, thanks Ross

PR #1079, opened 2020-09-25
Fix cross-talk and fine-calibration extensions

hoechenberger:
We accidentally swapped the Elekta/Neuromag/MEGIN cross-talk and fine-calibration files' extensions, `.fif` and `.dat`, respectively, in #1053.

### DIFF
--- a/bids-validator/tests/type.spec.js
+++ b/bids-validator/tests/type.spec.js
@@ -110,8 +110,8 @@ describe('utils.type.file.isTopLevel', function() {
     '/_task-test_physio.json',
     // cross-talk and fine-calibration files for Neuromag/Elekta/MEGIN data (.fif)
     // must be defined at file level.
-    '/acq-calibration_meg.fif',
-    '/acq-crosstalk_meg.dat',
+    '/acq-calibration_meg.dat',
+    '/acq-crosstalk_meg.fif',
   ]
 
   badFilenames.forEach(function(path) {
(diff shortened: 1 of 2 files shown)


adam2392 (approved):
I don't know meg stuff, but code itself LGTM!

agramfort (approved):

jasmainak (approved):

hoechenberger:
Is it okay to merge this?

bids-validator/tests/type.spec.js
```diff
@@ -135,14 +135,14 @@ describe('utils.type.file.isSubjectLevel', () => {
...
-    '/sub-12/sub-12_acq-calibration_meg.fif',
-    '/sub-12/sub-12_acq-crosstalk_meg.dat',
     '/sub-12/sub-12_acq-calibration_meg.dat',
     '/sub-12/sub-12_acq-crosstalk_meg.fif',
-    '/sub-12/acq-calibration_meg.fif',
-    '/sub-12/acq-crosstalk_meg.dat',
-    '/sub-12/acq-calibration.fif',
-    '/sub-12/acq-crosstalk.dat',
+    '/sub-12/sub-12_acq-calibration_meg.dat',
+    '/sub-12/sub-12_acq-crosstalk_meg.fif',
```
effigies:
```suggestion
```

These two are duplicates of L138-139.

effigies (commented):
Small suggestion.

Also, I couldn't find anything in the spec that makes this clear. If we're validating it, it should probably be in the spec.

hoechenberger:
@effigies 
> Small suggestion.

Thank you, I have applied it.

> Also, I couldn't find anything in the spec that makes this clear. If we're validating it, it should probably be in the spec.

OUCH we had the same mixup there! That must be why it was incorrect in the validator: we exactly matched the spec here, which itself was incorrect. I will file a PR to swap the `.dat` and `.fif` filename extensions in https://bids-specification.readthedocs.io/en/latest/99-appendices/06-meg-file-formats.html#cross-talk-and-fine-calibration-files

For the record, the correct extensions are:

- fine-calibration: **.dat**
- cross-talk: **.fif**

hoechenberger:
@effigies I've opened a PR to correct the specs at https://github.com/bids-standard/bids-specification

sappelhoff (approved):
Thanks for finding the mixup and submitting two fixes @hoechenberger (this one, and https://github.com/bids-standard/bids-specification/pull/621)

Did you try it with your real data again and with this fix everything works? I assume that yes ... and if so, we can merge it.

hoechenberger:
> Did you try it with your real data again and with this fix everything works?

Yes, this is, in fact, how I discovered this mixup :)

Issue #1080, opened 2020-09-27
After a release, bump the version number of the development version


hoechenberger:
Hello, I'm running into an issue with `bids-validator`; specifically, we're running a CI job that once installs the latest stable release from NPM, and in another set of runs installs the code from the `master` branch. Unfortunately, both versions carry the exact same version number (1.5.4), so there is no way for our tests to find out whether they're being run in an environment with a stable or a development version of the validator. Would it be possible to bump the version number after a release? One could use NPM pre-release version identifiers, such `as 1.5.5-alpha.0`. Any thoughts?

sappelhoff:
x-ref #803

hoechenberger:
Now that 1.5.6 has been released, can we bump the version number to 1.5.7-dev.0, similarly to how we handled this in https://github.com/bids-standard/bids-validator/pull/1081?

PR #1093, opened 2020-10-09
Bump version number to 1.5.7-dev.0

hoechenberger:
We should automate this…

x-ref GH-1080

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,7 +1,4 @@
 {
-  "packages": [
-    "bids-validator",
-    "bids-validator-web"
-  ],
-  "version": "1.5.6"
+  "packages": ["bids-validator", "bids-validator-web"],
+  "version": "1.5.7-dev.0"
 }
--- a/bids-validator/package-lock.json
+++ b/bids-validator/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "bids-validator",
-  "version": "1.5.6",
+  "version": "1.5.7-dev.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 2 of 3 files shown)


lerna.json
```diff
@@ -1,7 +1,4 @@
 {
-  "packages": [
-    "bids-validator",
-    "bids-validator-web"
-  ],
-  "version": "1.5.6"
+  "packages": ["bids-validator", "bids-validator-web"],
```
hoechenberger, line 2: formatting change seems to be due to a new pre-commit hook? I didn't add this myself…

sappelhoff:
realizing that 1.5.6 was skipped and looking into the release notes shows that there was some issue with the release process --- perhaps due to our usage of `dev` we should wait with merging this PR until @rwblair has had a look

hoechenberger:
> realizing that 1.5.6 was skipped and looking into the release notes shows that there was some issue with the release process --- perhaps due to our usage of `dev`

I don't think that's related though? 1.5.5 was skipped, and we only added our `-dev.0` change in the 1.5.6 cycle. But yes let's wait for feedback.

sappelhoff:
mmh true... no idea.

> We should automate this…

do you have a proposal?

hoechenberger:
> > We should automate this…
> 
> do you have a proposal?

Nope :)
We could also just add this to the "how to make a release" docs :) It really just takes a minute to do, so… not sure if it's even worth trying to automate it.

rwblair:
My thinking yesterday was that I would bump the version in the PR that would cause the latest commit on master to actually differ from latest tagged release. I'll spend some time today seeing if this can just be incorporated into the lerna publish workflow so that the repo only need to be touched once for two version changes.

rwblair:
Ok added a script here a96790e that should help do this automatically when I publish. Well see how it works soon.

sappelhoff:
nice, let's see how it'll work during next release :)

PR #1095, opened 2020-10-09
Fix recording regex.

rwblair:
https://neurostars.org/t/bids-validator-error-code-90-with-physio-files/3076/7


sappelhoff (approved):
regex changes and tests look good to me

PR #1096, opened 2020-10-12
ENH: schema to conditionally require Description

### LINKED ISSUE #945: Improve validation of coordsystem.json files

sappelhoff:
if *CoordinateSystem is 'Other', ensure that *CoordinateSystemDescription is supplied. closes #945 

This significantly enhances validation and will remind users to supply unusable data.

solved via https://stackoverflow.com/a/64299870/5201771

edit: In a follow-up PR I will add the "enum" to the schema, requiring a list of particular coordinate system to be supplied, where `"Other"` will be the wildcard --> not putting this in here for clarity.

### DIFF
--- a/bids-validator/tests/json.spec.js
+++ b/bids-validator/tests/json.spec.js
@@ -126,6 +126,44 @@ describe('JSON', function() {
     })
   })
 
+  var meg_coordsystem_file = {
+    name: 'sub-01/meg/sub-01_task-testing_coordsystem.json',
+    relativePath: '/sub-01/meg/sub-01_task-testing_coordsystem.json',
+  }
+
+  it('MEG *_coordsystem.json files should have required key/value pairs', function() {
+    var jsonObj = {
+      FiducialsDescription: 'Fiducials were digitized using  ... ',
+      MEGCoordinateSystem: 'CTF',
+      MEGCoordinateUnits: 'mm',
+      MEGCoordinateSystemDescription: 'this is the usual ...',
+      EEGCoordinateSystem: 'Other',
+      EEGCoordinateSystemDescription: 'RAS orientation ...',
+      HeadCoilCoordinateSystem: 'Other',
+      HeadCoilCoordinates: {
+        LPA: [-1, 0, 0],
+        RPA: [1, 0, 0],
+        NAS: [0, 1, 0],
+      },
+      AnatomicalLandmarkCoordinates: {
+        LPA: [-1, 0, 0],
+        RPA: [1, 0, 0],
+        NAS: [0, 1, 0],
+      },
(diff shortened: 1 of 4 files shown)


rwblair:
Awesome work.

PR #1098, opened 2020-10-12
ENH: validate against uri format

### LINKED ISSUE #1072: consider improving all schemas: fields that require a URI can be validated accordingly

sappelhoff:
closes #1072 

@rwblair the new "common schema" you added is really nice!

### DIFF
--- a/bids-validator/tests/json.spec.js
+++ b/bids-validator/tests/json.spec.js
@@ -76,6 +76,23 @@ describe('JSON', function() {
     relativePath: '/sub-01_run-01_eeg.json',
   }
 
+  it('*.json sidecars with CogPOID or CogAtlasID fields should require a uri format', function() {
+    var jsonObj = {
+      TaskName: 'rest',
+      SamplingFrequency: 1000,
+      EEGReference: 'Cz',
+      SoftwareFilters: 'n/a',
+      PowerLineFrequency: 1000,
+      CogAtlasID:
+        'we did a search on https://ww.idontexist.com for the word "atlas"',
+    }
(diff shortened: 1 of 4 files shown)


rwblair:
There is one dataset in the examples that has 'n/a' for these fields:
https://github.com/bids-standard/bids-examples/blob/8031430b2da07b7f2be58f92b0bd11e2f27a5645/ieeg_visual_multimodal/sub-som682/ses-somecog01/ieeg/sub-som682_ses-somecog01_task-spatialpattern_run-02_ieeg.json#L11

Should this be allowed for the cogat fields?

sappelhoff:
> Should this be allowed for the cogat fields?

I think if the data for that field is not available, the dataset curator should just not add the field (it's not required, so that'd be fine). What do you think?

rwblair:
@sappelhoff Works for me.

Issue #1083, opened 2020-09-30
range error on validation


riddlet:
Here's the error message I receive when trying to upload to openneuro:

```
Internal error. SOME VALIDATION STEPS MAY NOT HAVE OCCURRED
REASON:
Maximum call stack size exceeded; please help the BIDS team and community by opening an issue at (https://github.com/bids-standard/bids-validator/issues) with the evidence here.

EVIDENCE:
RangeError: Maximum call stack size exceeded at Zt (https://openneuro.org/0-c51dd9db.bundle.worker.js:69:122016) at https://openneuro.org/0-c51dd9db.bundle.worker.js:69:132485
```

riddlet:
Maybe related to #1063

hjmjohnson:
I too have a similar error: Maximum call stack size exceeded; please help the BIDS team and community by opening an issue at (https://github.com/bids-standard/bids-validator/issues) with the evidence here.

This is for a bids dirctory with 1500+ subjects, and 4000+ scan session.  The scan sessions have globally unique identifiers (to align with historical practices).    We do NOT want to verify that all subjects have the same session names.

rwblair:
@hjmjohnson If its convenient for you to test it here's a branch of the validator with a flag `--ignoreSessionConsistency` that should skip the code that verifies all subjects have the same session names:
https://github.com/bids-standard/bids-validator/tree/skip_session_checks

Also If possible could you re-run the validator passing `--heap-prof` to node. Depending on how you have the validator installed  might be something like 
`node --heap-prof $(which bids-validator) /path/to/dataset`  for a global npm install, or

`node --heap-prof /path/to/bids-validator/bids-validator/bin/bids-validator /path/to/dataset` for an install from source code.

`--heap-prof` fill generate a file in the directory where the command was called from, file name will start with "Heap" and end with ".heapprofile". I can load that in chrome and take a look at whats used memory up until the crash. Not guaranteed to identify the max call stack problem because its concerned with frame size, but it may point to where  the problem is. I haven't found an easy way to just profile the stack count.

@riddlet How many subjects and sessions is your dataset?

hjmjohnson:
@rwblair I will test this out later tonight, or tomorrow morning.  Thank you for your assistance.

hjmjohnson:
THIS WORKED!!! Thanks you.  I am making progress towards validation again. Hans

PR #1099, opened 2020-10-19
Skip session checks

rwblair:
On datasets with large numbers of subjects/sessions performing this check can cause the stack limit to be hit. Adding a flag to skip the checks can allow validation to complete for some datasets: #1083

### DIFF
--- a/bids-validator-web/components/App.jsx
+++ b/bids-validator-web/components/App.jsx
@@ -23,6 +23,7 @@ const initState = () => ({
   options: {
     ignoreWarnings: false,
     ignoreNiftiHeaders: false,
+    ignoreSubjectConsistency: false
   },
 })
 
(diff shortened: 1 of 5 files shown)


rwblair:
@sappelhoff any objection to adding this flag to allow skipping checks that generate these two warnings:
https://github.com/bids-standard/bids-validator/blob/556e86dff6e3b97249f3741bf76a8c5879d667ea/bids-validator/utils/issues/list.js#L207
https://github.com/bids-standard/bids-validator/blob/556e86dff6e3b97249f3741bf76a8c5879d667ea/bids-validator/utils/issues/list.js#L541

sappelhoff (approved):
no objections. it seems to be helpful for people who want to validate a massive amount of data at once 👍 

thanks for preparing it

effigies:
Does this add a checkbox for the online validator?

rwblair:
@effigies good catch, added.

PR #1101, opened 2020-10-30
Update HED validation to support version 2 of hed-validator

happy5214:
Also add new issue codes for new hed-validator issues.

Tests for the new issue codes will be added once version 8.0.0
of the HED schema is released.

### DIFF
--- a/bids-validator/validators/events/hed.js
+++ b/bids-validator/validators/events/hed.js
@@ -129,20 +129,28 @@ export default function checkHedStrings(events, headers, jsonContents, dir) {
     }
 
     // run HED validator
-    return hedValidator.buildSchema(schemaDefinition).then(hedSchema => {
-      for (const [file, hedString] of hedStrings) {
-        const [isHedStringValid, hedIssues] = hedValidator.validateHedString(
-          hedString,
-          hedSchema,
-          true,
-        )
-        if (!isHedStringValid) {
-          const convertedIssues = convertHedIssuesToBidsIssues(hedIssues, file)
-          issues = issues.concat(convertedIssues)
+    return hedValidator.validator
+      .buildSchema(schemaDefinition)
+      .then(hedSchema => {
+        for (const [file, hedString] of hedStrings) {
+          const [
+            isHedStringValid,
+            hedIssues,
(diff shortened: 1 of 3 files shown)


happy5214:
Any update on this?

bids-validator/validators/events/hed.js
```diff
@@ -129,20 +129,28 @@ export default function checkHedStrings(events, headers, jsonContents, dir) {
...
-      for (const [file, hedString] of hedStrings) {
-        const [isHedStringValid, hedIssues] = hedValidator.validateHedString(
-          hedString,
-          hedSchema,
-          true,
-        )
-        if (!isHedStringValid) {
-          const convertedIssues = convertHedIssuesToBidsIssues(hedIssues, file)
-          issues = issues.concat(convertedIssues)
+    return hedValidator.validator
```
effigies, line 132: Looks like the only change apart from formatting is the move from `hedValidator.buildSchema` to `hedValidator.validator.buildSchema`.

effigies (approved):
Looks reasonable to me. Will leave to @rwblair to merge if he has no objections.

VisLab:
Just to confirm the above.  The proposed pull request is a small change to
the bids-validator.  The brunt of the changes to support this are in
the hed-javascript. However, the bids-validator needed the above change in
order for the new version of hed-javascript to work.  Thanks....

rwblair:
@VisLab Do we need to mint a new release of the bids validator or are you able to work from the master branch to get hed-javascript working?

happy5214:
bids-validator depends on hed-javascript, not the other way around, so I don't understand the need for a new release of bids-validator to make hed-javascript work.

happy5214:
To try to re-read the above, the only changes this PR made were due to an API change that necessitated the bump to version 2.0.0. We can't really test the new code's compatibility with bids-validator until a stable version of a HED 3 schema is released. I've opened #1115 as a reminder to fill those in later. The PR CI run showed that the existing tests all pass with hed-validator 2, so it works already as far as we can tell.

Further changes to bids-validator will be posted in the next week or so in another PR to reorganize the HED validation so we can eventually support dataset-level validation.

PR #1110, opened 2020-11-10
add PPG as accepted ch type

sappelhoff:
"PPG" is a new allowed channel type: https://bids-specification.readthedocs.io/en/latest/04-modality-specific-files/03-electroencephalography.html#channels-description-_channelstsv

https://github.com/bids-standard/bids-specification/pull/570


effigies:
Hard to argue with.

PR #1120, opened 2020-11-27
add preregistration to dataset_description schema

Remi-Gau:
Related to its [counterpart PR on specs repo](https://github.com/bids-standard/bids-specification/pull/572)

#### For reviewers

I added this by working from the examples provided by the other fields: let me know if I missed something. :wink: #Newbie

### DIFF
--- a/bids-validator/validators/json/schemas/dataset_description.json
+++ b/bids-validator/validators/json/schemas/dataset_description.json
@@ -52,6 +52,9 @@
     "DatasetDOI": {
       "type": "string"
     },
+    "Preregistration": {
+      "type": "string"
+    },
     "GeneratedBy": {
       "type": "array",
       "minItems": 1,
--- a/bids-validator/validators/json/schemas/dataset_description.json
+++ b/bids-validator/validators/json/schemas/dataset_description.json
@@ -53,7 +53,8 @@
       "type": "string"
     },
     "Preregistration": {
-      "type": "string"
+      "type": "string",
+      "format": "uri"
     },
     "GeneratedBy": {
       "type": "array",


bids-validator/validators/json/schemas/dataset_description.json
```diff
@@ -52,6 +52,9 @@
     "DatasetDOI": {
       "type": "string"
     },
+    "Preregistration": {
+      "type": "string"
```
sappelhoff: All valid strings to this field should be URIs, right?

```suggestion
      "type": "string",
      "format": "uri"
```

Remi-Gau: Yes ! 

But that's where I got confused because it is also the case for `DatasetDOI` is supposed to be an URI and it is just a `string` in that JSON and also in the data type column in the table in specs.

So I was trying to have things be internally consistent, because I will admit that I have no idea how the validator works and does URI validation. :see_no_evil:

sappelhoff: Yes, your confusion is perfectly reasonable. The validator schema has grown over time, becoming more accurate - but some things were not improved along the way and are still like they used to be. We use https://json-schema.org/, and that's a little language in itself ... so a particular pair of contributor and reviewer may not have known that a field can get a much tighter validation than simply "string".

--> It's definitely worth it to question existing schemas and improve them if possible (improve = make them "tighter")

Remi-Gau: Apprently `common_definitions.json` has this:

```
"CogPOID": { "type": "string", "format": "uri" },
```

So I would assume this is it should be done in this case. Though I am now thinking whether I should update the `DatasetDOI` too.

I think I need a crash course on the validator.

Remi-Gau: Should I add "cross check that things that are URI are encoded as such in the validator schema" on my to do list?

sappelhoff: > Should I add "cross check that things that are URI are encoded as such in the validator schema" on my to do list?


@Remi-Gau that'd be very helpful 👍

PR #1122, opened 2020-12-02
remove isCorrectModality function and its call in quicktest

rwblair:
Quicktest looks to see if there was at least one file that either:
1. had a .nii extension and was in a either a func, anat, or dwi directory
2. had a json or tsv extension and was in any other accepted modality directory.

This is causing more problems that it solves:
https://neurostars.org/t/quick-validation-failed-on-physiological-files/17611

We now have regex to let users know what is a correctly or incorrectly formed filename. This PR makes quicktest just test that at least one subject session directory exists.

### DIFF
--- a/bids-validator/utils/modalities.js
+++ b/bids-validator/utils/modalities.js
@@ -39,37 +39,4 @@ export default {
 
     return modalities
   },
-
-  isCorrectModality: path => {
-    let isCorrectModality = false
-    // MRI
-    if (
-      path[0].includes('.nii') &&
-      ['anat', 'func', 'dwi'].indexOf(path[1]) != -1
-    ) {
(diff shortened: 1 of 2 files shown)


sappelhoff (approved):

effigies:
Sorry, I maybe should have blocked this. I thought we'd decided it was appropriate to require at least one NIfTI file in `func/`?

rwblair:
We did, merging it was premature. I woke up this morning and the isCorrectModality code rubbed me the wrong way so I hit merge. It should be replaced at some point, guess it would be good to make a json/yaml way of denoting required files.

I'll revert and expand the quick validation error message to make it more transparent what exactly its looking for.

PR #1125, opened 2020-12-08
[ENH] Add part entity to func and dwi data

### LINKED ISSUE #1117: Add support for part entity
Now that https://github.com/bids-standard/bids-specification/pull/424 has been merged, the validator needs to support the `part` entity. I can add the regular expressions in a PR, but there are a few things I may need some help on.

1. Files with `part-phase` should have `Units` specified in their sidecars. Can the validator check this?
1. The `_phase` suffix is now deprecated. It would be great if the validator could raise a warning when that suffix is detected.
1. ~~I placed `part` _after_ `recording` in the entity order. However, I just checked the validator regular expressions before opening this issue, and they seem to take advantage of the fact that `recording` is located after all of the other entities. Should I open a fix PR to the specification to change the order?~~ I opened https://github.com/bids-standard/bids-specification/pull/692, so there's no need to change the validator's regular expressions beyond adding the new entity.

tsalo:
Closes #1117.

The part entity allows values of `imag` (imaginary), `real`, `mag` (magnitude), and `phase`. This PR adds the part entity to the `dwi`, `func_bold`, and `anat` filename pattern collections. I don't know if the validator is able to check that `phase` data have a `Units` field in the associated json, or to raise a deprecation warning if the phase suffix is used instead of the bold suffix with a `part-phase` entity, but it would be nice to support both of these features.

Changes proposed:

- Add `part` entity to func and dwi data.
- Remove `defacemask` suffix for func data.
- Remove `echo` entity from func data with `stim` or `physio` suffixes.
- Remove `echo` entity from func data with `events` suffix.


rwblair:
Back in august I had tried to write the regex for bep001 in #1035 I created two separate sets of regex to differentiate between suffixes which allow or don't allow part:
https://github.com/bids-standard/bids-validator/blob/54ecabe90d31187fe149806d513f1c2e527a5402/bids-validator/bids_validator/rules/anatomical_file_level_rules.json#L88 

Is any of that worth saving or should I close that PR?

tsalo:
> Is any of that worth saving or should I close that PR?

That's probably better than what I have, although it would be nice to limit the values to `real`, `imag`, `mag`, and `phase`. Also `part` needs to be supported for `dwi` and `bold` too.

Is that PR mergeable at this point? If so, I can just close this one.

rwblair:
My PR probably covers parts of the bep that aren't merged yet and needs test. Ill copy over the appropriate regex to this PR to keep it limited in scope.

tsalo:
That would be awesome. Thanks!

tsalo:
@rwblair I think I could add the other entities and new qMRI suffixes to this PR, if you haven't already started on that.

EDIT: I've opened #1161.

sappelhoff (commented):
could you please add a couple of test cases for this?
I think this would be the right place:

https://github.com/bids-standard/bids-validator/blob/227b1f1fe4cd8f6976d650c0eb9dfef0dbe594de/bids-validator/tests/type.spec.js#L188-L200

1. Please have a look at how tests are done in that file
2. Then identify *where* it would make sense to test an additional filename with a `part` entity
3. Add these filenames (or edit existing ones, or both) and check that the tests pass

This can help verify your own changes and protect us against regression in future PRs

tsalo:
Thanks @sappelhoff. I've expanded the tested filenames and they're passing.

sappelhoff (commented):
cool thanks, could you just tell me why `part` is only added for func and dwi? From the [spec](https://bids-specification.readthedocs.io/en/latest/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#anatomy-imaging-data), it looks to me like one could also use it for anat?

also, why did you pull out "phase" regexps out of func and put them into a new "func_phase"?

tsalo:
>cool thanks, could you just tell me why part is only added for func and dwi? From the spec, it looks to me like one could also use it for anat?

The main reason is that getting it set up in `anat` would basically mean implementing all of qMRI. I didn't want things to sprawl too much in one PR, so I just handled that in #1161.

>also, why did you pull out "phase" regexps out of func and put them into a new "func_phase"?

The phase suffix doesn't allow the `part` entity, because it's equivalent to `part-phase_bold`. I couldn't get tests to pass without splitting it out. In the short-term, having a named group works best. Medium-term, I'm thinking that it would be good to model the filename patterns directly after the groups in the YAML files, and long-term, we'd of course want to render the patterns automatically from the YAML files (as @rwblair is working on in #1126).

sappelhoff (approved):
> I didn't want things to sprawl too much in one PR, so I just handled that in #1161.

okay

> The phase suffix doesn't allow the part entity, because it's equivalent to part-phase_bold. I couldn't get tests to pass without splitting it out. 

:+1: makes sense

---

Looks good from my side, but an MRI person should check and merge (not me)

bids-validator/bids_validator/rules/file_level_rules.json
```diff
@@ -128,7 +128,7 @@
   },
 
   "func": {
-    "regexp": "^[\\/\\\\](sub-[a-zA-Z0-9]+)[\\/\\\\](?:(ses-[a-zA-Z0-9]+)[\\/\\\\])?func[\\/\\\\]\\1(_\\2)?_task-[a-zA-Z0-9]+(?:_acq-[a-zA-Z0-9]+)?(?:_ce-[a-zA-Z0-9]+)?(?:_dir-[a-zA-Z0-9]+)?(?:_rec-[a-zA-Z0-9]+)?(?:_run-[0-9]+)?(?:_echo-[0-9]+)?((?:@@@_func_ext_@@@)|(?:_recording-[a-zA-Z0-9]+)?(?:@@@_cont_ext_@@@))$",
+    "regexp": "^[\\/\\\\](sub-[a-zA-Z0-9]+)[\\/\\\\](?:(ses-[a-zA-Z0-9]+)[\\/\\\\])?func[\\/\\\\]\\1(_\\2)?_task-[a-zA-Z0-9]+(?:_acq-[a-zA-Z0-9]+)?(?:_ce-[a-zA-Z0-9]+)?(?:_dir-[a-zA-Z0-9]+)?(?:_rec-[a-zA-Z0-9]+)?(?:_run-[0-9]+)?(?:_echo-[0-9]+)?(?:_part-(imag|mag|phase|real))?((?:@@@_func_ext_@@@)|(?:_recording-[a-zA-Z0-9]+)?(?:@@@_cont_ext_@@@))$",
```
effigies: This is permitting `events` and `defacemask`, `physio`, `stim` to have `part-` now. Is that intended?

tsalo: It is not intended, but I didn't want to start breaking down some of these regexes too much. I can fix them if that's cool, though?

- `defacemask` isn't valid for `func/`
- `echo` isn't valid for `physio` or `stim`.
- `echo` also probably _shouldn't_ be valid for `events`, but we have it as optional in the schema right now. We should fix that.

Should I fix these?

To clarify- in order to fix `part` I'll need to split the regexes up to be like the schema groups, so I can fix everything in one go.

effigies: That sounds good to me.

tsalo: Done!

bids-validator/bids_validator/rules/file_level_rules.json
```diff
@@ -137,17 +137,37 @@
         "_cbv\\.nii\\.gz",
         "_cbv\\.nii",
         "_cbv\\.json",
```
effigies, line 139: CBV can have complex components? @TheChymera probably knows best here.

bids-validator/bids_validator/rules/file_level_rules.json
```diff
@@ -127,8 +127,8 @@
     }
   },
 
-  "func": {
-    "regexp": "^[\\/\\\\](sub-[a-zA-Z0-9]+)[\\/\\\\](?:(ses-[a-zA-Z0-9]+)[\\/\\\\])?func[\\/\\\\]\\1(_\\2)?_task-[a-zA-Z0-9]+(?:_acq-[a-zA-Z0-9]+)?(?:_ce-[a-zA-Z0-9]+)?(?:_dir-[a-zA-Z0-9]+)?(?:_rec-[a-zA-Z0-9]+)?(?:_run-[0-9]+)?(?:_echo-[0-9]+)?((?:@@@_func_ext_@@@)|(?:_recording-[a-zA-Z0-9]+)?(?:@@@_cont_ext_@@@))$",
+  "func_group1": {
```
effigies: On the whole, I would again prefer explicit names. This one could just be
```suggestion
  "func": {
```

bids-validator/bids_validator/rules/file_level_rules.json
```diff
@@ -137,17 +137,37 @@
...
-        "_phase\\.json",
         "_sbref\\.nii\\.gz",
         "_sbref\\.nii",
-        "_sbref\\.json",
+        "_sbref\\.json"
+      ]
+    }
+  },
+
+  "func_group2": {
```
effigies:
```suggestion
  "func_phase_deprecated": {
```

bids-validator/bids_validator/rules/file_level_rules.json
```diff
@@ -137,17 +137,37 @@
...
+    "tokens": {
+      "@@@_func_ext_@@@": [
+        "_phase\\.nii\\.gz",
+        "_phase\\.nii",
+        "_phase\\.json"
+      ]
+    }
+  },
+
+  "func_group3": {
```
effigies:
```suggestion
  "func_events": {
```

bids-validator/bids_validator/rules/file_level_rules.json
```diff
@@ -137,17 +137,37 @@
...
-        "_events\\.json",
-        "_defacemask\\.nii\\.gz",
-        "_defacemask\\.nii"
-      ],
+        "_events\\.json"
+      ]
+    }
+  },
+
+  "func_group4": {
```
effigies:
```suggestion
  "func_timeseries": {
```

effigies (approved):
Looks good. Small suggestions and a clarification. Nothing blocking, though.

bids-validator/bids_validator/rules/file_level_rules.json
tsalo: I'll use these and update the schema as well.

bids-validator/bids_validator/rules/file_level_rules.json
tsalo: Good point. It's definitely allowed in the specification, but maybe it shouldn't be?

effigies: Well, if it's in the spec, then we should support it until the spec is clarified.

tsalo: I can open an issue, which will hopefully get more eyes on it than this PR.

tsalo: Opened bids-standard/bids-specification#726.

tsalo:
Should I merge?

bids-validator/bids_validator/rules/file_level_rules.json
TheChymera: Perhaps I misunderstand what you mean by complex components? You mean represented by 2-D vector values in each cell?

effigies:
@thechymera For a number of sequences, you can get pairs of images for phase and magnitude. Typical BOLD is just magnitude, but you can get the phase images as well. Does this ever happen for CBV?

TheChymera:
@effigies in the end it's always just one image — i.e. CBV is always just magnitude. How you get there can greatly vary, though. For contrast-agent-based CBV it's just one image from the start, like BOLD. For ASL CBV (of which there are numerous flavours) the raw data consists of the pre-pulse and post-pulse acquisition, so something like _cbv0 and _cbv1, though in truth the contrasts are _bold (pre-pulse) and _boldPLUScbv (post-pulse), from which you then separate the _cbv.

To keep things simple for the time being, I would suggest leaving the specification analogous to BOLD, and updating it once we have an ASL workflow. Incidentally, I'm looking at some ASL data later this year, so I might get back to you on that :)

Issue #1034, opened 2020-08-03
Reorganize file_level_rules.json


rwblair:
Current implementation has lots of repeated regex. I was thinking along the lines of putting all the entities and their regex in an object. Then when defining valid file names we can just have a list of entities key names (and if they are required or optional) instead of the regex itself.

I think we could also provide better feed back on filename issues if we parsed the file extension first and selectively chose which rules to apply, and if it fails return information on whats required and optional. Doesn't look like there's a way to get the regex show you points of failure. One option would be to apply entity regex consecutively against a file name, and consume the entities as they match. This way we could find specifically which ones fail.

rwblair:
We should use these schema files to enforce required entities: bids-standard/bids-specification#475

Issue #974, opened 2020-06-03
Use bids-specification files for filename patterns


tsalo:
This may be premature, but in https://github.com/bids-standard/bids-specification/pull/475 we are attempting to build a machine-readable version of the specification that can be versioned and pulled in for other packages, like `pybids` and `bids-validator`. While the PR is in early stages, I wanted to make sure that the `bids-validator` devs were aware of the effort, so they could provide input if they wanted. 

The goal would be to replace, at minimum, the validator's [filename rules](https://github.com/bids-standard/bids-validator/tree/8891c2978c41aeac28d89324baadc9b5be76365b/bids-validator/bids_validator/rules) with parsers that would rely on config files from the official specification.

yarikoptic:
Might be worth creating a little tiny validator right here to start with may be?

yarikoptic:
I believe that is WiP within https://github.com/bids-standard/bids-validator/tree/use_schema branch, which I do not think yet was proposed as a PR which then could close this issue.

yarikoptic:
actually I was wrong -- that branch is some old initial attempt or smth.  @rwblair @nellh -- where is current work done on RFing bids-validator to use BIDS schema?

effigies:
https://github.com/bids-standard/bids-validator/blob/master/bids-validator/src/README.md

PR #1126, opened 2020-12-08
[WIP] Use specificaiton schemas to generate regex.

rwblair:
As PRs like bids-standard/bids-specification#424 become more common, and require the regex in the validator to be updated, and as the regex gets more and more complicated it might be good to move towards automatically generated regex.

This adds a directory called schemas to the validator that has a complete set of the specifications schema files copied with cp from the current master specification. The index.js file in that directory will attempt to generate appropriate regex for valid filenames from the yaml files.

Thoughts and todos:
- [ ] Still need to have validator use this regex, only played with it by hand so far.
- [ ] There should be a npm command or bash script to handle generating both python and js regex and placing the generated files appropriately. 
- [ ] Find way for schema files to be tracked/copied when the specification changes, hopefully better than me calling cp.
- [ ] See if the regex for subject and session level files can be reflected upstream in the specification's schema if not already.
- [ ] Continuous and physio files can sneak into a lot of modalities, their regex are not currently able to be generated from the schemas.


sappelhoff:
related:
- https://github.com/bids-standard/bids-validator/issues/1034
- https://github.com/bids-standard/bids-validator/issues/974

Issue #1121, opened 2020-11-27
cross check that things that are URI are encoded as such in the validator schema


Remi-Gau:
Should I add "cross check that things that are URI are encoded as such in the validator schema" on my to do list?

_Originally posted by @Remi-Gau in https://github.com/bids-standard/bids-validator/pull/1120#discussion_r531684531_

sappelhoff:
We closed this in #1134. thanks again!

PR #1134, opened 2020-12-28
[FIX] Validate TermURL is formatted as a URI

Remi-Gau:
related to #1121 

A couple of instances of DOI and URL are just of the `type string` but should also include `format URI` according to the specs.

One thing I am not clear about the validator yet is how it tells apart things that `required` from the `recommended` ones: in more concrete fashion "are the changes that I am introducing here going to start breaking people's dataset or just throw warnings?"

### DIFF
--- a/bids-validator/validators/json/schemas/data_dictionary.json
+++ b/bids-validator/validators/json/schemas/data_dictionary.json
@@ -34,7 +34,8 @@
         "TermURL": {
           "title": "TermURL",
           "description": "URL pointing to a formal definition in an ontology available on the web.",
-          "type": "string"
+          "type": "string",
+          "format": "uri"
         }
       }
     }
--- a/bids-validator/validators/json/schemas/dataset_description.json
+++ b/bids-validator/validators/json/schemas/dataset_description.json
@@ -50,8 +50,7 @@
       "type": "array"
     },
     "DatasetDOI": {
-      "type": "string",
-      "format": "uri"
+      "type": "string"
     },
     "Preregistration": {
       "type": "string",
@@ -84,7 +83,7 @@
         "type": "object",
         "properties": {
           "URL": {"type": "string", "format": "uri"},
-          "DOI": {"type": "string", "format": "uri"},
+          "DOI": {"type": "string"},
           "Version": {"type": "string"}
         }
       }
--- a/bids-validator/validators/json/schemas/dataset_description.json
+++ b/bids-validator/validators/json/schemas/dataset_description.json
@@ -50,7 +50,8 @@
       "type": "array"
     },
     "DatasetDOI": {
-      "type": "string"
+      "type": "string",
+      "format": "uri"
     },
     "Preregistration": {
       "type": "string",
@@ -83,7 +84,7 @@
         "type": "object",
         "properties": {
           "URL": {"type": "string", "format": "uri"},
-          "DOI": {"type": "string"},
+          "DOI": {"type": "string", "format": "uri"},
           "Version": {"type": "string"}
         }
       }


effigies:
URLs must be URIs, but when we've used DOI, it generally has not meant URI in the past, and making it do so will invalidate many datasets.

Remi-Gau:
> URLs must be URIs, but when we've used DOI, it generally has not meant URI in the past, and making it do so will invalidate many datasets.

yeah that's what I thought too, so I suspect that Recommended things don't go in the schema.

I will remove the DOI part from this PR then.

Remi-Gau:
> URLs must be URIs

stupid brain is asking a question here: does that mean that enforcing a URI format for things that listed as URL is an overkill and therefore not necessary? If so we can just close this PR all together because that would be the only thing left.

sappelhoff:
> stupid brain is asking a question here: does that mean that enforcing a URI format for things that listed as URL is an overkill and therefore not necessary? 

i am not sure that I follow the question completely, but IMHO it definitely makes sense to enforce URI format for a field that must be a URL.

Sure, that way users could still put a DOI in that field and pass validation while actually supplying something wrong ... 

but at least they cannot put a value like `"1"` into the field anymore :-) (which would be permissible if the only restriction is `string`)

so it's definitely an improvement

Remi-Gau:
I am actually in favor of enforcing this: just not sure how many datasets are gonna break by enforcing the new DOI format.

sappelhoff:
> I am actually in favor of enforcing this: just not sure how many datasets are gonna break by enforcing the new DOI format.

ahh sorry for misunderstanding. --> I think not many datasets will be broken with your currently proposed changes.

all those that specified a valid URL will still pass (even if we actually want something like `doi:10237/2312`)

The only ones that will fail are those that specified something like:

- `see this url: my-website.com`
- `available from: my-perfect-site.net`
- etc

And I think "breaking" those is fine IMHO. It's the old song: :musical_note:  "we are not making a backward incompatible change, we are just extending coverage, thus revealing that you have been wrong all along"

:grimacing: it's terrible actually, and it's annoying. But it's also a way to move forward I think

sappelhoff (approved):

Remi-Gau:
> And I think "breaking" those is fine IMHO. It's the old song: musical_note "we are not making a backward incompatible change, we are just extending coverage, thus revealing that you have been wrong all along"
> 
> grimacing it's terrible actually, and it's annoying. But it's also a way to move forward I think

gotcha! :-)

@effigies what's your take on this?

bids-validator/validators/json/schemas/data_dictionary.json
```diff
@@ -34,9 +34,10 @@
...
           "title": "TermURL",
           "description": "URL pointing to a formal definition in an ontology available on the web.",
-          "type": "string"
+          "type": "string",
+          "format": "uri"
         }
       }
     }
   }
-}
```
sappelhoff, line 42: @Remi-Gau not sure if you already know, but to avoid this annoying "diff" that git shows up at the end of a file, you can edit a setting in your text-editor to always end a file with a blank new line.

At least you can do it in most of the commonly used text-editors.

See also:
- https://stackoverflow.com/questions/2287967/why-is-it-recommended-to-have-empty-line-in-the-end-of-a-source-file
- https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file

that way, the diff will only show up if you edit a file that somebody has edited on windows without being aware (or caring) that unix files should end with a blank new line :-)

Remi-Gau: Did not know that one. Thanks.

I tend to have `.gitattributes` with the following 
```
# Auto detect text files and perform LF normalization
* text=auto
```

To avoid some of the headaches introduced that by how Windows handles end of lines.

Will see how to set that up on VS code.

Do you want me to remove it?

sappelhoff: Does the `.gitattributes` solution work in a way that adding this file to the repo will solve the problem for everyone, irrespective of their text-editor settings?

What would it actually do if I close a file without a blank newline at the end? --> would git then edit my file (like the text-editor would do if you have the setting)?

> Do you want me to remove it?

it's fine I think :-) I just wanted to mention it in case you didn't know, but now it turns out I may learn something myself.

PS: I just rebased your branch on master and force-pushed, so in case you want to edit things, do `git pull` before please

Remi-Gau: For any kind of end of line issues, then yes this would solve the problem for everyone irrespective of editor. 

Otherwise, this will depend on the git config of each user (and OS)

https://stackoverflow.com/questions/21472971/what-is-the-purpose-of-text-auto-in-gitattributes-file

It drove me nuts when I started working on windows + linux + matlab on the same code base.

Now it is one of my defaults in most of my repo: though `* text=auto` is definitely a bit of a nuclear option according to the 1% of the world population who can actually parse and understand the official Git documentation.

sappelhoff: cool thanks! I would be +1 then to have this for bids-specification, bids-validator, and bids-starter-kit. WDYT?

Remi-Gau: I am OK with that.

Out of curiosity I will boot my windows machine to see if this could be an issue (will have to check my user config that I supsect I changed to avoid those issues)

effigies:
> Bare DOIs such as 10.18112/openneuro.ds000001.v1.0.0 are DEPRECATED.

Bare (non-URI) DOIs should be a warning, not an error. Enforcing them would violate

> DEPRECATED - A "deprecated" entity or metadata field SHOULD NOT be used in the generation of new datasets. It remains in the standard in order to preserve the interpretability of existing datasets. Validating software SHOULD warn when deprecated practices are detected and provide a suggestion for updating the dataset to preserve the curator's intent.

I am +1 for enforcing URLs as URIs, but -1 for enforcing existing DOI fields be promoted to URIs. If a new DOI field is proposed, we could enforce it, but at least my impression was that new fields would most often be specified as URIs rather than imposing a specific type of URI.

sappelhoff:
> Bare (non-URI) DOIs should be a warning, not an error.

argh, yes I forgot that case, thinking that people would either use a `doi:` or `http[s]:` scheme.

I agree with Chris.

Remi-Gau:
> Bare (non-URI) DOIs should be a warning,

I need to check when warnings are triggered then. :-)

sappelhoff:
> I need to check when warnings are triggered then. :-)

AFAIK This is non-trivial with JSON-schema and Ross was looking into that. I think for the purpose of this PR it'd suffice to only remove the strict requirements for DOIs that are currently still there.

Remi-Gau:
> > I need to check when warnings are triggered then. :-)
> 
> AFAIK This is non-trivial with JSON-schema and Ross was looking into that. I think for the purpose of this PR it'd suffice to only remove the strict requirements for DOIs that are currently still there.

OK I will nuke the DOI related stuff then. This will be the smallest contribution "EvEr"!

sappelhoff (approved):

bids-validator/validators/json/schemas/data_dictionary.json
Remi-Gau: Could not reproduce the "problem" just now but I dug out the PR In the starter kit where I was faced with this problem and Kirstie's pointer to the explanation: https://github.com/bids-standard/bids-starter-kit/pull/120#issuecomment-478514888

Apparently, stack overflow seems to agree that a good solution is to have a repo-based solution for this would be to have a `.gitattributes` file

Not sure how urgent that is. I am ok adding it but I am also OK to wait until the issue arises "again" (though having the file could help some git beginners avoid the problem in first place).

sappelhoff: > Not sure how urgent that is. I am ok adding it but I am also OK to wait until the issue arises "again" (though having the file could help some git beginners avoid the problem in first place).

yes, I am also not sure. I think I am also fine posting the explanation every once in a while that contributors can change their editor settings.

If we are both usure to add .gitattributes or not, let's keep it in mind for now and not do it.

effigies (approved):

sappelhoff:
Thanks Remi

PR #1142, opened 2021-01-11
allow json sidecars for tsv in top and session for MEG EEG iEEG channels and EEG iEEG electrodes

### LINKED ISSUE #1118: Allow channels.json and electrodes.json also in top and session level
channels.tsv and electrodes.tsv are allowed at the top level:

https://github.com/bids-standard/bids-validator/blob/556e86dff6e3b97249f3741bf76a8c5879d667ea/bids-validator/bids_validator/rules/top_level_rules.json#L68-L102

if arbitrary columns are defined in these files, they'd need to be specified in an accompanying JSON file.

Similar to how it's possible already at the file level:

https://github.com/bids-standard/bids-validator/blob/556e86dff6e3b97249f3741bf76a8c5879d667ea/bids-validator/bids_validator/rules/file_level_rules.json#L170-L173

We should extend the top level regexp to include channels.JSON and electrodes-JSON.

sappelhoff:
closes #1118 

If we allow the TSV file, we should allow an accompanying JSON (for electrodes and channels).

related: https://github.com/bids-standard/bids-starter-kit/pull/140

### DIFF
--- a/bids-validator/bids_validator/rules/session_level_rules.json
+++ b/bids-validator/bids_validator/rules/session_level_rules.json
@@ -54,6 +54,7 @@
       "@@@_meg_ses_type_@@@": [
         "_events.tsv",
         "_channels.tsv",
+        "_channels.json",
         "_meg.json",
         "_coordsystem.json",
         "_photo.jpg",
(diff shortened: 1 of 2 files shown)


effigies (approved):
LGTM.

PR #1143, opened 2021-01-12
Add --filenames option to validate filenames from stdin

nellh:
This adds support for passing filenames from stdin and running validation against the list, which is useful for OpenNeuro to perform some validation before a dataset is pushed. At the time this needs to run, there are only filenames available for new git tree (and file sizes for new files but not all files). Test 44 is disabled and a limitation is .bidsignore is not handled by this mode since file contents are not available when it runs.

I've hidden this option from the CLI help as it isn't that generally useful.

### DIFF
--- a/bids-validator/bin/bids-validator
+++ b/bids-validator/bin/bids-validator
@@ -2,7 +2,7 @@
 
 process.title = 'bids-validator'
 
-var argv = require('yargs')
+const argv = require('yargs')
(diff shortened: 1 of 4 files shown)


david-nishi (approved):
## Tested
- test suite passing
- confirmed from cli with single and multiple files

ckrountree:
### SPRINT REVIEW
bids 1143

PR #1146, opened 2021-01-14
pixels is a valid unit for iEEG coordsystem

sappelhoff:
see table in https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/04-intracranial-electroencephalography.html#coordinate-system-json-_coordsystemjson

I wasn't sure how to make this work with the "common definitions":

https://github.com/bids-standard/bids-validator/blob/a110f684ac96921d7ac582c1ce81eda3ecb78925/bids-validator/validators/json/schemas/common_definitions.json#L6

... that is - whether it's possible to say something like `accept all keywords you find in $common-definitions, and additionally this one: pixels`.

So I just added the keywords on their own, which is less elegant, but achieves what we need.

needed for https://github.com/bids-standard/bids-examples/pull/221 to pass.

### DIFF
--- a/bids-validator/tests/json.spec.js
+++ b/bids-validator/tests/json.spec.js
@@ -291,7 +291,7 @@ describe('JSON', function() {
   it('iEEG *_coordsystem.json schema should require *Description if *Coordsystem is "Other"', function() {
     var jsonObj = {
       iEEGCoordinateSystem: 'Other',
-      iEEGCoordinateUnits: 'mm',
+      iEEGCoordinateUnits: 'pixels',
     }
     jsonDict[ieeg_coordsystem_file.relativePath] = jsonObj
     validate.JSON(ieeg_coordsystem_file, jsonDict, function(issues) {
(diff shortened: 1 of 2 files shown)


sappelhoff:
test failure seems unrelated:

https://app.circleci.com/pipelines/github/bids-standard/bids-validator/500/workflows/29e7a14e-a7d0-4800-9753-6554f5680abd/jobs/6844

image

sappelhoff:
actually, perhaps this works: https://stackoverflow.com/q/29841175/5201771

EDIT: I played around a bit and tried to make it work (also using https://json-schema.org/understanding-json-schema/structuring.html#extending) but couldn't do it.

I think this is good to merge unless somebody wants to work some magic.

nellh:
> test failure seems unrelated:
> 
> https://app.circleci.com/pipelines/github/bids-standard/bids-validator/500/workflows/29e7a14e-a7d0-4800-9753-6554f5680abd/jobs/6844
> 
> image

This is from #1143, I'll make a PR to fix the linting issue.

PR #1147, opened 2021-01-14
Resolve linting issues from #1143

nellh:
Removes unused import and ignores the expected console logs.


sappelhoff (approved):
According to my tests this fixes the problem, thanks @nellh

PR #1158, opened 2021-01-29
Report warning instead of error if only warnings are returned by hed-validator

happy5214:
The existing implementation errors out even if only warnings are issued
by hed-validator. This is not desirable, and has been fixed here. A
new issue code has been added, and a new test case has also been added,
though it will not be correct until another change is made upstream.

This is of relatively high severity, as it prevents OpenNeuro from properly validating
any BIDS datasets with HED tag extensions in the sidecar, which generate warnings.


rwblair:
@happy5214 can you double check my work on the HED issue numbers, due to recent merge I moved them down, wanted to make sure the issue codes in the tests match the correct error messages. Might of been hasty with the merge.

happy5214:
> @happy5214 can you double check my work on the HED issue numbers, due to recent merge I moved them down, wanted to make sure the issue codes in the tests match the correct error messages. Might of been hasty with the merge.

@rwblair Sorry for the late reply. Issue code 209 and 210 were indeed flipped (the error was _before_ the warning in the original, and the issue codes in the actual code reflect that), so that needs to be fixed.

PR #1179, opened 2021-02-05
Correct issue code problems from #1158

happy5214:
The sidecar warning and error codes were inadvertently flipped in the process of writing #1158. This rectifies this error and also renames the error issue for consistency.

### DIFF
--- a/bids-validator/utils/issues/list.js
+++ b/bids-validator/utils/issues/list.js
@@ -1064,13 +1064,13 @@ export default {
     reason: 'A HED tag extension was found.',
   },
   209: {
+    key: 'HED_INVALID_SIDECAR_ERROR',
+    severity: 'error',
+    reason: 'A HED error was found in the JSON sidecar.',
+  },
+  210: {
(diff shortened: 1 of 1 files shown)


happy5214:
The issue created by the flipped issue code levels results in warnings in sidecars causing failed validation in OpenNeuro, so this is an urgent fix.

effigies:
Is it plausible to create a unit test that will verify that errors raise one code and warnings the other?

happy5214:
> Is it plausible to create a unit test that will verify that errors raise one code and warnings the other?

In a way they already do. See https://github.com/bids-standard/bids-validator/blob/3c42673fded74ae5c7bb94169169eccd0c2a3c4e/bids-validator/tests/events.spec.js#L1253 and https://github.com/bids-standard/bids-validator/blob/3c42673fded74ae5c7bb94169169eccd0c2a3c4e/bids-validator/tests/events.spec.js#L1343 for examples of existing test assertions for the sidecar "meta-issues" (for lack of a better term).

PR #1181, opened 2021-02-09
Refactor filename only mode to avoid fullTest and read file calls

nellh:
This is a fix for OpenNeuro to succeed on when receiving pushes for more existing datasets by limiting validation to only filenames. This disables a few rules that worked in full test but attempting to use the entire full test pass as is was rejecting many valid datasets.


david-nishi (approved):
## Tested:
 - test suite passing
 - ran validator on 2 datasets

PR #1183, opened 2021-02-09
[FIX] Add qMRI fmap filename rules and reorganize fmap patterns

tsalo:
Stems from https://github.com/bids-standard/bids-examples/pull/245#issuecomment-776167347.

Changes proposed:
- Add qMRI field map patterns (TB1EPI and TB1DAM).
- Reorganize field map patterns based on schema. This will make the `dir` entity required for `_epi` and `_m0scan` field maps.

PR #1184, opened 2021-02-09
FIX: Add echo entity to MPM/MTS

effigies:
By my reading of [anat.yaml](https://github.com/bids-standard/bids-specification/blob/7cdeadf6d0f448cbf5fe0d30d75787fb3166bbee/src/schema/datatypes/anat.yaml#L140-L157), this needs fixing. Also leading to an error in https://github.com/effigies/BIDS-examples/tree/qmri/qmri_mpm.

### DIFF
--- a/bids-validator/bids_validator/rules/file_level_rules.json
+++ b/bids-validator/bids_validator/rules/file_level_rules.json
@@ -138,7 +138,7 @@
   },
 
   "anat_vfa_mt": {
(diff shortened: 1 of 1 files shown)


tsalo (approved):
LGTM! Not sure why the test is failing though.

effigies:
Yeah, I restarted it and it still fails. @rwblair any ideas? It's possible that some dependency just made a release and broke everything. Will know that when you update #1183.

effigies:
Just searching, we might need to look into something like https://github.com/facebook/jest/issues/8384#issuecomment-674351539 or https://github.com/facebook/jest/issues/9881#issuecomment-654627853.

nellh:
Yeah, this looks like it's just inconsistent performance on Circle. A higher timeout can be passed as the third argument to the it() call - the two node spawn tests that reference the valid_headers dir should just have the timeout raised. It's probably slow just because it ends up accessing more of node_modules in the new node process than the other tests.

PR #1185, opened 2021-02-11
Make the python package callable

kousu:
I did this really really quickly. It doesn't support all the options the javascript version does. Particularly, it doesn't print explanations of what fails validation. The URL in the help string is wrong. The code is probably in the wrong files. But I wanted to throw this out here to get the conversation going.

### DIFF
--- a/bids-validator/setup.cfg
+++ b/bids-validator/setup.cfg
@@ -21,6 +21,10 @@ classifiers =
 [options]
 packages = find:
 
+[options.entry_points]
+console_scripts = 
+    bids-validator = bids_validator:main
+
 [options.package_data]
 bids_validator =
     rules/*.json
--- a/bids-validator/README.md
+++ b/bids-validator/README.md
@@ -50,7 +50,8 @@
    1. From a terminal run `pip install bids_validator` to acquire the
       [BIDS Validator PyPi package](https://pypi.org/project/bids-validator/)
    1. Open a Python terminal `python`
-   1. Import the BIDS Validator package `from bids_validator import BIDSValidator`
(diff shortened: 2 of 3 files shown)


kousu:
I haven't taken the time to the understand the difference between the javascript and python APIs, but it looks like the javascript one can report what went wrong and the python one just says 

https://github.com/bids-standard/bids-validator/blob/c1ad2821b4eb3676ed7e87c51e89613c9089f563/bids-validator/bids_validator/bids_validator.py#L85

which makes the result opaque.

Also the javascript CLI doesn't allow you to pass in individual files, only a complete BIDS folder, while the python one does work on individual files, but only if they are relative paths relative to some root. I'm not really sure how to unify these.

bids-validator/setup.cfg
```diff
@@ -21,6 +21,10 @@ classifiers =
 [options]
 packages = find:
 
+[options.entry_points]
+console_scripts = 
+    bids-validator = bids_validator:main
```
effigies, line 26: Given that the JS CLI is also called `bids-validator`, this seems unwise.

If you create a `__main__.py` ([example](https://github.com/nipreps/fmriprep/blob/master/fmriprep/__main__.py)), then it can be called with `python -m bids_validator <args>`. Would that be sufficient, or do we need to come up with a name to call as an executable?

sappelhoff (commented):
May I ask what your use case is? I imagine the python bids-validator as a CLI tool to be quite underwhelming --> it really just checks the *name* of the file and tells you whether it's BIDS-conform or not.

PR #1201, opened 2021-02-27
[MRG] Adding check for participant_id pattern.

### LINKED ISSUE #1194: Add check for validity of the participants.tsv participant_id column
Raised in: https://github.com/bids-standard/bids-specification/issues/737

This would check that if a participants.tsv file is present, then the `participant_id` MUST be of a certain form: `sub-<label>`.

adam2392:
Closes: #1194


sappelhoff (approved):
LGTM

PR #1203, opened 2021-03-01
Add npmignore

rwblair:
Ignore test data, should reduces npm tarball by 50~mb.


bids-validator/.npmignore
```diff
@@ -0,0 +1 @@
+tests/data
```
sappelhoff, line 1: couldn't we then also ignore large parts of the test suite itself, given that it's unusable without the test data?

or is that a moot point, because the test suite is very low in file size anyhow?

rwblair: Correct on both counts. Was just trying to get the low hanging fruit.

PR #1232, opened 2021-03-18
Use pako to inflate remote file gzip buffers

nellh:
This improves browser compatibility with a thinner bundling process we're testing for OpenNeuro. Pako is already used for decompressing niifti headers and downstream packages can avoid shimming zlib in the browser since Pako works well there.


david-nishi (approved):
## Tested:
- test suite passing
- ran validator via yarn on dataset containing .nii.gz files

PR #1234, opened 2021-03-22
Browser build fixes for esbuild

nellh:
This is two more fixes required to include bids-validator in esbuild projects. Related: #1232


david-nishi (approved):
## Tested:
- test suite passing
- upload dataset with .nii files

PR #1235, opened 2021-03-22
Upgrade aws-sdk to 3.9.0

nellh:
Includes a fix for import issues in some browser environments.

See https://github.com/aws/aws-sdk-js-v3/pull/1991

This also fixes an issue with multiple lockfiles, the root one is updated with the contents of the bids-validator package lockfile.


david-nishi (approved):
## Tested:
- test suite passing
- ran `yarn`
- ran validator on dataset with s3 annexed files

PR #1239, opened 2021-03-30
FIX: Stop purging summary.subjects when doing phenotype checks

### LINKED ISSUE #1226: Misleading "No BIDS compatible data found for at least one subject" error
If you create a BIDS dataset that has a phenotype file (e.g. phenotype/NIH_Cognition_summary.tsv) with two participant id's (say 100 and 101) in it, but only have a subject directory for one of them (say 100) you'll get the following error (from v1.6.2 of the validator):

```
[ERR] No BIDS compatible data found for at least one subject. (code: 67 - NO_VALID_DATA_FOUND_FOR_SUBJECT)
		./sub-100
```

...despite the fact that BIDS-compatible data are in fact there for subject 100 - they're actually missing for subject 101. Remove the phenotype file and the error goes away.

(Interestingly, this doesn't happen if you have a participant id in the participants.tsv file that doesn't have a corresponding subject directory.)

The error message should ideally list the id's that lack subject directories, but it definitely should not list directories that do, in fact, have BIDS-compatible data.

sappelhoff:
closes #1226 

This func https://github.com/bids-standard/bids-validator/blob/1f50c70526877321d3a51afc311ddf854c9def1a/bids-validator/validators/tsv/checkPhenotype.js#L27-L30

used to swallow up potential participants when checking `phenotype`  specific data. As a result, a later check may say "oups you have data but no corresponding subject" ... which is a false positive warning.

I don't really get **why**working on the `subjects` input directly modifies the `summary.subjects` list, but it *does* happen.

This is a simple fix.

### DIFF
--- a/bids-validator/validators/tsv/checkPhenotype.js
+++ b/bids-validator/validators/tsv/checkPhenotype.js
@@ -25,7 +25,8 @@ const checkPhenotype = (phenotypeParticipants, summary) => {
 }
 
 const constructMissingPhenotypeEvidence = (fileParticipants, subjects) => {
-  const diffs = utils.array.diff(fileParticipants, subjects)
+  const subjectsClone = subjects.slice()
+  const diffs = utils.array.diff(fileParticipants, subjectsClone)
   const subjectsNotInSummarySubjects = diffs[0]
   const subjectsNotInFileParticipants = diffs[1]
   const evidenceOfMissingParticipants = subjectsNotInFileParticipants.length


rwblair:
The diff function from utils calls the splice function on the second argument. Splice is destructive in that it operates directly on the array. As opposed to some array functions like slice that will leave the original array untouched and only return a new value.

It doesn't come into play here but slice only does a shallow copy of the array so the the objects saved as values in subjects and subjectsClone are the same. Updating the object from one array would be reflected in the other.

sappelhoff:
> The diff function from utils calls the splice function on the second argument. Splice is destructive in that it operates directly on the array. As opposed to some array functions like slice that will leave the original array untouched and only return a new value.

thanks!

> It doesn't come into play here but slice only does a shallow copy of the array so the the objects saved as values in subjects and subjectsClone are the same. Updating the object from one array would be reflected in the other.

Interesting ... the change I added here does fix the problem though :thinking: 

Do you suggest to do something else?

rwblair:
> Do you suggest to do something else?

Lots of ways to do a shallow copy, no preferred method is coming to mind. Curious what @nellh thinks.

PR #1114, opened 2020-11-17
FIX: Normalize TSV end-of-line without stripping all whitespace

effigies:
Fixes #1113.

TODO:

* [x] Test that empty columns (e.g., `\t\t` or `\t$`) raise an error

PR #1240, opened 2021-03-30
Replace: FIX: Normalize TSV end-of-line without stripping all whitespace

### LINKED ISSUE #1113: Validator should ensure TSV files do not contain a tab character before the linebreak
As discussed [here](https://github.com/bids-standard/bids-matlab/issues/96), the bids validator accepts this file from openneuro:
https://openneuro.org/crn/datasets/ds003195/snapshots/1.0.3/files/participants.tsv
that contains a tab at the end of each line. The question is whether this is indeed a compliant file.

sappelhoff:
fixes #1113 

replaces to #1114 ... cherry picking the relevant commits from there and putting them on an up to date master branch.

### DIFF
--- a/bids-validator/validators/tsv/__tests__/validateTsvColumns.spec.js
+++ b/bids-validator/validators/tsv/__tests__/validateTsvColumns.spec.js
@@ -55,4 +55,18 @@ describe('validateTsvColumns', () => {
     const issues = validateTsvColumns(tsvs, {}, [])
     assert.lengthOf(issues, 0)
   })
+  it('should generate error with empty columns', () => {
+    const tsvs = [
+      {
+        contents: '\t',
+        file: file,
+      },
+      {
+        contents: '\t\t',
+        file: file,
+      },
+    ]
+    const issues = validateTsvColumns(tsvs, {})
+    assert.lengthOf(issues, 2)
+  })
 })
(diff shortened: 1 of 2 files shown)


bids-validator/validators/tsv/__tests__/validateTsvColumns.spec.js
```diff
@@ -55,4 +55,18 @@ describe('validateTsvColumns', () => {
...
+      {
+        contents: '\t',
+        file: file,
+      },
+      {
+        contents: '\t\t',
+        file: file,
+      },
+    ]
+    const issues = validateTsvColumns(tsvs, {})
```
effigies: Extra parameter added in ce5a154ff8d1365bf9d427abf5d31a0351544ecf.

```suggestion
    const issues = validateTsvColumns(tsvs, {}, {})
```

We should update the docs as well:

https://github.com/bids-standard/bids-validator/blob/d048d087a1ca4e7cac0310da275cd4a45e1c425a/bids-validator/validators/tsv/validateTsvColumns.js#L48-L53

effigies:
Tried to install locally to track down the test failure, but can't get a Node environment that will run tests. Good luck!

sappelhoff:
> Tried to install locally to track down the test failure, but can't get a Node environment that will run tests. Good luck!

I appreciate the effort :smile:

rwblair:
Thanks @effigies that did it.

rwblair:
I'm seeing this generate errors on ds000117 in bids-examples. I made it a global search replace in this commit e59124e

@effigies  Thoughts on if this is an overtly bad idea? I don't think people are sprinkling carriage returns in willy nilly.

effigies:
Nope, looks right. I assumed replace was global.

PR #1249, opened 2021-04-07
FIX: Strip Unicode byte order mark from TSV files

rwblair:
bids-examples ds000248 has this in its participants.tsv.

### DIFF
--- a/bids-validator/validators/tsv/tsvParser.js
+++ b/bids-validator/validators/tsv/tsvParser.js
@@ -3,7 +3,11 @@
  * Module for parsing TSV (and eventually other formats)
  */
 
-const normalizeEOL = str => str.replace('\r\n', '\n').replace('\r', '\n')
+const normalizeEOL = str =>
+  str
+    .replace('\r\n', '\n')
+    .replace('\r', '\n')
+    .replace(/\uFEFF/g, '')
 const isContentfulRow = row => row && !/^\s*$/.test(row)
 
 function parseTSV(contents) {
@@ -11,11 +15,16 @@ function parseTSV(contents) {
     headers: [],
     rows: [],
   }
+  console.log('------------------')
   content.rows = normalizeEOL(contents)
     .split('\n')
     .filter(isContentfulRow)
     .map(str => str.split('\t'))
-  content.headers = content.rows.length ? content.rows[0] : [] 
+  console.log(content.rows[0][0][0])
+  if (content.rows[0][0][0]) {
+    console.log(content.rows[0][0][0].charCodeAt())
+  }
+  content.headers = content.rows.length ? content.rows[0] : []
   return content
 }
 
--- a/bids-validator/validators/tsv/tsvParser.js
+++ b/bids-validator/validators/tsv/tsvParser.js
@@ -15,15 +15,10 @@ function parseTSV(contents) {
     headers: [],
     rows: [],
   }
-  console.log('------------------')
   content.rows = normalizeEOL(contents)
     .split('\n')
     .filter(isContentfulRow)
     .map(str => str.split('\t'))
-  console.log(content.rows[0][0][0])
-  if (content.rows[0][0][0]) {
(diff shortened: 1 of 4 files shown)


bids-validator/validators/tsv/tsvParser.js
```diff
@@ -3,7 +3,11 @@
  * Module for parsing TSV (and eventually other formats)
  */
 
-const normalizeEOL = str => str.replace('\r\n', '\n').replace('\r', '\n')
+const normalizeEOL = str =>
+  str
+    .replace('\r\n', '\n')
+    .replace('\r', '\n')
+    .replace(/\uFEFF/g, '')
```
effigies: The existing function is self-documenting, and this change makes it less clear. What about a second function?

```suggestion
const stripBOM = str => str.replace(/^\uFEFF/, '')
const normalizeEOL = str => str.replace('\r\n', '\n').replace('\r', '\n')
```

Also, I use `/^\uFEFF/` since BOM should only be valid at the start of a stream.

rwblair: Truth and wisdom, all of it.

sappelhoff:
should we warn users to remove this `&#65279` character? Or is it expected to be there? Why do only some datasets / files have it?

effigies:
https://en.wikipedia.org/wiki/Byte_order_mark

rwblair:
I have one inclination to just close this PR, fix the examples and have users strip it themselves if they want to pass the validator. I have no idea how common its usage is.

Node apparently decided against stripping it by default:
https://github.com/nodejs/node/issues/20649#issuecomment-388016410

sappelhoff:
I remember @hoechenberger having strong opinions about this issue (BOM) before, so I am tagging him here in case he wants to chime in.

hoechenberger:
> I remember @hoechenberger having strong opinions about this issue (BOM) before, so I am tagging him here in case he wants to chime in.

The BOM is super annoying, but should be supported unless we specifically state that we don't. This is what most Windows / Excel users will end up with when saving TSV files. It should not cause the validator to fail.

The situation is different when it comes to JSON, as the BOM is illegal there. But for any other type of file, we should support it unless otherwise specified.

effigies:
New test is passing. I don't know what's up with the test-bids-examples.

sappelhoff:
looking into it: https://github.com/bids-standard/bids-validator/pull/1240

--> they were also passing on that PR.

BUT: that's because in the workflow, we do not install the PR specific bids-validator ... but `master` instead.

So I think that's the case here as well. If we could somehow get the bids-examples github action workflow to take the PR files as "validator" instead of master, that'd solve it (I think)

https://github.com/bids-standard/bids-validator/blob/326d7c8d0599d7d9e4d7194ed1e310f054ba8f66/.github/workflows/test-bids-examples.yml#L30-L33

effigies:
Got it.

effigies:
Does not seem to be resolved by using this branch.

sappelhoff:
but we seem to get a different error now: `ReferenceError: files is not defined`

effigies:
That's the error we've been getting.

sappelhoff:
... compared to: a couple of commits earlier in this PR 0885bb7

https://github.com/bids-standard/bids-validator/runs/2289200555

sappelhoff:
okay, weird at some point the error changed from what I was thinking about all the time:

image

to this new one `ReferenceError: files is not defined`

sappelhoff:
also, looking at the bids examples repo workflow that ran 7 hours ago using bids-validator `master` failed with the other error (not ReferenceError: files is not defined)

https://github.com/bids-standard/bids-examples/runs/2333089352?check_suite_focus=true

effigies:
😠

sappelhoff:
:rofl: something is seriously wrong. Changing some innocent lines on bids-examples now yields:

image

https://github.com/bids-standard/bids-examples/pull/270

PR #1251, opened 2021-04-07
Fix circular dependency issue

nellh:
This fixes an import loop: utils/files/index.js > utils/files/readDir.js > utils/files/readFile.js > utils/files/testFile.js > utils/options.js

### DIFF
--- a/bids-validator/utils/options.js
+++ b/bids-validator/utils/options.js
@@ -1,5 +1,5 @@
 import path from 'path'
-import files from './files'
+import getFileStats from '../utils/files/getFileStats'
 import json from './json'
 import isNode from '../utils/isNode.js'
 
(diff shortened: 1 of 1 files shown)


effigies:
@nellh I notice that there's still a reference to `files` that you probably didn't catch because the `.` is on the next line: https://github.com/bids-standard/bids-validator/blob/f75de798563c50a72871b3daf8a045dfeeeebe62/bids-validator/utils/options.js#L60-L61

I think this is breaking `master`.

cc @rwblair who is also looking into this at the moment.

nellh:
@effigies @rwblair Thanks for catching that. I saw it got fixed in 1083e63ee0d5c4afb572a8f626e51c45d6e3f403

effigies:
Cool. Wasn't sure how quick it was going to be to solve when I sent that...

PR #1261, opened 2021-04-09
FIX JSON schema: inclusiveMaximum -> maximum, and letter casing

sappelhoff:
there is no keyword `inclusiveMaximum` in the JSON schema (see [spec](https://json-schema.org/understanding-json-schema/reference/numeric.html?highlight=minimum#range)). Conceptually, this is `maximum`.

Furthermore, there are no `Minimum` and `Maximum` keys in the JSON schema. Only `minimum` and `maximum` --> letter casing is important (I think).

### DIFF
--- a/bids-validator/validators/json/schemas/M0Scan.json
+++ b/bids-validator/validators/json/schemas/M0Scan.json
@@ -3,7 +3,7 @@
   "properties": {
     "EchoTime": {
       "type": "number",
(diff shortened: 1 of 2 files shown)


sappelhoff (commented):
test failures in bids-examples are due to https://github.com/bids-standard/bids-validator/pull/1249

PR #1264, opened 2021-04-09
Move bids-validator source files into src directory

nellh:
This is enable easier bundling of just the source without data or tests and support multiple entry points. Utility scripts that are not part of the actual bids-validator package can be placed above the src directory and not be parsed by compilers/bundlers without configuring a more complex pattern for all possible source files.


nellh:
@rwblair This is the most likely to cause merge conflicts of the changes I was hoping to get in, happy to help rebase things if needed. End goal is a static bundle that avoids the need for npm/yarn to install the validator.

david-nishi (approved):

sappelhoff:
does this have an impact on user and/or dev installation/documentation?

PR #1267, opened 2021-04-14
fix deployment step in circle ci

rwblair:
https://app.circleci.com/pipelines/github/bids-standard/bids-validator/834/workflows/60dc40da-837a-4b10-bb9b-09ce00a34367/jobs/8426

Validator site is not updating because of this.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -129,7 +129,7 @@ jobs:
             pip install dist/*.tar.gz
   deployment:
     docker:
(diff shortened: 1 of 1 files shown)


rwblair:
node12-alpine uses alpine 3.11 still.

Version of rust installable in this version of alpine is too old to build cryptography python package.

rwblair:
ran portion of the deployment step in  bb052f1 and cryptography installed properly.

PR #1270, opened 2021-04-19
ENH: Document --no-color option, respect NO_COLOR environment variable

rwblair:


rwblair:
Cannot figure out why the `--no-color` flag doesn't work in circleci jest run. I ssh'd into the instance and ran the same command that child_process.spawn calls and it worked as expected. Tried inserting `--no-colors` as well.

bids-validator/bin/bids-validator
```diff
@@ -25,6 +25,9 @@ const argv = require('yargs')
   .describe('verbose', 'Log more extensive information about issues')
   .boolean('json')
   .describe('json', 'Output results as JSON')
+  .boolean('no-color')
+  .describe('no-color', 'Disable colors in output text.')
+  .default('no-color', false)
```
effigies, line 30: Where is the effect of passing `--no-color`?

rwblair: https://github.com/Marak/colors.js#enablingdisabling-colors

Built in from the colors library.

effigies: Ah, so this is just documentation?

rwblair: Yee that was the idea.

effigies: Okay. I'm going to try to install locally to test, but otherwise I'd say this is good to go.

PR #1272, opened 2021-04-21
Fix a missing callback for some nifti header checks

nellh:
Looks like the scope for this callback was wrong, leading to an early exit if the `mergedDictionary.invalid` was true while checking nifti headers. An example affected dataset is [OpenNeuro ds003459:1.0.1](https://openneuro.org/datasets/ds003459/versions/1.0.1). Without this change this dataset does not return any validator output.


effigies (approved):
Looks correct, but holy god what a function.

nellh:
> Looks correct, but holy god what a function.

Yeah, it would be nice to break this one up...

Issue #1266, opened 2021-04-14
`next export` (npm run web-export) doesn't generate an index page


linusdm:
Although the documentation explains that running `npm run web-export` can help "to ensure that the web application compiles successfully in production", the output isn't ready to be deployed as a static site. An index page is missing. It's unclear to me whether this was done intentionally. Adding an index.js file to the pages directory easily fixes this (the component in this index.js file won't be used, as the _app.js file doesn't render any page components, another option would be to move logic from the _app.js file to index.js).

More generally I'm a bit confused why this is a nextjs project in the first place. Since most logic depends on code running client-side, it seems an odd choice. Although next.js is a superb framework for static site generation and server side rendering, none of those features are used here. Exporting the next.js site works (with a minor fix, as suggested above), but I think a leaner setup can be achieved with CRA (create-react-app), which can be easily deployed as a static site.
Maybe I'm missing something, or maybe features are in the pipeline that will require server-side handling of stuff. But at our university deployment of an online version of the BIDS validator would be even simpler without the next.js dependency.

effigies:
I'm not qualified to speak on your questions, so I'll leave that to another. But, out of curiosity, what is the motivation for a separate deployment of the validator?

linusdm:
Good question. At our research group we're trying to implement the BIDS specification. A very valuable step towards compliance would be a low-threshold online deployment where researchers can easily check whether their local files comply with the specification. This way we avoid any friction and hope for better adoption.

I would like to build the web-validator locally (or as part of a CI/CD pipeline), and deploy it somewhere on a shared server to make it available to everyone involved in our research group.

effigies:
That seems reasonable, but I just want to check whether this makes sense for your setup. A couple good reasons:

1) Your researchers cannot access https://bids-standard.github.io/bids-validator/ from their computers due to network restrictions.
2) Local control over the version of the validator that is deployed.

If the concern is transferring private data off-network, then I don't think it's a good investment of time. All validation is run in the browser, not transferred to a remote server. If you don't trust this, I think it's not going to be easily resolved without careful sandboxing. `github.io` websites don't provide access to server-side data, so if we were trying to collect data we would need to do it by sending it to another server, as would a malicious actor inserting code into the validator. This code would still be run from browsers visiting a local deployment, so only network restrictions would be effective at defeating this case.

linusdm:
[...]

PR #1273, opened 2021-04-22
Fixes for next build and export in deployment

rwblair:
Would explain #1266

Still need to spruce up the commands leading to the build and export. Npm commands are a mess right now.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -127,16 +127,12 @@ jobs:
             python --version
             pip install -U pip setuptools>=27.0 wheel
             pip install dist/*.tar.gz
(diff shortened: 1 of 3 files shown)


rwblair:
So npm run web-export still works in the gh-pages test workflow.
Given how many problems having the pypi and gh-pages deployment in the same workflow has caused I'm going to seperate them out. Make the pypi deploy look like the pypi test and the gh-pages deploy look like the gh-pages test workflow environment.

effigies (approved):
AFAICT this looks good. I don't think we'll be sure until we make another release...

PR #1274, opened 2021-04-23
Include unused sidecar HED strings in event-level validation

happy5214:
Previously, sidecar HED strings that are not referenced in the TSV file were only validated at the string level, not at the event or dataset levels. This would cause issues with definitions once support for those is implemented in hed-validator, since definitions will be included in sidecar HED strings that are not referenced in the TSV. This commit keeps a log of which sidecar strings have been referenced in the TSV and dumps the unused ones into the TSV before passing to the dataset-level validation (which includes event-level).

The tests do not currently pass, since they include invalid strings, previously skipped, that are now being validated (and failed). I need to come up with a good solution for this problem.

### DIFF
--- a/bids-validator/validators/events/hed.js
+++ b/bids-validator/validators/events/hed.js
@@ -30,6 +30,7 @@ export default function checkHedStrings(events, headers, jsonContents, dir) {
       let hedStringsFound = false
       const sidecarIssueTypes = {}
       let sidecarErrorsFound = false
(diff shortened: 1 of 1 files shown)


sappelhoff:
@happy5214 is this still relevant or should we close it?

happy5214:
This was fixed upstream in  https://github.com/hed-standard/hed-javascript/pull/56, so we can close this.

PR #1276, opened 2021-04-27
1245 separate endpoints

david-nishi:
This is the first step in creating two distributions of the validator for both ES and CommonJS modules. The next will be to abstract file handling to a common API that will be passed from either endpoint to the common codebase.

## Major changes:
- Separated endpoints to /commonjs and /esm
- pulled the start function out of bids and duplicated into each endpoint (this is where the different file handling will be done)
- removed `require()` calls from the common codebase


nellh (approved):

sappelhoff:
@nellh / @rwblair  is this still relevant, or should we close it?

PR #1280, opened 2021-05-11
Fix for HED validation issue file field

nellh:
Few fixes related to https://github.com/OpenNeuroOrg/openneuro/issues/2118

Some of the HED issues returned a string for the file field. Almost every other Issue passes the file object with a path, not the string. OpenNeuro accepts any object here but not a string. This fix does change which file is returned in the issue, instead of the sidecar you get the event file and the sidecar name is provided as evidence.

image

* Fixes the misleading jsdoc on Issue
* Extends a related test to check for this
* Fixes the issues to use a file object

### DIFF
--- a/bids-validator/tests/events.spec.js
+++ b/bids-validator/tests/events.spec.js
@@ -1342,6 +1342,9 @@ describe('Events', function() {
         assert.strictEqual(issues.length, 3)
         assert.strictEqual(issues[0].code, 210)
         assert.strictEqual(issues[1].code, 208)
+        // Check for correct file properties
+        expect(issues[0].file).toHaveProperty('path')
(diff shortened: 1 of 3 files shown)


sappelhoff:
cc @happy5214

rwblair (approved):

happy5214:
This is only an interim solution. The sidecar issues that were "fixed" in this PR actually have nothing to do with the specific event TSV files, other than using them to generate the list of sidecar JSON files to pull data from. However, the sidecar `*_events.json` path objects (assuming they actually exist; I don't remember that part of the code off the top of my head) aren't passed into this function. Using those would be the correct solution. There needs to be an issue opened for that.

Issue #1244, opened 2021-04-05
Abstract dataset file handling from validation


nellh:
One of the more difficult aspects of working with the validator code is understanding and handling support for different runtime environments (various Node.js versions, browsers, different embedding use cases). Most of this could be avoided by separating the initial file tree ingest step and file access from validation logic. The validation phase would use one consistent file access API to hide any environment specific details.

There's two relevant web APIs that could serve as the foundation for this. Lower level is the [File API](https://w3c.github.io/FileAPI/#file-section) which is created by input tag file selections. Higher level is [File System Access](https://wicg.github.io/file-system-access/) which skips the input tag step and provides a directory tree navigation providing File API objects.

Right now the validator uses the browser File API in some places but better would be File System Access to bring some higher level file access to validation as there are a number of rules that perform tree level cross checks (currently with directory access implemented for each case individually).

nellh:
Thinking about it more, if we pass in a FileSystemDirectoryHandle object as the main API for calling the validator that would give validation rules the ability to test across trees of files and skip the awkward Form APIs in when used in web applications. It also exposes the ability to handle things like permission errors for files within a dataset more gracefully.

Here's a [guide for how the API works for consumers](https://web.dev/file-system-access/) running in a browser. For Node, we'd wrap the native fs module with a matching interface. The lower level objects (FileList, File, Blob, etc) have existing library implementations for Node.

PR #1281, opened 2021-05-11
Enh: Raise warning if recommended README file is too small

### LINKED ISSUE #1279: How to check/"stat" text file within BIDS validator?
I want to add a feature to the bids-validator: 

- if the recommended README file is present:
- check how "large" it is (either how many lines, or how much content?)
- raise a warning if it's only very little (threshold to be determined)

I think it could go well after this line as an `else`:

https://github.com/bids-standard/bids-validator/blob/b40aa16a4250bb4aaacd2c063dad108084f6b83f/bids-validator/validators/bids/checkReadme.js#L10-L12

How would I read the file contents? ... or maybe even a simply `stat README` (like in bash) and extracting the size would be sufficient for me to either say: "yup, that seems large enough", versus "that's a suspiciously small README. I'll raise a warning to make the user aware that the README is crucial and that they please spend more effort on it"

Related to 
- https://github.com/bids-standard/bids-specification/issues/662
- https://github.com/bids-standard/bids-specification/pull/793

sappelhoff:
closes #1279 

as discussed in https://github.com/bids-standard/bids-specification/issues/662, the README is a very important file for making sense of a dataset. Next to providing a template for a README (see also https://github.com/bids-standard/bids-specification/pull/793), one thing we wanted to add was a simple check that raises a warning for suspiciously small README files.

That's what I do here. I have a very threshold for size right now: "150"

Practically that means this [README](https://github.com/bids-standard/bids-examples/blob/3f503bf19823284754a53796551a05946d49e4e8/pet001/README#L1) would just barely pass (size: 176):

> This dataset consists of a single PET measurement of a pig using [11C]CIMBI-36 to measure serotonin 2A availability. There are arterial measurements available for this tracer.

Whereas this [README](https://github.com/bids-standard/bids-examples/blob/3f503bf19823284754a53796551a05946d49e4e8/7t_trt/README#L1) would fail (size: 12):

> 7t_trt data

To get an impression on how best to tune this "size" param, you can use `find . -name 'README' -exec stat --printf="%n %s\n" {} +;` from the root of bids-examples.

My threshold currently is a bit too low, but it's hard to determine what would be a good thing to encourage best practice without annoying everybody.

cc @Remi-Gau

### DIFF
--- a/bids-validator/validators/bids/checkReadme.js
+++ b/bids-validator/validators/bids/checkReadme.js
@@ -14,7 +14,7 @@ const checkReadme = fileList => {
   } else {
     // Get the README file object
     for (var key in fileList) {
-      if (fileList[key].name == '/README') {
+      if (fileList[key].relativePath == '/README') {
         var readmeFile = fileList[key]
         break
       }
--- a/bids-validator/validators/bids/checkReadme.js
+++ b/bids-validator/validators/bids/checkReadme.js
@@ -22,9 +22,9 @@ const checkReadme = fileList => {
 
     // Check size and raise warning if too small
     const size = !isNode ? readmeFile.size : readmeFile.stats.size
-    var failsSizeRequirement = size <= 5000
+    var failsSizeRequirement = size <= 150
     if (failsSizeRequirement) {
-      issues.push(new Issue({ code: 101 }))
+      issues.push(new Issue({ code: 213 }))
     }
   }
   return issues
--- a/bids-validator/validators/bids/checkReadme.js
+++ b/bids-validator/validators/bids/checkReadme.js
@@ -5,27 +5,17 @@ const Issue = require('../../utils').issues.Issue
 const checkReadme = fileList => {
   const issues = []
   const fileKeys = Object.keys(fileList)
-  const hasReadme = fileKeys.some(key => {
-    const file = fileList[key]
-    return file.relativePath && file.relativePath == '/README'
-  })
-  if (!hasReadme) {
-    issues.push(new Issue({ code: 101 }))
-  } else {
-    // Get the README file object
-    for (var key in fileList) {
-      if (fileList[key].relativePath == '/README') {
-        var readmeFile = fileList[key]
-        break
-      }
-    }
-
-    // Check size and raise warning if too small
+  const readmeFile = Array.from(Object.values(fileList)).find(
+    file => file.relativePath && file.relativePath == '/README',
+  )
+  if (readmeFile) {
     const size = !isNode ? readmeFile.size : readmeFile.stats.size
     const failsSizeRequirement = size <= 150
     if (failsSizeRequirement) {
       issues.push(new Issue({ code: 213 }))
     }
+  } else {
+    issues.push(new Issue({ code: 101 }))
   }
   return issues
 }
--- a/bids-validator/validators/bids/checkReadme.js
+++ b/bids-validator/validators/bids/checkReadme.js
@@ -14,7 +14,7 @@ const checkReadme = fileList => {
   } else {
     // Get the README file object
     for (var key in fileList) {
-      if (fileList[key].name == 'README') {
+      if (fileList[key].name == '/README') {
         var readmeFile = fileList[key]
         break
       }
--- a/bids-validator/validators/bids/checkReadme.js
+++ b/bids-validator/validators/bids/checkReadme.js
@@ -1,3 +1,5 @@
+import isNode from '../../utils/isNode'
+
 const Issue = require('../../utils').issues.Issue
 
 const checkReadme = fileList => {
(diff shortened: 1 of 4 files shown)


bids-validator/validators/bids/checkReadme.js
```diff
@@ -9,6 +11,21 @@ const checkReadme = fileList => {
   })
   if (!hasReadme) {
     issues.push(new Issue({ code: 101 }))
+  } else {
+    // Get the README file object
+    for (var key in fileList) {
+      if (fileList[key].relativePath == '/README') {
```
effigies: Is this requiring it to be `/README`? Don't we allow `/README.md` and maybe `/README.rst`?

sappelhoff: we actually don't and that's something I've been thinking about raising. There are several datasets out there that bidsignore README.md and have to live with a warning that they don't have README.

That's a separate issue though, I think.

sappelhoff: example: https://gin.g-node.org/lnnrtwttkhn/highspeed-bids

effigies: I believe OpenNeuro will render markdown, though it might be we just interpret the contents of README as markdown, as opposed to using the extension.

I would be +1 for permitting markdown, at least. It opens up a smallish can of worms as we now get into the role of arbitrating the set of acceptable markup languages.

bids-validator/validators/bids/checkReadme.js
```diff
@@ -9,6 +11,21 @@ const checkReadme = fileList => {
   })
   if (!hasReadme) {
     issues.push(new Issue({ code: 101 }))
+  } else {
+    // Get the README file object
+    for (var key in fileList) {
```
nellh:
```suggestion
    for (const file of fileList) {
```
`for ... in` loops over properties which may not always be files. `for ... of` loops over the elements of the iterable. The existing check makes the same mistake too.

This could replace the `fileKeys.some` and bind readmeFile earlier for both checks.

```js
const readmeFile = Array.from(fileList).find(file => file.relativePath && file.relativePath == '/README')
if (readmeFile) {
  // size check
} else {
  // no readme
}
```

bids-validator/validators/bids/checkReadme.js
```diff
@@ -9,6 +11,21 @@ const checkReadme = fileList => {
...
+      if (fileList[key].relativePath == '/README') {
+        var readmeFile = fileList[key]
+        break
+      }
+    }
+
+    // Check size and raise warning if too small
+    const size = !isNode ? readmeFile.size : readmeFile.stats.size
+    var failsSizeRequirement = size <= 150
```
nellh:
```suggestion
    const failsSizeRequirement = size <= 150
```

bids-validator/validators/bids/checkReadme.js
sappelhoff: This is much nicer, thanks. I only had to make a minor change to use `Array.from(Object.values(fileList))` instead - else it worked (at least locally).

bids-validator/validators/bids/checkReadme.js
```diff
@@ -1,13 +1,20 @@
...
-  const hasReadme = fileKeys.some(key => {
-    const file = fileList[key]
-    return file.relativePath && file.relativePath == '/README'
-  })
-  if (!hasReadme) {
+  const readmeFile = Array.from(Object.values(fileList)).find(
+    file => file.relativePath && file.relativePath == '/README',
+  )
+  if (readmeFile) {
+    const size = !isNode ? readmeFile.size : readmeFile.stats.size
```
effigies, line 12: This works on an `Array`? Apologies for the dumb JS question.

VisLab:
The BIDS specification states this about README:

>  In addition a free form text file (README) describing the dataset in more details SHOULD be provided.  
>  The README file MUST be either in ASCII or UTF-8 encoding.

The display on OpenNeuro does some type of layout and bolds some headings --- though I am not clear on what the rules are.

It would be really nice if the specification were changed to also allow GitHub markdown. It would allow users to make more readable READMEs.

nellh:
> The BIDS specification states this about README:
> 
> > In addition a free form text file (README) describing the dataset in more details SHOULD be provided.
> > The README file MUST be either in ASCII or UTF-8 encoding.
> 
> The display on OpenNeuro does some type of layout and bolds some headings --- though I am not clear on what the rules are.
> 
> It would be really nice if the specification were changed to also allow GitHub markdown. It would allow users to make more readable READMEs.

OpenNeuro parses it as Markdown using [markdown-to-jsx](https://www.npmjs.com/package/markdown-to-jsx) and applies some CSS to the resulting elements to fit the rest of the page.

bids-validator/validators/bids/checkReadme.js
nellh: Not sure which part your question is referring to...

Array.from shallow copies iterables (like FileList) as a new Array instance.

Array.find returns the first matching element or `undefined` if nothing matched the lambda.

Object.values is needed for the bids-validator specific Node.js variant of FileList. With #1244 done, it could be dropped because ```Array.from(fileList: FileList)``` does work.

effigies: Just meant really specifically the `readmeFile.size` and `readmeFile.stats.size`. Does array pass through properties like this? Or did I misunderstand and `readmeFile` isn't an array?

nellh: readmeFile is a file object because of the Array.find() call.

PR #1283, opened 2021-05-12
[FIX] Update regexp for scandate in PET schema

### LINKED ISSUE #1282: PET: Invalid pattern for ScanDate in PET schema
Currently the ScanDate field is specified as the following in the PET schema:

```
"ScanDate": {
      "type": "string",
      "pattern": "date"
    },
```

However, this only allows for "date" to be entered as a valid ScanDate when validating.

A quick fix would be to update the code to the following:

```
"ScanDate": {
      "type": "string",
      "pattern": "^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(.\d+)?Z?$"
    },

```

Similar to in checkAcqTimeFormat.js. 

https://github.com/bids-standard/bids-validator/blob/1e92dc1da31046e741c80081cfb67e54ac11a763/bids-validator/validators/tsv/checkAcqTimeFormat.js#L5

mnoergaard:
closes #1282 

This PR updates the regexp for scandate in the PET schema so it follows the ~ISO8601 format~ BIDS specific subset of the rfc3339 format.

### DIFF
--- a/bids-validator/validators/json/schemas/pet.json
+++ b/bids-validator/validators/json/schemas/pet.json
@@ -70,7 +70,7 @@
     "FrameDuration": { "type": "array", "items": { "type": "number" } },
     "ScanDate": {
       "type": "string",
-      "pattern": "date"
+      "pattern": "^(19|20)[0-9][0-9]-(0[0-9]|1[0-2])-(0[1-9]|([12][0-9]|3[01]))T([01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9]$"
     },
     "InjectionEnd": { "type": "number" },
 


sappelhoff (approved):

sappelhoff:
Thanks @mnoergaard

sappelhoff:
uhm, maybe I have been a bit too fast with merging. See: https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/09-positron-emission-tomography.html#time

For ScanDate it says: 

> Date of scan in the default unit "YYYY-MM-DD[Z]", with the Z indicator being optional for indicating UTC timezone (see Units).

So is it a good idea to also allow datetimes? Shouldn't that field not have been names `ScanDateTime` then?

The description of `ScanDate` does link to the Units section, which allows date specifications according to:

`^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(.\d+)?Z?$`

Shouldn't we permit that format then, instead of the one that you proposed (and that I just merged):

`^(19|20)[0-9][0-9]-(0[0-9]|1[0-2])-(0[1-9]|([12][0-9]|3[01]))T([01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9]$`

sappelhoff:
My suggestion:

1. Change the repexp from this PR to `^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(.\d+)?Z?$` to be consistent in what types of datetimes we allow (EDIT: we'd have to make the time part optional in the regexp)
2. Change the text in the BIDS spec for `ScanDate` like so:

```diff
- Date of scan in the default unit "YYYY-MM-DD[Z]", with the Z indicator being optional for indicating UTC timezone (see Units).
+ Date (and optionally time) of scan formatted as YYYY-MM-DD[Thh:mm:ss[.000000]][Z] (see also Units).
```

sappelhoff:
I reverted the changes from this PR in https://github.com/bids-standard/bids-validator/commit/a3dccf93df338880b9c996dfb93b147419dc7df1 until we have discussed a solution based on my comments above. Sorry for clicking merge before thinking this through properly

mnoergaard:
I think that sounds good. I originally proposed the `^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(.\d+)?Z?$` change, but my editor was not happy about that expression. Therefore, I found the other solution - which is not correct as you say, because of the timepart, and then the spec would need to be updated.

mnoergaard:
Or can the pattern simply be changed to "date-time"? (https://json-schema.org/understanding-json-schema/UnderstandingJSONSchema.pdf).

sappelhoff:
The `date-time` format doesn't fit with the BIDS usage of RFC3339 unfortunately, because we don't allow arbitrary offsets (`+00:00`): We only allow the UTC offset (`[Z]`), else it is "assumed that the time is with reference to the time in the recording site"

> Therefore, I found the other solution - which is not correct as you say, because of the timepart, and then the spec would need to be updated.

But even when adding `^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(.\d+)?Z?$` (with the addition of making the time part optional in that regexp), we'd have to amend (extend/clarify) the spec, I think. In the direction that I outlined above with that short example sentence. WDYT?

PR #1285, opened 2021-05-14
Ignore package lockfiles and use lifecycle hook to create dev versions

### LINKED ISSUE #1284: Package and Sub-Package versions are messed up
I already briefly discussed this with @rwblair over the Brainhack mattermost. It seems to me that we need to update/tweak the release process in some way, because currently the package versions and sub-package versions are not correct. This is an attempt to outline the problems:

`lerna.json` shows 1.7.1, but should be 1.8.0.dev0 or 1.7.2.dev0 to signal that `master` is the development version of the upcoming release.

https://github.com/bids-standard/bids-validator/blob/15083df81384e5e5f83887617bd6972f014f075a/lerna.json#L1-L4

This was originally discussed and "fixed" here:

- https://github.com/bids-standard/bids-validator/issues/1080
- https://github.com/bids-standard/bids-validator/commit/a96790e11959f85f4366a42e7c20bb598667b008

but we somehow lost these benefits again.

web validator shows version 1.7.0, but this should **at least** be 1.7.1 (the current release), but actually rather 1.8.0.dev0 (see above)

https://github.com/bids-standard/bids-validator/blob/15083df81384e5e5f83887617bd6972f014f075a/bids-validator-web/package.json#L3

The bids-validator sub-package has no `package.lock`? bids-validator-web, and the overall package do have it :thinking: ... and the bids-validator sub-package seems to be tracked in the overall package.

Also, its version should also rather be `1.8.0.dev0` (or `1.7.2.dev0`) than the 1.7.1 that it is:

https://github.com/bids-standard/bids-validator/blob/15083df81384e5e5f83887617bd6972f014f075a/bids-validator/package.json#L3

Furthermore, the `package.lock` in the overall package (which somehow tracks the bids-validator sub-package dependencies, rather than that one having its own `package.lock`) shows a completely off version for bids-validator.

https://github.com/bids-standard/bids-validator/blob/15083df81384e5e5f83887617bd6972f014f075a/package-lock.json#L21

The web-package's `package.lock` version is yet again a different one: https://github.com/bids-standard/bids-validator/blob/15083df81384e5e5f83887617bd6972f014f075a/bids-validator-web/package-lock.json#L8

---

I think we need to:

1. fix this
2. document how to release in the future to avoid these problems
3. (automate step 2 as much as robustly possible)

nellh:
This removes an extra lockfile in the web package that was causing it to be treated as separate from the others. I added .gitignore entries for the npm lockfiles, we had that for yarn.lock but it didn't get changed with the switch to npm. npm 6 will make those lockfiles in the subpackages but this should keep it from breaking again in the repo even if someone is accidentally working with npm 6.

I'm also proposing tweaking the release process to automate the dev package version better. It's added here as a lifecycle hook to run after any successful version created with `lerna publish`.

Fixes #1284


rwblair:
Thank you for fixing my mess once again, had no idea about ignore-scripts and using postversion in scripts like that.

sappelhoff:
nice, thanks!

Two more things:

1. shouldn't the  `postversion` script have updated `lerna.json` and the `package.json` files to the "dev" version right after the release?

They are still at 1.7.2 now: https://github.com/bids-standard/bids-validator/blob/0c74992dabbf48d16e9d4866eb5622e11c2138d3/lerna.json#L1-L4

(but they are consistent, yay)


2.  The last, remaining [package.lock](https://github.com/bids-standard/bids-validator/blob/master/package-lock.json) in the root is still a little odd. It shows bids-validator with version 1.7.1, and bids-validator-web with 1.7.0. --> shouldn't a release also involve updating that? (at least so that it's 1.7.2. like the package.jsons and lerna.json?)


---

finally: running `npm prune` and `npm audit fix` seems to remove/fix some packages --> is that something reasonable to do right now?

nellh:
Yeah, I think it should have, I added a few fixes for it in #1286 but @rwblair - did that not work as expected?

rwblair:
Unfortunately prerelease tag can't be set with postversion command:
lerna notice cli v3.22.1
lerna info current version 1.7.2
lerna notice Current HEAD is already released, skipping change detection.
lerna success No changed packages to version

I haven't seen another package that automatically sets a dev tag right after publish. Eslint for example just leaves the latest regular version in package.json for developing on master. Jest has its master version set to 0.0.0, must be some fancy facebook deploy deal. Next has a sophisticated lerna setup, using canary releases where their master stays tracking latest stable version.

nellh:
Added another fix to #1286 for that.

rwblair:
re:versions in lockfile, package-lock.json is only updates after running an npm install after publishing a new version. Maybe weird interaction with lerna and npm workspaces.

sappelhoff:
> re:versions in lockfile, package-lock.json is only updates after running an npm install after publishing a new version. Maybe weird interaction with lerna and npm workspaces.

okay, fixed in https://github.com/bids-standard/bids-validator/pull/1287

and somehow you got the `-dev...` thing working :tada: looks good now. Thanks both!

PR #1287, opened 2021-05-14
run: npm audit fix

sappelhoff:
to top off the recent cleanups I ran `npm audit fix` from a fresh `npm clean-install` after `git reset --hard upstream/master`.

That updated the bids-validator versions in the package-lock.json to `1.7.3-dev.0` as they should be .

It also fixed some minor security issues.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
       }
     },
     "bids-validator": {
-      "version": "1.7.1",
+      "version": "1.7.3-dev.0",
       "license": "MIT",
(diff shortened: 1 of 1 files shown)


sappelhoff:
I reverted this in https://github.com/bids-standard/bids-validator/commit/fbaef203d28c8cb61871bee664464036481c84b6 because apparently this broke the GitHubPagesTest check in CircleCI.

No idea why and no time to figure it out right now -> but we should address the issue I wanted to fix with this in the future.

PR #1290, opened 2021-05-20
[FIX] Update ScanDate for PET schema

### LINKED ISSUE #1282: PET: Invalid pattern for ScanDate in PET schema

mnoergaard:
This PR updates the `ScanDate` regexp to be in accordance with the specification's definition for dates. 

closes #1282

### DIFF
--- a/bids-validator/validators/json/schemas/pet.json
+++ b/bids-validator/validators/json/schemas/pet.json
@@ -70,7 +70,7 @@
     "FrameDuration": { "type": "array", "items": { "type": "number" } },
     "ScanDate": {
       "type": "string",
-      "pattern": "date"
+      "pattern": "^(19|20)[0-9][0-9]-(0[0-9]|1[0-2])-(0[1-9]|([12][0-9]|3[01]))T([01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9]$"
     },
     "InjectionEnd": { "type": "number" },
 
--- a/bids-validator/validators/json/schemas/pet.json
+++ b/bids-validator/validators/json/schemas/pet.json
@@ -68,10 +68,6 @@
     "InjectionStart": { "type": "number" },
     "FrameTimesStart": {"type": "array", "items": { "type": "number" } },
     "FrameDuration": { "type": "array", "items": { "type": "number" } },
-    "ScanDate": {
-      "type": "string",
-      "pattern": "^\d{4}-\d{2}-\d{2}(T\d{2}:\d{2}:\d{2}(.\d+)?)?Z?$"
-    },
     "InjectionEnd": { "type": "number" },
 
     "AcquisitionMode": { "type": "string", "minLength": 1 },
--- a/bids-validator/validators/json/schemas/pet.json
+++ b/bids-validator/validators/json/schemas/pet.json
@@ -70,7 +70,7 @@
     "FrameDuration": { "type": "array", "items": { "type": "number" } },
     "ScanDate": {
       "type": "string",
-      "pattern": "^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(.\d+)?Z?$"
+      "pattern": "^\d{4}-\d{2}-\d{2}(T\d{2}:\d{2}:\d{2}(.\d+)?)?Z?$"
     },
     "InjectionEnd": { "type": "number" },
 
--- a/bids-validator/validators/json/schemas/pet.json
+++ b/bids-validator/validators/json/schemas/pet.json
@@ -70,7 +70,7 @@
     "FrameDuration": { "type": "array", "items": { "type": "number" } },
     "ScanDate": {
       "type": "string",
-      "pattern": "^(19|20)[0-9][0-9]-(0[0-9]|1[0-2])-(0[1-9]|([12][0-9]|3[01]))T([01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9]$"
+      "pattern": "^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(.\d+)?Z?$"
     },
     "InjectionEnd": { "type": "number" },
 


sappelhoff (commented):
You'll have to make the "time" part of the regexp optional, see: https://regex101.com/r/ayj2cx/1

with the current version, one would HAVE to supply a time (which is not what we want right?)

sappelhoff:
I think `^\d{4}-\d{2}-\d{2}(T\d{2}:\d{2}:\d{2}(.\d+)?)?Z?$` would do the job, but could you double check, please?

bids-validator/validators/json/schemas/pet.json
```diff
@@ -70,7 +70,7 @@
     "FrameDuration": { "type": "array", "items": { "type": "number" } },
     "ScanDate": {
       "type": "string",
-      "pattern": "date"
+      "pattern": "^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(.\d+)?Z?$"
```
mnoergaard:
```suggestion
      "pattern": "^\d{4}-\d{2}-\d{2}(T\d{2}:\d{2}:\d{2}(.\d+)?)?Z?$"
```

mnoergaard:
You are right, @sappelhoff! Thanks.

sappelhoff (approved):
+1, conditional on https://github.com/bids-standard/bids-specification/pull/798 being merged

bids-validator/validators/json/schemas/pet.json
```diff
@@ -68,10 +68,6 @@
     "InjectionStart": { "type": "number" },
     "FrameTimesStart": {"type": "array", "items": { "type": "number" } },
     "FrameDuration": { "type": "array", "items": { "type": "number" } },
-    "ScanDate": {
-      "type": "string",
-      "pattern": "date"
-    },
```
sappelhoff, line 74: The idea in https://github.com/bids-standard/bids-specification/pull/798 is now to keep this field, but deprecate it, recommending that people use AcquisitionTime metadata instead.

So we also need to keep this field in the validator. However, it must be adjusted because the current [pattern `date`](https://json-schema.org/understanding-json-schema/reference/string.html?highlight=date#dates-and-times) does not accept the optional `Z` to indicate UTC timezone.

The pattern should only accept: `"YYYY-MM-DD[Z]"`

The easiest would be to accept `\d\d\d\d-\d\d-\d\d[Z]*`, but that would allow things like 30th of February. Maybe somebody has an improvement.

mnoergaard: Can't we use your suggestion from previously? (i.e. `^\d{4}-\d{2}-\d{2}(T\d{2}:\d{2}:\d{2}(.\d+)?)?Z?$`)

sappelhoff: Unfortunately that has the same shortcoming, and it had some parts that we don't need anymore (the time)

mnoergaard: @effigies and @rwblair - do you have any suggestions here?

rwblair: ~Draft 7 has dates built in:~
~https://json-schema.org/understanding-json-schema/reference/string.html#dates-and-times~

~We are specifying draft 6 but we should probably bump this and see how it goes since 7 has so many useful things in it:~
https://github.com/bids-standard/bids-validator/blob/da684b211179f90c99823aff5d40e94b565097b7/bids-validator/validators/json/json.js#L4

Sorry didn't read the initial comment closely!

sappelhoff: yes, but I don't think it supports the exact date parsing that we want: `\d{4}-\d{2}-\d{2}[Z]{0,1}`, particularly the optional `Z` (EDIT: and no other UTC offsets)

rwblair: From the RFC mentioned in schema docs it does allow 'z':
https://datatracker.ietf.org/doc/html/rfc3339#section-5.6

Looking into if this will work for existing scandates.

rwblair: Was not able to get it to accept a 'z' on a date without a time.

Personally I am fine with a loose regex allowing for 30th of february especially if this is deprecated and not commonly used. Shoot ourselves more in the foot with clever regex than users will with nonsensical dates I figure.

mnoergaard: Thanks @rwblair - I am also fine with that solution. Right now the validator won't event accept anything reasonable for the `ScanDate`, so the datasets containing this will simply not be valid.

effigies: Ah, sorry, was out last week. I agree with @rwblair that `\d{4}-\d{2}-\d{2}Z?` is fine, and we don't need to worry about invalid dates.

There's some conceivable scope for a security failure, if somebody uses a really complicated date parser that tries to do something fancy when it gets an invalid date, but given that this was a short-lived field, I expect most tools will simply not use it.

sappelhoff: @rwblair and I discussed in the last maintainers meeting that we could just close this PR, because leaving the currently implemented `date`:

1. doesn't break existing data (because that's how it was since the start of PET validation)
2. is fine for 90% what used to be advertised in the spec: `YYYY-MM-DD`
3. Only the `Z` that was optionally allowed in the spec cannot be parsed
4. But it's probably fine if we don't cover/allow this for **future** datasets, because the field is deprecated and **future** datasets shouldn't use this field anyhow (see point 1, **past** datasets are fine, because they managed to navigate the problem from point 3 already)

effigies: I see. That seems fine. Do we have a way of suggesting that this be replaced with `acq_time` when this is found?

sappelhoff: Not currently, unfortunately - these fields just get checked with JSON schema. The machinery you are hinting at would be what we'd want to also warn about JSON fields / provide further info about some of those fields, and as far as I am informed we haven't made progress on that front.

effigies:
Sounds good. Closing this for now, and deprecation warnings can be part of a larger push.

PR #1298, opened 2021-06-14
Refactor HED module

### LINKED ISSUE #1205: Fails with internal error Error: Could not load HED schema version "Latest" from remote repository
I'm trying to run the BIDS validator from a singularity container on a HPC system. I built a container from the latest version on DockerHub (1.6.20) with the following command:
```
 $PWD/bids_validator.simg docker://bids/validator
```

I've tried running it on a compute node on the HPC and I get the following error:
```
(/vf/users/ABCD_MBDU/dcan_abcd_dcm_bids_env) [nielsond@cn2878 notebooks]$ singularity run --cleanenv --no-home -B /data/AB
CD_MBDU/abcd_bids/test_download:/data /data/ABCD_MBDU/singularity_images/bids_validator.simg /data                        
bids-validator@1.6.2                                                                                                      
                                                                                                                          
        1: [ERR] Internal error. SOME VALIDATION STEPS MAY NOT HAVE OCCURRED (code: 0 - INTERNAL ERROR)                   
                .undefined                                                                                                
                        Evidence: Error: Could not load HED schema version "Latest" from remote repository - "Error: getad
drinfo EAI_AGAIN raw.githubusercontent.com".                                                                              
    at /src/node_modules/hed-validator/utils/schema.js:40:13                                                              
    at tryCallOne (/src/node_modules/promise/lib/core.js:37:12)                                                           
    at /src/node_modules/promise/lib/core.js:123:15                                                                       
    at flush (/src/node_modules/asap/raw.js:50:29)                                                                        
    at processTicksAndRejections (internal/process/task_queues.js:76:11)                                                  
                                                                                                                          
        Please visit https://neurostars.org/search?q=INTERNAL ERROR for existing conversations about this issue.          
                                                                                                                          
                                                                                                                          
        Summary:                 Available Tasks:                 Available Modalities:                                   
        13 Files, 13.49MB        Stop-Signal Task                 T1w                                                     
        1 - Subject              Monetary Incentive Delay         epi                                                     
        1 - Session              n-back                                                                                   
                                 Resting State (eyes open)                                                                
                                                                                                                          
                                                                                                                          
        If you have any questions, please post on https://neurostars.org/tags/bids.                                       
                                                                                              
```
Working backwards, I found that if I build an image from version 1.5.7 it runs successfully without this error, but fails with version 1.5.8 with the same error. My guess is that the validator is trying to request some resource from raw.githubusercontent.com and getting stymied by the proxy rules the HPC has in place on the compute nodes. Can anyone tell me what the request it's trying to make is? Does anyone have any advice on the needed proxy settings?

### LINKED ISSUE #1292: Pass file objects for JSON event sidecars to HED event validation
The eventual fix for the issues raised in #1280 is to pass the file objects for the event sidecars through the event validation and onto the issue. I don't know how complicated that would be to accomplish.

happy5214:
closes #1205 
closes #1292

This commit refactors the HED module completely by splitting it into smaller functions and creating a new function that checks for HED data in the sidecars and TSV event files before loading a HED schema, thus addressing #1205. I've also taken the opportunity to fix #1292 by passing `jsonFiles` from `fullTest.js` through to the HED module and using that to find the file parameter to Issue, thus providing the issue with the actual sidecar name.


happy5214:
I reverted a dependency bump on our end and pushed a new version, so this PR also bumps the hed-validator dependency to 3.1.1. Disregard the 3.1.0 commit, which was deleted from the branch. This version brings several new checks and is a pretty high priority.

VisLab:
This pull request fixes one of the issues that people have encountered because they couldn't download the HED schema due to faulty network connections or proxy issues. The main change proposed in this PR besides some refactoring is to move the check for schema later in the process so that it is only checked if the data set actually has HED tags.  This partially addresses #1205 and should probably be done sooner rather than later. 
@rwblair @sappelhoff @effigies we would appreciate a review and suggestions on moving forward with this.

The second part of #1205 is to have a fall back hedschema stored with the hed-javascript package.  We will do that also in a week or two, but this will change is internal and will not affect the bids validator directly.

sappelhoff:
I'll leave the proper review to Ross, but @happy5214 could you investigate why some tests are failing?

happy5214:
> I'll leave the proper review to Ross, but @happy5214 could you investigate why some tests are failing?

I spent a significant chunk of time trying to understand the issue. They appear to be related to the pet blood TSV test data, and the issue isn't new. I'm pretty sure it's not directly related to HED, though I'm not quite sure why it's popped up again here. The best I could come up with is that the `_blood.json` file in the last testcase in `bids.spec.js` is not being included in the `jsonContents` dictionary, for whatever reason. I don't see what that could possibly have to do with my code, since anything I could have done would pop up in more places than just that.

happy5214:
> > I'll leave the proper review to Ross, but @happy5214 could you investigate why some tests are failing?
> 
> I spent a significant chunk of time trying to understand the issue. They appear to be related to the pet blood TSV test data, and the issue isn't new. I'm pretty sure it's not directly related to HED, though I'm not quite sure why it's popped up again here. The best I could come up with is that the `_blood.json` file in the last testcase in `bids.spec.js` is not being included in the `jsonContents` dictionary, for whatever reason. I don't see what that could possibly have to do with my code, since anything I could have done would pop up in more places than just that.

FWIW, I'd also like to note that I ran the CLI validator with this PR code on the test dataset in question, and it correctly generated an issue with code 211, so this problem is only occurring when it's run from the testcase.

bids-validator/tests/events.spec.js
```diff
@@ -1330,12 +1782,23 @@ describe('Events', function() {
...
+      const jsonFiles = [
+        {
+          relativePath: '/sub01/sub01_task-test_events.json',
+          path: '/sub01/sub01_task-test_events.json',
+        },
+        {
+          relativePath: '/dataset_description.json',
+          path: '/dataset_description.json',
+        },
+      ]
```
rwblair: Would it be helpful to define a couple arrays of json files to use in tests up top to prevent so many duplicate declarations? I'm not aware of any reason that they need to be redefined in each test other than that's the way its ~always~ been done.

rwblair: I can do that re-org in this PR or another, not necessary for merging.

bids-validator/validators/events/hed.js
```diff
@@ -23,213 +72,266 @@ export default function checkHedStrings(events, headers, jsonContents, dir) {
...
+
+function splitTsv(eventFile) {
+  const rows = eventFile.contents
+    .split('\n')
+    .filter(row => !(!row || /^\s*$/.test(row)))
+
+  const columnHeaders = rows[0].trim().split('\t')
+  const dataRows = rows.slice(1)
+  return [columnHeaders, dataRows]
+}
```
rwblair: Would there be any issue with using the tsv parser here?: 
bids-validator/validators/tsv/tsvParser.js

Catches issues we've had with byte order marks and the like.

rwblair:
@VisLab Is there a public dataset that should pass HED validation without warning? I tried the PR out on ds003645 and it was throwing HED_MISSING_VALUE_IN_SIDECAR, HED_TAG_EXTENSION, and HED_INVALID_SIDECAR_WARNING warnings.

VisLab:
The warnings on ds003645 were legitimate, but minor so we had postponed correcting them until the next version, which will include additional data. 

Last night we uploaded the corrections to the top-level `events.json` file (which fixed the warnings). However, now the data set  no longer validates because the `eeg.json` files don't "comply with the schema".  The data set validates with the BIDS validator outside of OpenNeuro.  

I will create an example data set based on ds003645 and do a PR to bids-examples in the next couple of days. @dungscout96 is working on the OpenNeuro version.

VisLab:
@rwblair:
I have created a [PR #281 on bids-examples](https://github.com/bids-standard/bids-examples/pull/281) with an example of HED in a BIDS dataset.

bids-validator/validators/events/hed.js
```diff
@@ -23,213 +72,266 @@ export default function checkHedStrings(events, headers, jsonContents, dir) {
...
+
+function validateDataset(hedStrings, hedSchema, eventFile) {
+  const [
+    isHedDatasetValid,
+    hedIssues,
+  ] = hedValidator.validator.validateHedDataset(hedStrings, hedSchema, true)
```
rwblair: As note to self from meeting earlier today, we should wrap this call in a try...catch in case the call to hedvalidator itself fails for whatever unknown reason. May reclaim one of the freed error codes for this case.

happy5214: Done. I freed up 209 and 210 here and used 209 for the new error code temporarily. The other issue codes will be freed in a later PR, since the hed-validator code that includes the HED 3 error codes hasn't been released yet.

bids-validator/validators/events/hed.js
happy5214: Done. It's much cleaner than my version, so why reinvent the wheel?

bids-validator/tests/events.spec.js
happy5214: Done. There was only one version of the array, so I moved it to the top of the HED describe block scope.

rwblair:
Made a couple changes to the PR happy5214/bids-validator/pull/1 wanted your approval/thoughts on them @happy5214 instead of pushing them directly to this PR.

PR #1299, opened 2021-06-15
Simplify HED issue codes

### LINKED ISSUE #1300: Reducing HED's foot print in the BIDS validator
We have been discussing how to reduce the foot print of HED in the BIDS validator for easier maintenance and would like some feedback as we proceed.  We see 3 changes that would improve the overall structure and usability.

1. Only load the HED schema if the data set explicitly has HED fields.  This is the subject of PR #1298.
2. Store the latest version of the HED schema in the HED JavaScript validator as a fall back so things can proceed in the case of a recalcitrant firewall. (We intended to do this in a couple of weeks when HED 3 is officially released. This change should be invisible with respect to BIDS.)
3. **Restructure the errors**.  The idea is that HED would only use two error codes (104 renamed as HED_ERROR and 105 renamed as HED_WARNING).  All of the details about the error would be provided in the evidence.

The restructuring would allow us to align the errors to the HED specification without impacting BIDS.  We would release the rest of the BIDS error codes used by HED.  @happy5214  and I have been looking at the current code base, and we think this restructuring would be feasible and would not involve too many changes on the BIDS side.  We would appreciate comments and or suggestions on this before we proceed with the restructuring. @sappelhoff @rwblair @effigies

happy5214:
This PR trims the number of HED issue codes to 7, one of which is temporary pending the inclusion of a validation check in an upcoming version of hed-validator, and simplifies the event tests.

Fixes #1300


happy5214:
The failing tests from the first commit should pass with hed-validator 3.1.0 (which added the new validation code that I deleted from bids-validator), but I can't seem to get npm to upgrade to that version. Assistance is requested.

rwblair:
@VisLab current error in the tests:

> Validating dataset eeg_hed_small/
> bids-validator@1.7.3-dev.0
> 
> 	1: [ERR] The validation on this HED string returned an error. (code: 104 - HED_ERROR)
> 		./task-FacePerception_events.json
> 			Evidence: ERROR: [HED_VALUE_INVALID] Invalid placeholder value for tag "Description/A white fixation cross on a black background in the center of the screen."

There's a couple hundred similar errors thrown.

VisLab:
Note: The correction I made on bids-examples PR 282 doesn't fix the issue.  It has to do with which characters are allowed in different versions of the schema.  @happy5214 is looking at the issue.

happy5214:
The issue was that it was treating spaces in values as invalid, which was due to me misreading the spec. This has been fixed in a patch version of hed-validator, and the bids-example datasets appear to pass now.

PR #1303, opened 2021-07-06
Clean up summary modalities

nellh:
This summary field currently mixes modalities, data types, and occasionally some other values by relying on file suffixes. This instead categorizes based on the modality tests used elsewhere in the validator and limits values to PET, MEG, EEG, iEEG, and MRI. Other values formerly reported as modalities will be broken out into their own data types field.


david-nishi (approved):

ckrountree:
Also covers Gitlab 263

PR #1308, opened 2021-07-19
Revert "feat: Extend modality summary with secondary modalities"

rwblair:
Reverts bids-standard/bids-validator#1306

bids-standard/bids-validator/issues/1307

Should be re-released as a major possibly. Testing to see if reversion fixes fmriprep related issue.


rwblair:
@effigies unfortunately reverting #1306 still causes the error to be thrown on datasets with t1s. I'll poke around to see what else it may have been.

PR #1309, opened 2021-07-19
Add summary dataTypes field

### LINKED ISSUE #1307: Changing T1w modality to MRI causes all datasets to report NO_T1W
#1306 now causes error 53 (NO_T1W) to be emitted in all cases.

https://github.com/bids-standard/bids-validator/blob/1bfe6170d30cc23dedb2deff8f21b4315c1ed9ce/bids-validator/validators/bids/fullTest.js#L83-L90

Setting this to an error (as fMRIPrep does to detect missing T1w images before attempting to run) means all datasets will fail.

nellh:
This adds a similar field to the old modalities field called dataTypes that uses almost exactly the same logic as before and updates Issue 53 to check for T1w in this new field to fix #1307


bids-validator/validators/bids/fullTest.js
```diff
@@ -81,7 +81,7 @@ const fullTest = (fileList, options, annexed, dir, callback) => {
   })
 
   // check if dataset contains T1w
-  if (summary.modalities.indexOf('T1w') < 0) {
+  if (summary.dataTypes.includes('T1w')) {
```
effigies: I think this is inverted...

```suggestion
  if (!summary.dataTypes.includes('T1w')) {
```

nellh:
Added an actual test for Issue 53 as well.

PR #1311, opened 2021-07-28
Behavior schema and validation

Remi-Gau:
To keep in line with https://github.com/bids-standard/bids-specification/pull/804 to introduce some metadata for behavioral experiments this PR:
- creates a `beh.json` file
- makes sure it is used during validation

Question: none of the beh field is "required" so the json schema has no "required" key.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -157,6 +157,8 @@ You can get a good impression using the following links:
 - [How JSON schemas are tested](https://github.com/bids-standard/bids-validator/blob/master/bids-validator/tests/json.spec.js)
 - [How TSV files are tested](https://github.com/bids-standard/bids-validator/blob/master/bids-validator/tests/tsv.spec.js)
 
+For more information on how to run the tests check the [testing section](./README.md#testing) of the README.
+
 [link_git]: https://git-scm.com/
 [link_handbook]: https://guides.github.com/introduction/git-handbook/
 [link_swc_intro]: http://swcarpentry.github.io/git-novice/
--- a/bids-validator/tests/json.spec.js
+++ b/bids-validator/tests/json.spec.js
@@ -587,5 +587,4 @@ describe('JSON', function() {
       assert(issues[1].evidence == '.CogPOID should match format "uri"')
     })
   })
-
 })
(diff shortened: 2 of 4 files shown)


sappelhoff:
nice! Could you perhaps try to add a test here --> https://github.com/bids-standard/bids-validator/blob/master/bids-validator/tests/json.spec.js ?

Remi-Gau:
> nice! Could you perhaps try to add a test here --> https://github.com/bids-standard/bids-validator/blob/master/bids-validator/tests/json.spec.js ?

ok was actually thinking whether tests should be added somewhere.

that shall be my first javascript test. :scream:

sappelhoff (approved):
thanks @Remi-Gau 

(my commit above is supposed to fix the linter, let's see if it worked)

sappelhoff (commented):
Most of the test failures are due to HED (unrelated to this PR), but the test you added is also failing, see screenshot:

image


I usually edit the code and put a `console.log(...)` somewhere (... being whatever I want to print) and then just execute that single test that fails to debug. But perhaps you already have a better JS debugging workflow :-)

Remi-Gau:
> Most of the test failures are due to HED (unrelated to this PR), but the test you added is also failing, see screenshot:

I was getting a lot of failed tests when running locally as well.

> I usually edit the code and put a `console.log(...)` somewhere (... being whatever I want to print) and then just execute that single test that fails to debug. But perhaps you already have a better JS debugging workflow :-)

Hahaha. You assume I have a JS debugging workflow that goes beyond "assume foetal position and cry a lot": nice to see you have hope in me.

sappelhoff (approved):
Approving once more.

I fixed the bug in 8b7b588 --> "beh.json" can also occur *outside* of a `/beh/` subdirectory, which was also what the test "mocked", however the code didn't permit for that (now fixed).

PS: After our talk this morning, this thing didn't leave me in peace - I hope I didn't deprive you of a learning opportunity to fix it yourself :grimacing:

sappelhoff:
remaining failures are all due to HED to be fixed in #1313 I am merging this one now, thanks @Remi-Gau

Remi-Gau:
Cool!! Thanks.

PR #1312, opened 2021-07-29
make InstitutionalDepartmentName available for all datatypes

### LINKED ISSUE #1310: eeg, meg json schema include `InstitutionalDepartmentName` but the spec does not
see here:
- https://github.com/bids-standard/bids-validator/blob/39922c4159688fcc67ba6b723d391991c75211e3/bids-validator/validators/json/schemas/eeg.json#L10
- https://github.com/bids-standard/bids-validator/blob/39922c4159688fcc67ba6b723d391991c75211e3/bids-validator/validators/json/schemas/meg.json#L10

I think this is only used for MRI data in the spec:

https://github.com/bids-standard/bids-specification/blob/master/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md#institution-information

and this seems to keep in line with a dicom tag

https://github.com/bids-standard/bids-specification/blob/master/src/schema/metadata/InstitutionalDepartmentName.yaml

My gut feeling is to remove it from the json schema in the validator rather than modify the spec to align with the validator. But I may be missing something.

Remi-Gau:
fixes #1310 

# to do 

- [x] update beh.json schema to also contain InstitutionalDepartmentName

### DIFF
--- a/bids-validator/validators/json/schemas/beh.json
+++ b/bids-validator/validators/json/schemas/beh.json
@@ -7,7 +7,8 @@
     "CogAtlasID": { "$ref": "common_definitions.json#/definitions/CogAtlasID" },
     "CogPOID": { "$ref": "common_definitions.json#/definitions/CogPOID" },
     "InstitutionName": { "$ref": "common_definitions.json#/definitions/InstitutionName" },
-    "InstitutionAddress": { "$ref": "common_definitions.json#/definitions/InstitutionAddress" }
+    "InstitutionAddress": { "$ref": "common_definitions.json#/definitions/InstitutionAddress" },
+    "InstitutionalDepartmentName": { "$ref": "common_definitions.json#/definitions/InstitutionalDepartmentName" }
   },
   "additionalProperties": false
 }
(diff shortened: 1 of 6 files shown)


sappelhoff (commented):
:| apparently there is at least one dataset in BIDS-examples that makes use of this field: https://github.com/bids-standard/bids-examples/blob/89211d4afc5ee3effee3d7fd4b093e1eaf15837f/eeg_rishikesh/sub-001/ses-01/eeg/sub-001_ses-01_task-meditation_eeg.json#L4

So despite what I said earlier, maybe it's more elegant to add this metadata to EEG and MEG in the spec, rather than removing it from the validator and breaking existing datasets?

I feel like this metadata could be interesting for some datasets 🤷‍♂️  it's not completely irrelevant at least.

WDYT?

Remi-Gau:
Then I'd be of the opinion to make this a piece of metadata that we put in "common definitions" and that we make available for all data types.

This would mean updating the spec for iEEG, EEG, MEG, PET, behavior.

I care more mostly about internal consistency #DataOCD

Remi-Gau:
I would suggest to bring the spec up to speed before we merge this one.

Remi-Gau:
Will update the beh.json once #1311 is merged

Remi-Gau:
@sappelhoff 
Also updated the behavior schema so all the required changes should be there.

sappelhoff (approved):
wonderful, thanks!

PR #1313, opened 2021-07-29
Migrate placeholder validation to hed-validator

happy5214:
hed-validator supports proper value placeholder support as of version 3.4.0, which was released today. This PR deletes the remaining placeholder validation code left over in BIDS in favor of the more robust and correct hed-validator code, while splitting the categorical and value string-level validation loops to ensure that only value strings are allowed to have non-definition value placeholders.

Issue code 110, which deals with placeholder validation, has been deleted as redundant in favor of the generic HED error issue, as this problem is now reported by hed-validator instead of the BIDS code. This is the last major change in the HED-related BIDS issue codes that foresee at this time.

### DIFF
--- a/bids-validator/package-lock.json
+++ b/bids-validator/package-lock.json
@@ -5,7 +5,7 @@
   "requires": true,
   "packages": {
     "": {
-      "version": "1.7.3-dev.0",
+      "version": "1.8.1-dev.0",
       "license": "MIT",
       "dependencies": {
         "@aws-sdk/client-s3": "^3.9.0",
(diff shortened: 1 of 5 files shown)


sappelhoff (approved):

happy5214:
Right now, categorical string validation is broken (the current code tries to validate everything as a value string, which fails in hed-validator 3.4.0), which is fixed in this PR, so this needs to be merged ASAP.

happy5214:
Our plan over the next week or two (after this PR is merged) is to migrate most of the code in `hed.js` to hed-validator and make `hed.js` much leaner. That will also allow us to make changes without needing to bug you all as much. But as #1311 shows, the hed-validator update broke several tests (the existing `hed.js` code is buggy), so this needs to be merged ASAP.

sappelhoff:
Nobody else had a chance to look at it, but I am going to trust the CIs and my gut feeling here. Thanks @happy5214

rwblair:
@happy5214 apologies on not reviewing sooner. I just read through PR and it looks great. Will look at making a new release this week.

PR #1319, opened 2021-08-17
Fix hed-validator dependency

### LINKED ISSUE #1317: package-lock.json in root still uses old hed-validator version
The `package-lock.json` file in the root directory still contains instructions for installing version 3.0.5 of hed-validator. While it may be poor semver compliance on our part, the current code in bids-validator is no longer compatible with that version of hed-validator, so the web BIDS validator is effectively unusable with HED data. This is a must-fix, and I simply cannot figure out why the package manager will not put 3.4.0 (or any recent version for that matter) in the lock file.

happy5214:
This was fixed by "reinstalling" it using the `-w` flag of `npm install` from the root.

Fixes #1317

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "bids-validator",
       "workspaces": [
         "bids-validator",
         "bids-validator-web"
@@ -18,7 +19,7 @@
       }
     },
     "bids-validator": {
-      "version": "1.8.1-dev.0",
+      "version": "1.8.1",
       "license": "MIT",
       "dependencies": {
         "@aws-sdk/client-s3": "^3.9.0",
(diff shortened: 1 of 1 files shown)


rwblair:
@happy5214 do you need this PR issued in a new release?

happy5214:
> @happy5214 do you need this PR issued in a new release?

@rwblair Yes, please. The previous PR for this issue unfortunately didn't address the problem at all.

rwblair:
Sorry about that:
https://github.com/bids-standard/bids-validator/releases/tag/v1.8.2

VisLab:
I had to be the bearer of bad news but it doesn't look like the
bids-validator-web has been updated to 1.8.2 with the release.

rwblair:
https://github.com/bids-standard/bids-validator/tree/gh-pages

I'm seeing the v1.8.2 on https://bids-standard.github.io/bids-validator/   could be caching issue? or does the validator on the site not do what its supposed to?

VisLab:
You are right --- I had to reclone the repository, but now it worked......
Thanks for your quick response..... The datasets validate now!

PR #1320, opened 2021-08-24
Migrate bulk of HED code to hed-validator

happy5214:
This branch will migrate the majority of the HED-specific code to
the hed-validator codebase, thus freeing the BIDS code of the burden.
Only the minimum required to sustain the interface is kept, including
determining whether HED validation is needed, extracting the schema
version info, and interfacing with the other BIDS code. I will trim
down the testcases further before this PR is merged.

lodash was added as a direct dependency to add the union function.
It already existed as an indirect dependency.


happy5214:
The failing tests all relate to placeholder handling. If they are kept, they will need to be rewritten to use HED 3.

happy5214:
The failing tests previously mentioned above were removed and are now tested in hed-validator. Currently, the failing tests are from the `cli.js` module, and they appear to have nothing to do with the code I modified.

happy5214:
@rwblair I think this is ready to be merged. Is there anything holding this up?

PR #1324, opened 2021-09-10
fix: Avoid missing arguments when bids-validator script is called directly

nellh:
This was changed in bids-standard/legacy-validator#1294 to allow tests to call the options parser but it broke directly calling the script (`./bin/bids-validator`)

### DIFF
--- a/bids-validator/bin/bids-validator
+++ b/bids-validator/bin/bids-validator
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 try {
   const { default: cli } = require('bids-validator/cli')
-  cli().catch((code) => {
+  cli(process.argv.slice(2)).catch(code => {
     process.exit(code)
   })
 } catch (err) {


sappelhoff:
related to: bids-standard/bids-validator#51 

Could we perhaps add a test that checks that calling the bids-validator script directly works?

rwblair:
Is this the error that you were seeing generated when calling './bin/bids-validator'?
`Error: Cannot find module 'bids-validator/cli'`

I'm seeing that error when running './bin/bids-validator' in both this branch and master at the moment.

nellh:
> Is this the error that you were seeing generated when calling './bin/bids-validator'?
> `Error: Cannot find module 'bids-validator/cli'`
> 
> I'm seeing that error when running './bin/bids-validator' in both this branch and master at the moment.

That's because the entry point only runs the built version (when installed from npm or built locally with `npm run build`) so a separate issue. We should have a fallback for that but it's a separate issue.

rwblair:
Ah I see, current master is always returning:
`Not enough non-option arguments: got 0, need at least 1`

PR #1325, opened 2021-09-10
bids-specification schema validation for data types

nellh:
This adds an option to apply the bids-specification schema to some parts of validation. The top level schema is also loaded but not used for validation here. Based on @rwblair's start in #1126, this integrates a schema flag that takes a tag for the bids-specification repo and loads the yaml definitions (with a fallback to a bundled copy of v1.6.0 for when network access is unavailable and tests).

### DIFF
--- a/bids-validator/schema/datatypes/anat.yaml
+++ b/bids-validator/schema/datatypes/anat.yaml
@@ -1,0 +1,174 @@
+---
+# Nonparametric
+- suffixes:
+    - T1w
+    - T2w
+    - PDw
+    - T2starw
+    - FLAIR
+    - inplaneT1
+    - inplaneT2
+    - PDT2
+    - angio
+    - T2star  # deprecated
+    - FLASH  # deprecated
+    - PD  # deprecated
+  extensions:
+    - .nii.gz
+    - .nii
+    - .json
+  entities:
+    subject: required
+    session: optional
+    run: optional
+    acquisition: optional
+    ceagent: optional
+    reconstruction: optional
+    part: optional
+# Parametric
+- suffixes:
+    - T1map
+    - T2map
+    - T2starmap
+    - R1map
+    - R2map
+    - R2starmap
+    - PDmap
+    - MTRmap
+    - MTsat
+    - UNIT1
+    - T1rho
+    - MWFmap
+    - MTVmap
+    - PDT2map
+    - Chimap
+    - S0map
+    - M0map
+  extensions:
+    - .nii.gz
+    - .nii
+    - .json
+  entities:
+    subject: required
+    session: optional
+    run: optional
+    acquisition: optional
+    ceagent: optional
+    reconstruction: optional
+# Defacemask
+- suffixes:
+    - defacemask
+  extensions:
+    - .nii.gz
+    - .nii
+    - .json
+  entities:
+    subject: required
+    session: optional
+    run: optional
+    acquisition: optional
+    ceagent: optional
+    reconstruction: optional
+    modality: optional
+# Multi-echo
+- suffixes:
+    - MESE
+    - MEGRE
+  extensions:
+    - .nii.gz
+    - .nii
+    - .json
+  entities:
+    subject: required
+    session: optional
+    run: optional
+    acquisition: optional
+    ceagent: optional
+    reconstruction: optional
+    echo: required
+    part: optional
+# Multi-flip
+- suffixes:
+    - VFA
+  extensions:
+    - .nii.gz
+    - .nii
+    - .json
+  entities:
+    subject: required
+    session: optional
+    run: optional
+    acquisition: optional
+    ceagent: optional
+    reconstruction: optional
+    flip: required
+    part: optional
+# Multi-inv
+- suffixes:
+    - IRT1
+  extensions:
+    - .nii.gz
+    - .nii
+    - .json
+  entities:
+    subject: required
+    session: optional
+    run: optional
+    acquisition: optional
+    ceagent: optional
+    reconstruction: optional
+    inversion: required
+    part: optional
+# MP2RAGE
+- suffixes:
+    - MP2RAGE
+  extensions:
+    - .nii.gz
+    - .nii
+    - .json
+  entities:
+    subject: required
+    session: optional
+    run: optional
+    acquisition: optional
+    ceagent: optional
+    reconstruction: optional
+    echo: optional
+    flip: optional
+    inversion: required
+    part: optional
+# VFA+MT
+- suffixes:
+    - MPM
+    - MTS
+  extensions:
+    - .nii.gz
+    - .nii
+    - .json
+  entities:
+    subject: required
+    session: optional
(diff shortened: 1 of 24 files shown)


sappelhoff:
cc @tsalo

tsalo:
@sappelhoff Thanks! I wasn't aware of this PR.

bids-validator/schema/datatypes/anat.yaml
```diff
@@ -0,0 +1,174 @@
...
+    - .nii
+    - .json
+  entities:
+    subject: required
+    session: optional
+    run: optional
+    acquisition: optional
+    ceagent: optional
+    reconstruction: optional
+    modality: optional
```
effigies, line 72: Going by our current [regex](https://github.com/bids-standard/bids-validator/blob/fcc1c296c02d09c434aeea066b7d65e8836cc0c6/bids-validator/bids_validator/rules/file_level_rules.json#L49-L95), this should not have `ceagent`. Also, your regex construction seems to depend on the ordering of these entities, which may be why the test is failing.

```suggestion
    subject: required
    session: optional
    acquisition: optional
    reconstruction: optional
    run: optional
    modality: optional
```

tsalo: Makes sense. The entity order should be inferred from the `entities.yaml` file, rather than the datatype files.

nellh: Thanks, fixing the ordering issue pointed out another test issue with modality vs mod (not reading the entity key from `entities.yaml`). That gets the anat example working as expected.

The eeg example taken from the existing tests `/sub-05/eeg/sub-05_task-matchingpennies_recording-eyetracking_physio.tsv.gz` looks like it wouldn't match the schema because a physio suffix isn't specified for eeg, does that seem accurate or I am misinterpreting it?

rwblair: Physio is an odd one out, specification adds it too a bunch of modalities in its own section:
https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/06-physiological-and-other-continuous-recordings.html

Might explain how it could be overlooked when making/updating schemas.

effigies:
Do we know why the `test_docker` is [failing](https://app.circleci.com/pipelines/github/nellh/bids-validator/59/workflows/3348dc79-62e9-4b17-98ee-f5118b85df32/jobs/649)?

```
docker run --rm -it -v $PWD/bids-validator/tests/data/valid_headers:/data bids/validator:latest /data --ignoreNiftiHeaders
docker run --rm -it -v $PWD/bids-validator/tests/data/valid_headers:/data bids/validator:latest /data --ignoreNiftiHeaders --json

^@^@internal/modules/cjs/loader.js:796
    throw err;
    ^

Error: Cannot find module 'esbuild-runner/register'
Require stack:
- /src/bin/bids-validator
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:793:17)
    at Function.Module._load (internal/modules/cjs/loader.js:686:27)
    at Module.require (internal/modules/cjs/loader.js:848:19)
    at require (internal/modules/cjs/helpers.js:74:18)
    at Object.<anonymous> (/src/bin/bids-validator:14:5)
    at Module._compile (internal/modules/cjs/loader.js:955:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:991:10)
    at Module.load (internal/modules/cjs/loader.js:811:32)
    at Function.Module._load (internal/modules/cjs/loader.js:723:14)
    at Function.Module.runMain (internal/modules/cjs/loader.js:1043:10) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ '/src/bin/bids-validator' ]
}

Exited with code exit status 1
```

bids-validator/src/schemaTypes.js
```diff
@@ -0,0 +1,163 @@
...
+        const entityDefinion = schema.entities[entity]
+        if (entity in datatype.entities) {
+          // sub and ses entities in file name handled by directory pattern matching groups
+          if (entity === 'subject' || entity === 'session') {
+            continue
+          }
+          const entityKey = entityDefinion.entity
+          const format = regex[schema.entities[entity].format]
+          if (format) {
+            // Limitation here is that if format is missing an essential entity may be skipped
```
effigies, line 132: Does it make sense to default to alphanumeric (label) if format is unknown?

rwblair:
Have a proposed fix for docker issue here #1335 

Can't figure out why the gitlab tests there are failing.

effigies (dismissed):
LGTM. One question. I've also opened #1342 to make sure we're exercising all three schema modes on bids-examples.

bids-validator/schema/datatypes/meg.yaml
```diff
@@ -0,0 +1,85 @@
+---
+# First group
+- suffixes:
+    - meg
+  extensions:
+    - /  # corresponds to BTi/4D data
+    - .ds/
```
effigies, line 7: We're not handling directories correctly. We should be accepting any files within. The original regex was `"\\.ds[\\/\\\\].*",`

https://github.com/bids-standard/bids-validator/blob/b3746cb05dbe024711134ffe04332b5a06ebc7e5/bids-validator/bids_validator/rules/file_level_rules.json#L466-L475

bids-validator/schema/datatypes/meg.yaml
```diff
@@ -0,0 +1,85 @@
+---
+# First group
+- suffixes:
+    - meg
```
effigies, line 4: @tsalo Need to update the schema to accommodate two special files defined in https://bids-specification.readthedocs.io/en/v1.6.0/99-appendices/06-meg-file-formats.html#cross-talk-and-fine-calibration-files.

Corresponding regexes:

https://github.com/bids-standard/bids-validator/blob/b3746cb05dbe024711134ffe04332b5a06ebc7e5/bids-validator/bids_validator/rules/file_level_rules.json#L496-L502

bids-validator/schema/datatypes/ieeg.yaml
```diff
@@ -0,0 +1,68 @@
...
+    run: optional
+- suffixes:
+    - coordsystem
+  extensions:
+    - .json
+  entities:
+    subject: required
+    session: optional
+    acquisition: optional
+    space: optional
```
effigies, line 39: Seeing

```
	2: [ERR] A json sidecar file was found without a corresponding data file (code: 90 - SIDECAR_WITHOUT_DATAFILE)
		./sub-cm4/ieeg/sub-cm4_space-Pixels_coordsystem.json
```

I don't think this should happen, but I'm not clear on why it's happening for ieeg, but not MEG/EEG.

bids-validator/schema/datatypes/meg.yaml
tsalo: I think @Remi-Gau has a proposal for those in https://github.com/bids-standard/bids-specification/issues/810.

PR #1327, opened 2021-09-13
Security updates (Lerna 4.0, Next 11) --> run: npm audit fix --force

sappelhoff:
This is to address a security warning and includes two SemVer breaking changes (Lerna from 3 to 4 [dev dependency only], Next from 10 to 11 [web dependency only]).

done on an up-to-date branch after running `npm ci` (clean install), and then `npm audit fix --force`, all from the root of the repository (npm version 7.23.0, node version 12.22.6)

---

When the initial fix did not work, I continued with the following (list in progress):

- `npm install react@latest react-dom@latest` from the `bids-validator-web` subfolder, following the [NEXT upgrade guide](https://nextjs.org/docs/upgrading)
- Tried to opt out of the new default "webpack 5", and use webpack 4 instead (possible until Next version 12), see https://nextjs.org/docs/messages/webpack5

### DIFF
--- a/bids-validator-web/next.config.js
+++ b/bids-validator-web/next.config.js
@@ -4,6 +4,7 @@ const withTM = require("next-transpile-modules");
 module.exports = withTM({
   transpileModules: ["bids-validator"],
   assetPrefix: "./",
+  webpack5: false,
   webpack: (config, {}) => {
     config.resolve.symlinks = false;
     return config;
--- a/bids-validator-web/package.json
+++ b/bids-validator-web/package.json
@@ -7,7 +7,7 @@
   "dependencies": {
     "bootstrap": "^4.3.0",
     "bowser": "^1.0.0",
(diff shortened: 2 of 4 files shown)


nellh:
If webpack 5 is causing issues, it may work using the esbuild output, webpack 5 and vite (used by OpenNeuro) have the same issues with the validator and that's why the esbuild script was added.

sappelhoff:
Yes, webpack 5 was the problem. It's possible to use webpack 4 until NEXT 12, ... should we just go with that?

nellh:
> Yes, webpack 5 was the problem. It's possible to use webpack 4 until NEXT 12, ... should we just go with that?

Sure, it's possible dependencies will resolve the issues (it's caused by dependencies doing technically invalid things but webpack 4 covering for it) and that is the smoother path forward.

PR #1331, opened 2021-09-13
fix: Use an inline build when running bin/bids-validator from source

### LINKED ISSUE #1329: Fallback for cli if esbuild output not present.
> Is this the error that you were seeing generated when calling './bin/bids-validator'?
> `Error: Cannot find module 'bids-validator/cli'`
> 
> I'm seeing that error when running './bin/bids-validator' in both this branch and master at the moment.

That's because the entry point only runs the built version (when installed from npm or built locally with `npm run build`) so a separate issue. We should have a fallback for that but it's a separate issue.

_Originally posted by @nellh in https://github.com/bids-standard/bids-validator/issues/1324#issuecomment-918266959_

nellh:
This uses esbuild-runner to transpile everything on the fly while running ./bin/bids-validator in a development source tree. That way you can test without needing to rebuild but still get an accurate idea if things are likely to work when the esbuild version is published.

Fixes #1329

### DIFF
--- a/bids-validator/bin/bids-validator
+++ b/bids-validator/bin/bids-validator
@@ -1,9 +1,20 @@
 #!/usr/bin/env node
-try {
-  const { default: cli } = require('bids-validator/cli')
+function entry(cli) {
   cli(process.argv.slice(2)).catch(code => {
     process.exit(code)
   })
+}
+
+try {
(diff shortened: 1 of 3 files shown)


rwblair:
Wow this is amazing... thank you so much.

PR #1335, opened 2021-09-14
upgrade npm in docker file for workspace compatability

rwblair:


effigies (approved):
Gitlab tests are also failing on `master`, so this seems like an improvement.

Dockerfile
```diff
@@ -1,7 +1,10 @@
...
+FROM node:14-alpine
 
-COPY ./bids-validator /src
+COPY . /src
 
-RUN npm install -g /src
+RUN npm install -g npm
+WORKDIR /src
+RUN npm install 
+RUN npm install -g ./bids-validator
```
nellh, line 8: @rwblair  I'm late but I'd suggest:

```sh
npm install
npm -w bids-validator run build
npm -w bids-validator pack
npm install -g bids-validator-*.tgz
```

That way the build doesn't run every time you run the validator (though esbuild is pretty fast).

PR #1336, opened 2021-09-14
Upgrade versioneer, fix flake8 + other misc fixes

DimitriPapadopoulos:
https://lgtm.com/projects/g/bids-standard/bids-validator/?severity=error

Fix the first of two errors, the fix seems rather obvious.

The other error is in `_version.py` and has been fixed upstream in https://github.com/python-versioneer/python-versioneer/pull/250.


sappelhoff (commented):
for the `_version` related problem, we could update our vendored versioneer version: https://github.com/python-versioneer/python-versioneer#updating-versioneer

That should fix it, be backward compatible, and give us an up to date versioneer that is now being maintained by our very own @effigies

DimitriPapadopoulos:
CI fails with:
```
bids-validator/bids_validator/_version.py:339:80: E501 line too long (84 > 79 characters)
```
I don't want to modify the `_version.py` file generated by _versioneer_, I cannot find a _versioneer_ option to obtain a different `_version.py` file. The only option I can think of is to change the CI configuration to accept lines < 85 characters instead of lines < 80 characters. Or perhaps < 89 characters like black?

sappelhoff (commented):
you can setup a section on flake8 in our setup.cfg to ignore the versioneer files.

[Example how to do it](https://github.com/sappelhoff/pyprep/blob/04fbf2e51fdf0d05c27f234a114ed38e1cace1af/setup.cfg#L54-L69)

[Place where to do it here](https://github.com/bids-standard/bids-validator/blob/master/bids-validator/setup.cfg)

DimitriPapadopoulos:
Should be OK for flake8.

 Not sure how to fix that for [pytest](https://docs.pytest.org/en/latest/reference/customize.html#setup-cfg):
> `setup.cfg` files are general purpose configuration files, used originally by [distutils](https://docs.python.org/3/distutils/configfile.html), and can also be used to hold pytest configuration if they have a `[tool:pytest]` section.
> [...]
> ### Warning:
> Usage of `setup.cfg` is not recommended unless for very simple use cases. `.cfg` files use a different parser than `pytest.ini` and `tox.ini` which might cause hard to track down problems. When possible, it is recommended to use the latter files, or `pyproject.toml`, to hold your pytest configuration.

Is it OK to create a `pytest.ini` file?

DimitriPapadopoulos:
Actually not OK for flake8, because as far as I can see, it is run from `bids-validator`, not from `bids-validator/bids-validator`.

sappelhoff:
> Is it OK to create a pytest.ini file?

if you can figure out a minimal solution that involves us having to add a file, that'd still be fine I think.

But reusing existing config files would be better :) (if possible)

effigies:
Pytest can be configured in [setup.cfg](https://docs.pytest.org/en/6.2.x/customize.html#setup-cfg).

DimitriPapadopoulos:
I know, but the [pytest documentation](https://docs.pytest.org/en/latest/reference/customize.html#setup-cfg) warns:
> ### Warning:
> Usage of `setup.cfg` is not recommended unless for very simple use cases. `.cfg` files use a different parser than `pytest.ini` and `tox.ini` which might cause hard to track down problems. When possible, it is recommended to use the latter files, or `pyproject.toml`, to hold your pytest configuration.

Hopefully this is a very simple case :smiley:

DimitriPapadopoulos:
I had to add options directly to `flake8` and `pytest` because the config files seem broken. It appears you **have** to run `flake8` and probably `pytest` from the directory which contains the config file, which is not the case here. Fixing the config files is a considerable effort, beyond the scope of this PR.

DimitriPapadopoulos:
OK, all options have been added to `setup.cfg`.

Now I need to tell Code Climate not to bother with `versioneer.py` and `_version.py`.

DimitriPapadopoulos:
See #1341 for shutting up Cold Climate. Hopefully it'll work...

DimitriPapadopoulos:
Still these **codeclimate** errors. I just don't know how to exclude the _versioneer_ files. Any clues?

The **ci/circleci: test_docker** failure looks like a more general problem, unrelated to this PR.

sappelhoff:
I think code climate only complains on new changes. So after committing the versioneer files initially, it won't be a problem in the next prs. 

It's not as satisfying, but I don't know how else to fix it.

DimitriPapadopoulos:
I thought #1341 would help:
```
exclude_patterns:
  - "**/versioneer.py"
  - "**/_version.py"
  ```
It should according to the documentation, but doesn't :-(

sappelhoff:
In 754bf25 I tried whether the exclude patterns work when they are absolute (relative to the .codeclimate.yml file), instead of globs. That didn't work.

Now I am wondering whether `.codeclimate.yml` is perhaps ignored in general, and should be at the root of the REPO, not nested inside the bids-validator directory :thinking:

DimitriPapadopoulos:
> Now I am wondering whether `.codeclimate.yml` is perhaps ignored in general, and should be at the root of the REPO, not nested inside the bids-validator directory thinking

Makes sense. I'm moving it around in #1347.

DimitriPapadopoulos:
Eventually got CodeClimate to work properly by (I think) moving `bids-validator/.codeclimate.yml` to `.codeclimate.yml` (#1347).

effigies (approved):

sappelhoff (approved):
wonderful, thanks for sticking with this.

PR #1339, opened 2021-09-14
Fix LGTM.com Duplicate 'if' condition 

### LINKED ISSUE #1337: LGTM.com warning: Duplicate 'if' condition
Before 4cdd841 (#975), [`bids-validator/validators/nifti/nii.js`](https://github.com/bids-standard/bids-validator/blob/a9a9fd1/bids-validator/validators/nifti/nii.js#L1089-L1098) looks like this:
```js
      [...]
      } else if (mergedDictionary.RepetitionTime && header) {
      [...]
      } else if (mergedDictionary.RepetitionTime) {
      [...]
```

After 4cdd841 (#975), [`bids-validator/validators/nifti/nii.js`](https://github.com/bids-standard/bids-validator/blob/4cdd841/bids-validator/validators/nifti/nii.js#L1089-L1098) looks this:
```js
      [...]
      } else if (mergedDictionary.RepetitionTime && header) {
      [...]
      } else if (mergedDictionary.RepetitionTime && header) {    
      [...]
```
Note sure what the intent was, but now the 2nd `if` is dead code.

https://lgtm.com/projects/g/bids-standard/bids-validator/snapshot/a0d2ccd6c140c37b2f86c062b77056b066b84fcc/files/bids-validator/validators/nifti/nii.js?sort=name&dir=ASC&mode=heatmap#xe5a908768003ec35:1

rwblair:
fixes #1337, was an accident on my part, should of merged the two else ifs with matching conditions together.


bids-validator/validators/nifti/nii.js
```diff
@@ -1125,7 +1125,7 @@ export default function NIFTI(
             }),
           )
         }
-      } else if (mergedDictionary.RepetitionTime && header) {
+
         const niftiTR = Number(repetitionTime).toFixed(3)
```
effigies, line 1129: Should this be an `else` condition on the `if (repetitionUnit !== 's') {`? Or it's possible to have issues 11 and 12 simultaneously?

rwblair: It is technically possible to have 11 and 12 at the same time, one might be using the incorrect units and have the TR mismatch in the header and metadata. The error message is misleading for error 12 since it explicitly says "sec".

effigies: If `RepetitionTime` is 1 and `pixdim[4]` is 1000 and `xyzt_units` is `msec` I think we'd hit both, even though they are a match. But that probably also doesn't matter.

effigies:
Merged master in to fix gitlab test and then saw it's not fixed in master. Ah well.

Issue #1322, opened 2021-09-01
Newest Docker image in DockerHub?


pvelasco:
Hi,
Are there any plans on pushing the latest official Docker image to DockerHub?
The latest version [there](https://hub.docker.com/r/bids/validator/tags?page=1&ordering=last_updated) seems to be v1.8.0
Thanks!

sappelhoff:
true, that's weird because the images should be automatically pushed upon a release. cc @rwblair

pvelasco:
I think Docker now requires a subscription for [automated builds](https://docs.docker.com/docker-hub/builds/).

sappelhoff:
True 🤔 thanks! Perhaps we need to extend our circleci script to make a deployment. Something like here: https://github.com/BIDS-Apps/HCPPipelines/blob/4923bb3bb29543a0b74b9b5940b694fb1595d101/circle.yml#L29-L37

sappelhoff:
Seems to work now (thanks @pvelasco for starting and @rwblair for finishing up): https://hub.docker.com/r/bids/validator/tags?page=1&ordering=last_updated

@rwblair what kinds of images will be pushed now? For each PR on each commit? For each commit to master as "master" (not "latest"?)? For tags?

sappelhoff:
also, the docker action fails in #1351 :thinking:

sappelhoff:
okay, it seems to be working now

- https://hub.docker.com/r/bids/validator/tags
- and also on the GitHub container registry: https://github.com/bids-standard/bids-validator/pkgs/container/bids-validator

Thanks @rwblair

PR #1345, opened 2021-09-23
ENH: push docker image to DockerHub from CircleCI

pvelasco:
Addresses #1322 by deploying the docker image from CircleCI.


.circleci/config.yml
```diff
@@ -253,3 +277,13 @@ workflows:
               ignore: /.*/
             tags:
               only: /v\d+\.\d+\.\d+$/
+      - docker-deployment:
+          context: test
```
pvelasco, line 281: For this to work, there needs to be a **Context** in the CircleCI's **Organization Settings**.
When I did the testing, I called it `test`.  If named something else, it needs to be changed accordingly here.

PR #1348, opened 2021-09-28
ENH: push docker image to DockerHub from CircleCI 

rwblair:
Changes from @pvelasco (#1345), getting PR wholly within bids-standard org for circle testing/debugging.


rwblair:
@pvelasco I ended up using github actions over circle to do the build and push in this PR. circleci machine executor by default would store docker credentials in plaintext with docker login which is only an issue in so far as other people in the bids standard org could get at the containers via SSH. I couldn't figure out how to smoothly configure docker credential store to keep the credentials stored encrypted. github actions doesn't have ssh access on its builds. 

Still trying to get singularity image push to ghcr.io to work.

PR #1351, opened 2021-09-29
File check in tsv files to accept MEF3 data

MaxvandenBoom:
Hi @sappelhoff,

Long time no post from me, but I think I found an issue in the validator (command-line/web). When .tsv files have a filename column (for example in a scans.tsv file), then existence of that file will be checked. However, with MEF3 data that reference is a folder, causing the validator to produce a `SCANS_FILENAME_NOT_MATCH_DATASET `error.

Since only files are passed to the tsv.js function and not directories, I thought that the changes in this PR would provide an simple solution to the problem by adding the root of the MEF3 dataset (`_ieeg.mefd`) to the list variable that the filename is being checked against (in the same way other files and CTF or BTI data is being checked).

Hope this is an acceptable solution, let me know what you think!


sappelhoff (commented):
nice to hear from you @MaxvandenBoom, the fix looks good to me.

Why do you use `const fDir = fPath.substring(0, fPath.indexOf('_ieeg.mefd/') + 10);` instead of the `const fDir = path.dirname(fPath)` that is being used for CTF and BTI data?

Also, could you please add a test?

MaxvandenBoom:
Unfortunately that doesn't work because the files in MEF3 are nested into a segment subfolder. So for example when fPath is
`/sub-01/ses-ieeg01/ieeg/sub-01_ses-ieeg01_task-finger_run-1234_ieeg.mefd/RG35.timd/RG35-000000.segd/RG35-000000.tmet`
then fDir, using the same line as for CTF and BTI data, would become:
`/sub-01/ses-ieeg01/ieeg/sub-01_ses-ieeg01_task-finger_run-1234_ieeg.mefd/RG35.timd/RG35-000000.segd`
while with the code I proposed it becomes:
`/sub-01/ses-ieeg01/ieeg/sub-01_ses-ieeg01_task-finger_run-1234_ieeg.mefd`


Not sure whether you want me to make a custom test in the bids-validator, or just provide you with a dataset?
I made a dataset available here: https://surfdrive.surf.nl/files/index.php/s/fOJSVioAeR6BwzC
This is the same that is already on OSF but includes `sub-01/ses-ieeg01/sub-01_ses-ieeg01_scans.tsv` on which the validator breaks

sappelhoff:
> Not sure whether you want me to make a custom test in the bids-validator, or just provide you with a dataset?

I had thought about a short custom test within bids-validator, somewhere here: https://github.com/bids-standard/bids-validator/blob/master/bids-validator/tests/tsv.spec.js

But to be honest, I don't know where it'd be smart to add it. Perhaps it suffices to have your personal tests for now; these changes are not invasive and/or impacting anything apart from MEF.

sappelhoff (approved):

MaxvandenBoom:
Thanks stefan!

Do you think it might be an idea to add a small MEF3 dataset to the test data? That way it would be easy to add standard tests like the ones that are already being done for other file formats (for example in tsv.spec.js).

I mean, MEF3 is already in the BIDS specification, so then it should have and pass standard tests in the BIDS-validator, I think.
I'm ok with supplying a dataset and writing a test as a pull request. 
Also @dorahermes what are your thoughts on this?

sappelhoff:
I think that'd be great!

dorahermes:
+1

sappelhoff:
@MaxvandenBoom can you merge `master` into your branch and add that small MEF3 testing data set to increase coverage for that modality a bit?

Or do you want to add the test data in a separate PR and have this merged as a minor fix first?

MaxvandenBoom:
@sappelhoff I think I prefer to have this as a minor fix and add the test data in a separate PR, if you don't mind.
I merged `master` into my fork, so it's good to go.

sappelhoff:
Thanks!

PR #1357, opened 2021-10-06
Only push to dockerhub on new tag on master.

rwblair:
Previous condition was too permissive, polluting dockerhub with PR builds.


.github/workflows/docker-build-push.yml
```diff
@@ -5,10 +5,7 @@ on:
     branches:
       - 'master'
```
effigies: I think this is an OR not an AND, though I'm not positive.

```suggestion
```

PR #1359, opened 2021-10-07
[ENH] BEP031: Microscopy

mariehbourget:
This PR aims to incorporate Microscopy to the bids-validator (BEP031, @mariehbourget, @jcohenadad)

We have encountered some issues with the schema and with an external library and would appreciate some feedback and guidance before we continue.

### Questions about schema
We had a working version of the microscopy validator in commit 5d0a27e before we merged recent changes from master (commit f2588ed). If we understand correctly, the bids-validator now uses the `yaml` files from the schema to validate the datasets. We will make the appropriate changes related to that in order to have a working version again but we have some questions first:

1. Is the validation through `yaml` schema files mandatory or should we still be able to run it like the previous version?
2. When using `yaml` files, are those files duplicated from the` bids-specification/src/schema`?
3.  If yes, is there any sort of "synchronisation" between bids-spec and bids-validator or should we update them in both repos?

Note: Some `yaml` files were not included for Microscopy in bids-spec to facilitate community review. If we have to add them to bids-validator, I would like to add them as well in bids-spec for consistency.

### Questions about external libraries
For microscopy, we need to read metadata from the XML header of OME-TIFF files. We are able to do that via the `exifreader`, `esm` and `xml2js` libraries.
However, when installing `exifreader` with `npm`, the node_module `bids-validator` is automatically uninstalled.
We think we might not do the installation correctly or there might be a security issue with `xmldom` (a dependance of `exifreader`). Running `npm audit` gives us this message:

npm_audit

As we are not very experienced with javascript, we would need help with integrating and installing external librairies with the bids-validator. Thank you in advance!


mariehbourget:
Hi @bids-standard/maintainers,

We are a bit stuck on the validator for Microscopy because of recent changes where the validator uses the `bids_schema` variable.

As the schema for Microscopy will not be completed on the bids-spec until the community review is done, is there a way to bypass the `bids_schema` in the validator so it works as it was before? (Details are my first comment above)

Thank you in advance for your help!

nellh:
> Hi @bids-standard/maintainers,
> 
> We are a bit stuck on the validator for Microscopy because of recent changes where the validator uses the `bids_schema` variable.
> 
> As the schema for Microscopy will not be completed on the bids-spec until the community review is done, is there a way to bypass the `bids_schema` in the validator so it works as it was before? (Details are my first comment above)
> 
> Thank you in advance for your help!

Schema based validation is optional and disabled by default at the moment. The goal is to use it in more of the validator in the future but at the moment it shouldn't block adding support for a new BEP. Right now the copy of the schema in the validator is  mainly for the test suite and changes should be made in the bids-specification repository only.

For the external libraries, you'll want to make sure they are added to the bids-validator package rather than the top level of the repo. You can do that with `npm install -w bids-validator exifreader` from the root (make sure you have npm 7 or later). That will add it to that package while considering the dependency tree for all packages in the monorepo.

The security concern with xmldom looks like it is [already reported upstream](https://github.com/mattiasw/ExifReader/issues/145) and affects features of the library that are not used in exifreader, so that shouldn't affect the use here.

etiennebergeron:
Hi @nellh,

Thank you for your indications about the schema based validation. Unfortunately, it seems that it is still interfering with the microscopy JSON schema and rules added through this BEP. I'm using the command-line version of the bids-validator on Windows.

Since we merged the master branch with our bep031 (commit f2588ed), the validator doesn't recognize anymore the microscopy files, even though the `file_level_rules.json` didn't change. Here's the result I get when I validate the `dataset microscopy_SEM001` ([bids-examples PR #286](https://github.com/bids-standard/bids-examples/pull/286)), which was BIDS-compatible before our merging :

image

I tried to declare `let bids_schema = false` and to comment the lines 23 to 25 in `type.js`, but the same result is observed.

Thanks again for your help,


Etienne

nellh:
Okay, I've gave it a quick try and was able to validate the bids-examples microscopy datasets using this branch as expected.

Do you have a `dist` directory in the repo root or the bids-validator package? You might need to remove that directory if it exists, it could have a old build of the validator if you've used the build or publish scripts.

The other idea to try is see if you have another version of bids-validator in any node_modules in your node resolution path. Looking at it, the implementation as it is will use an installed module ahead of the git tree if both are present and we might want to invert that.

etiennebergeron:
I'm currently having some issues with the bids-validator installation. As I deleted the folder `dist` that was in the repo, I couldn't run the bids-validator command anymore, saying `Error: Cannot find module 'esbuild-runner/register'`, so I think I was indeed using an old build of the validator.

With that said, I tried to reinstall the validator with the commands `npm install --global npm@^7` and `npm install -g bids-validator`, but I didn't seem to work; the folder bids-validator under `C:\Users\<my_user>\AppData\Roaming\npm\node_modules` disappeared, a new `dist` folder was generated and some warnings are raised in the prompt:

image

So, is there a way to completely remove the validator from my computer to then reinstall it?

Thanks again!

nellh:
To test this branch, you need a local node_modules with the dependencies instead of the global install of the validator you get with `npm install -g bids-validator`. The `-g bids-validator` option fetches the package from npm and installs it into a shared directory with any other globally installed packages.

`npm install` from the repo root should be all you need to fetch the missing esbuild-runner package. That will create or synchronize a node_modules directory at the repo root. Then if you are running this branch with `./bids-validator/bin/bids-validator` it will look in there for dependencies.

etiennebergeron:
Hi again,

Since the bids-validator folder in `AppData\Roaming\npm\node_modules` was pretty much empty after I did `npm install -g bids-validator`, I pasted in what's in the Github repo. From there I ran `npm install`, but it didn't work (here's a part of what was in the prompt):

image

I am not sure to understand what that all means... It was working really well previously, at some point I had to delete manually the bids-validator files and reinstalled the validator; this is why I was wondering if there was a way to remove the bids-validator files and reinstall it for a fresh start.

Thanks for the fast replies :)

nellh:
node_modules/bids-validator would be the module installed with `npm install -g bids-validator`, which is fine for installing a release but not for testing a development branch. You could remove the entire node_modules directory to clear things out.

Clone the git branch and run `npm install` at the root to obtain the dependencies. Then run it with `./bids-validator/bin/bids-validator`. If that's working for you, there's [instructions for making the development version accessible as `bids-validator` by adding it to your path here](https://github.com/bids-standard/bids-validator/blob/master/CONTRIBUTING.md#using-the-development-version-of-bids-validator).

etiennebergeron:
I see, this make more sense for a development branch.

When I did `npm install` at the root cloned repo, it seemed to work, but I had couple of high and moderate vulnerabilities, suggesting me to do `npm audit fix --force`. As I type in that, the prompt crashes everytime...

At the end, if I try the validator with an example dataset, I get this:
image

Looks like it is still seeking for the `bids-validator` in `node_modules`... Do you have any idea what could be the problem? Thanks a lot for your patience, it helps us a lot!

nellh:
That makes sense about the npm audit warning. Until upstream fixes them, exifreader will cause npm install to warn about that. You can ignore the warning for now, the ones caused the changes in this PR do not affect this use case and the others are unrelated.

Can you post the full command you're using to run the validator with an example dataset when you get that error?

etiennebergeron:
Alright, thanks about the npm audit warning.
And sure, here it is:
image

nellh:
If you try `./bids-validator/bin/bids-validator` instead of `bids-validator` does that work for you?

etiennebergeron:
Unfortunately, I got the same result (if my syntax is correct):

image

etiennebergeron:
Hi @nellh ,

Just wondering if you saw my last comment, still not sure what I could do to make the validator work properly again.

Thanks again for your patience!

nellh:
Here's what I did to test this branch on Windows with PowerShell and it seems to work:

```
C:\Users\nell>git clone https://github.com/neuropoly/bids-validator.git
Cloning into 'bids-validator'...
remote: Enumerating objects: 17703, done.
remote: Counting objects: 100% (1813/1813), done.
remote: Compressing objects: 100% (789/789), done.
remote: Total 17703 (delta 1208), reused 1466 (delta 959), pack-reused 15890 eceiving objects: 100% (17703/17703), 8.21 Receiving objects: 100% (17703/17703), 13.84 MiB | 18.03 MiB/s, done.

Resolving deltas: 100% (12663/12663), done.

C:\Users\nell>cd bids-validator

C:\Users\nell\bids-validator>npm install -g npm@7

removed 222 packages, changed 3 packages, and audited 37 packages in 13s

1 moderate severity vulnerability

To address all issues, run:
  npm audit fix

Run `npm audit` for details.

C:\Users\nell\bids-validator>npm install

up to date, audited 1904 packages in 11s

102 packages are looking for funding
  run `npm fund` for details

38 vulnerabilities (7 moderate, 31 high)

To address issues that do not require attention, run:
  npm audit fix

To address all issues (including breaking changes), run:
  npm audit fix --force

Run `npm audit` for details.

C:\Users\nell\bids-validator>node bids-validator/bin/bids-validator
Usage: bids-validator <dataset_directory> [options]

Options:
  -h, --help                      Show help                            [boolean]
  -v, --version                   Show version number                  [boolean]
      --ignoreWarnings            Disregard non-critical issues        [boolean]
      --ignoreNiftiHeaders        Disregard NIfTI header content during
                                  validation                           [boolean]
      --ignoreSubjectConsistency  Skip checking that any given file for one
                                  subject is present for all other subjects.
                                                                       [boolean]
      --verbose                   Log more extensive information about issues
                                                                       [boolean]
      --json                      Output results as JSON               [boolean]
      --no-color                  Disable colors in output text.
                                                      [boolean] [default: false]
      --ignoreSymlinks            Skip any symlinked directories when validating
                                  a dataset                            [boolean]
      --remoteFiles               Validate remote files.
                                                      [boolean] [default: false]
      --gitTreeMode               Improve performance using git metadata. Does
                                  not capture changes not known to git.[boolean]
      --gitRef                    Targets files at a given branch, tag, or
                                  commit hash. Use with --gitTreeMode.
                                  [default: "HEAD"]                     [string]
  -c, --config                    Optional configuration file. See https://githu
                                  b.com/bids-standard/bids-validator for more
                                  info  [default: ".bids-validator-config.json"]
  -s, --schema                    BIDS specification schema version to use for
                                  validation, e.g. "v1.6.0" (beta)
                   [choices: "disable", "v1.6.0", "master"] [default: "disable"]

This tool checks if a dataset in a given directory is compatible with the Brain
Imaging Data Structure specification. To learn more about Brain Imaging Data
Structure visit http://bids.neuroimaging.io

Not enough non-option arguments: got 0, need at least 1

C:\Users\nell\bids-validator>node -v
v14.18.1

C:\Users\nell\bids-validator>npm -v
7.24.2
```

npm@7 is important though, I tried with npm 8 and ran into issues.

etiennebergeron:
Hi @nellh ,

Following the steps you did to test the branch, I recloned and the repo and the npm installation seems to work:

image


I then tried to run the validator, but here's what I have:

image

I think I'm closer to a working validator, let me know if you have an idea on what's wrong here :)

mattiasw:
Hi! Maintainer of ExifReader here. For what it's worth, version 4.0.0 of ExifReader includes an upgraded xmldom package so the audit warning should now be gone.

etiennebergeron:
Hi again @nellh ,

I was wondering if you had time to look at what I got following your steps.

And perfect @mattiasw , once my installation is fixed I'll be able to continue to work with ExifReader, thank you!

Best,

Etienne

rwblair:
@etiennebergeron very strange. I fired up a windows virtual machine and did a fresh clone/install of your validator branch and examples, and was able to run the command that you pictured above against SEM001 successfully. Even tried running under cmd to see if there was any difference between it and powershell, as well as double checking on my host machine and got the same results.

BidsSideCar has been a constructor in the hed validator for a while now, package.json has latest version of that validator in it, so its hard to imagine there being a caching issue or wonky interplay between global and local modules.

etiennebergeron:
Hi @rwblair,

Thank you for your comment. As you did, I installed a Windows 10 Virtual Machine on my computer for a fresh installation. Fortunately, it works just fine on the VM, so I'll continue the tests for the BEP031 on it. I have no clue why it's bugged on my real PC and how to eradicate all the bids-validator files to then reinstall the validator, but my problem is solved.

Thanks again and have a great day!

Etienne

bids-validator/validators/tsv/tsv.js
```diff
@@ -240,6 +240,174 @@ const TSV = (file, contents, fileList, callback) => {
...
+        }
+        samples.push(sample)
+      }
+
+      // check if a sample froma same subject is described by one and only one row
+      var doesSampleIdHaveDuplicates = sampleIdColumnValues.some(
+        (val, i) => sampleIdColumnValues.indexOf(val) !== i
+      )
+
+      for (let r = 0; r < rows.length-1; r++) {
```
rwblair: From how I read this doesSampleIdHaveDuplicates when set in this loop or as set on line 348 could be overwritten many times, if we want to use a flag it should start false, and we break when its true and immediately generate an error. These samples files aren't too big so I'm not worried about nested loops for speed sake, but to reduce complexity of the flag logic we could try this:

First we make an array of strings combining the sample and participant ids into one string, we know this string must be unique:
`samplePartIds = row.map(x => x[sampleIdColumnValue] + x[participantIdColumnValue])`

Then we can generate a set from this which will deduplicate  it:
`samplePartIdsSet = new Set(samplePartIds)`

Finally if the array of ids we made is the same size as the set, we know there were no duplicates, but if the set is smaller we know it removed a duplicate:
`hasDuplicates = !(samplePartIdsSet.size === samplePartIds.length)`

If we wanted to find the actual duplicates we could skip the Set but still use the array of strings, iterate through it and update a dictionary using the sample-participant combo we see as a key, and its current count as the value. Then raise issues for each sample-participant key  with count greater than 1 in the dictionary after we're done iterating through the array.

bids-validator/validators/tsv/tsv.js
```diff
@@ -240,6 +240,174 @@ const TSV = (file, contents, fileList, callback) => {
...
+      const row = rows[c]
+      if ((row[sampleTypeColumn] != 'cell line') &&
+          (row[sampleTypeColumn] != 'in vitro differentiated cells') &&
+          (row[sampleTypeColumn] != 'primary cell') &&
+          (row[sampleTypeColumn] != 'cell-free sample') &&
+          (row[sampleTypeColumn] != 'cloning host') &&
+          (row[sampleTypeColumn] != 'tissue') &&
+          (row[sampleTypeColumn] != 'whole organisms') &&
+          (row[sampleTypeColumn] != 'organoid') &&
+          (row[sampleTypeColumn] != 'technical sample')) {
```
rwblair: Could simplify if statement along these lines:
```
const validSampleTypes = ['cell line', 'in vitro differentiated cells', 'primary cell', 'cell-free sample', 'cloning host', 'tissue', 'whole organisms', 'organoid', 'technical sample']

if (validSampleTypes.indexOf(row[sampleTypeColumn]) === -1) {
```

bids-validator/bids_validator/rules/file_level_rules.json
```diff
@@ -523,5 +523,60 @@
     "tokens": {
       "@@@_pet_ext_@@@": ["blood\\.tsv\\.gz", "blood\\.tsv", "blood\\.json"]
     }
+  },
+
+  "microscopy": {
+    "regexp": "^[\\/\\\\](sub-[a-zA-Z0-9]+)[\\/\\\\](?:(ses-[a-zA-Z0-9]+)[\\/\\\\])?microscopy[\\/\\\\](sub-[a-zA-Z0-9]+)(?:(_ses-[a-zA-Z0-9]+))?(?:_sample-[a-zA-Z0-9]+)(?:_acq-[a-zA-Z0-9]+)?(?:_stain-[a-zA-Z0-9]+)?(?:_run-[0-9]+)?(?:_chunk-[a-zA-Z0-9]+)?(((@@@_microscopy_type_@@@)(@@@_microscopy_ext_@@@))|(_photo\\.jpg))$",
```
rwblair: looks like spec has updated to directory to be `micr` instead of `microscopy`

The _photo regex may need to be separated out into its own entry based on what is currently in the `Photos of the samples (*_photo.<extension>)`  since they can't use chunk or stain entities.

mariehbourget:
Thanks @rwblair for the review!

It has been a while since we updated the PR, there are still a few missing pieces and we did not address all the changes following the community review of the spec yet.
We are actively working to bring this PR up-to-date and we will address your suggestions in the process, thanks again!

Here is a list of the remaining items that we will implement shortly:
- [x] Check for consistency for PixelSize between JSON and OME-TIFF metadata
- [x] Check for consistency with optional OME-TIFF metadata if present for `Immersion`, `NumericalAperture` and `Magnification`
- [x] `ChunkTransformationMatrix` was `OPTIONAL` and is now `RECOMMENDED` if `<chunk-index>` is used in filenames
- [x] `BodyPartDetails` was `OPTIONAL` and is now `RECOMMENDED`
- [x] `BodyPartDetailsOntology` is a `string` but must also be a valid `URI`
- [x] `SampleEnvironement` accepted values MUST be: `“in vivo”`, `“ex vivo”` or `“in vitro”`
- [x] `_photo` suffix extension MUST be either `.jpg`, `.png` or `.tif`
- [x] the datatype `microscopy` is now changed to `micr`

mariehbourget:
Update:
The implementations from the TODO list above are completed.
We are now working on addressing the comments from the previous review.

bids-validator/bids_validator/rules/file_level_rules.json
```diff
@@ -523,5 +523,70 @@
     "tokens": {
       "@@@_pet_ext_@@@": ["blood\\.tsv\\.gz", "blood\\.tsv", "blood\\.json"]
     }
+  },
+
+  "microscopy": {
+    "regexp": "^[\\/\\\\](sub-[a-zA-Z0-9]+)[\\/\\\\](?:(ses-[a-zA-Z0-9]+)[\\/\\\\])?micr[\\/\\\\](sub-[a-zA-Z0-9]+)(?:(_ses-[a-zA-Z0-9]+))?(?:_sample-[a-zA-Z0-9]+)(?:_acq-[a-zA-Z0-9]+)?(?:_stain-[a-zA-Z0-9]+)?(?:_run-[0-9]+)?(?:_chunk-[a-zA-Z0-9]+)?((@@@_microscopy_type_@@@)(@@@_microscopy_ext_@@@))$",
```
mariehbourget:
```suggestion
    "regexp": "^[\\/\\\\](sub-[a-zA-Z0-9]+)[\\/\\\\](?:(ses-[a-zA-Z0-9]+)[\\/\\\\])?micr[\\/\\\\](sub-[a-zA-Z0-9]+)(?:(_ses-[a-zA-Z0-9]+))?(?:_sample-[a-zA-Z0-9]+)(?:_acq-[a-zA-Z0-9]+)?(?:_stain-[a-zA-Z0-9]+)?(?:_run-[0-9]+)?(?:_chunk-[0-9]+)?((@@@_microscopy_type_@@@)(@@@_microscopy_ext_@@@))$",
```
In `chunk-<index>`, the `<index>` should be numerical only (non-negative integer like "run") and not alphanumeric value.

bids-validator/bids_validator/rules/file_level_rules.json
```diff
@@ -523,5 +523,70 @@
...
+    "tokens":{
+      "@@@_photo_ext_@@@": [
+        "_photo\\.jpg",
+        "_photo\\.png",
+        "_photo\\.tif"
+      ]
+    }
+  },
+  "microscopy_json": {
+    "regexp": "^[\\/\\\\](sub-[a-zA-Z0-9]+)[\\/\\\\](?:(ses-[a-zA-Z0-9]+)[\\/\\\\])?micr[\\/\\\\](sub-[a-zA-Z0-9]+)(?:(_ses-[a-zA-Z0-9]+))?(?:_sample-[a-zA-Z0-9]+)?(?:_acq-[a-zA-Z0-9]+)?(?:_stain-[a-zA-Z0-9]+)?(?:_run-[0-9]+)?(?:_chunk-[a-zA-Z0-9]+)?(@@@_microscopy_type_@@@)\\.json$",
```
mariehbourget:
```suggestion
    "regexp": "^[\\/\\\\](sub-[a-zA-Z0-9]+)[\\/\\\\](?:(ses-[a-zA-Z0-9]+)[\\/\\\\])?micr[\\/\\\\](sub-[a-zA-Z0-9]+)(?:(_ses-[a-zA-Z0-9]+))?(?:_sample-[a-zA-Z0-9]+)?(?:_acq-[a-zA-Z0-9]+)?(?:_stain-[a-zA-Z0-9]+)?(?:_run-[0-9]+)?(?:_chunk-[0-9]+)?(@@@_microscopy_type_@@@)\\.json$",
```
In `chunk-<index>`, the `<index>` should be numerical only (non-negative integer like "run") and not alphanumeric value.

bids-validator/bids_validator/rules/session_level_rules.json
```diff
@@ -201,5 +201,30 @@
         "UNCInfant2V23"
       ]
     }
+  },
+
+  "microscopy_ses": {
+    "regexp": "^[\\/\\\\](sub-[a-zA-Z0-9]+)[\\/\\\\](?:(ses-[a-zA-Z0-9]+)[\\/\\\\])?\\1(_\\2)?(?:_sample-[a-zA-Z0-9]+)?(?:_acq-[a-zA-Z0-9]+)?(?:_stain-[a-zA-Z0-9]+)?(?:_run-[0-9]+_)?(?:_chunk-[a-zA-Z0-9]+)?(@@@_microscopy_ses_type_@@@)$",
```
mariehbourget:
```suggestion
    "regexp": "^[\\/\\\\](sub-[a-zA-Z0-9]+)[\\/\\\\](?:(ses-[a-zA-Z0-9]+)[\\/\\\\])?\\1(_\\2)?(?:_sample-[a-zA-Z0-9]+)?(?:_acq-[a-zA-Z0-9]+)?(?:_stain-[a-zA-Z0-9]+)?(?:_run-[0-9]+_)?(?:_chunk-[0-9]+)?(@@@_microscopy_ses_type_@@@)$",
```
In `chunk-<index>`, the `<index>` should be numerical only (non-negative integer like "run") and not alphanumeric value.

bids-validator/bids_validator/rules/top_level_rules.json
```diff
@@ -133,5 +135,30 @@
     "tokens": {
       "@@@_other_top_files_ext_@@@": ["physio\\.json", "stim\\.json"]
     }
+  },
+
+  "microscopy_top": {
+    "regexp": "^[\\/\\\\](?:ses-[a-zA-Z0-9]+_)?(?:_sample-[a-zA-Z0-9]+)?(?:_acq-[a-zA-Z0-9]+)?(?:_stain-[a-zA-Z0-9]+)?(?:_run-[0-9]+)?(?:_chunk-[a-zA-Z0-9]+)?(?:@@@_microscopy_top_ext_@@@)$",
```
mariehbourget:
```suggestion
    "regexp": "^[\\/\\\\](?:ses-[a-zA-Z0-9]+_)?(?:_sample-[a-zA-Z0-9]+)?(?:_acq-[a-zA-Z0-9]+)?(?:_stain-[a-zA-Z0-9]+)?(?:_run-[0-9]+)?(?:_chunk-[0-9]+)?(?:@@@_microscopy_top_ext_@@@)$",
```
In `chunk-<index>`, the `<index>` should be numerical only (non-negative integer like "run") and not alphanumeric value.

bids-validator/validators/json/schemas/microscopy.json
```diff
@@ -0,0 +1,78 @@
...
+        { "type": "array", "items": { "type": "string" } }
+      ]
+    },
+    "SampleSecondaryAntibody": { 
+      "anyOf": [
+        { "type": "string", "minLength": 1 },
+        { "type": "array", "items": { "type": "string" } }
+      ]
+    },
+    "SliceThickness": { "type": "number" },
```
mariehbourget:
```suggestion
    "SliceThickness": { "type": "number", "exclusiveMinimum": 0 },
```

bids-validator/validators/json/schemas/microscopy.json
```diff
@@ -0,0 +1,78 @@
...
+    "SampleExtractionProtocol": { "type": "string", "minLength": 1 },
+    "SampleExtractionInstitution": { "type": "string", "minLength": 1 },
+    "TissueDeformationScaling": { "type": "number" },
```
mariehbourget, line 39:
```suggestion
    "TissueDeformationScaling": { "type": "number", "exclusiveMinimum": 0 },
```

bids-validator/validators/json/schemas/microscopy.json
```diff
@@ -0,0 +1,78 @@
...
+    "PixelSize": {"type": "array", "items": { "type": "number" } },
```
mariehbourget:
```suggestion
    "PixelSize": {"type": "array", "minItems": 2, "maxItems": 3, "items": { "type": "number", "minimum": 0 } },
```

bids-validator/validators/json/schemas/microscopy.json
```diff
@@ -0,0 +1,78 @@
...
+    "PixelSizeUnits": { "type": "string", "enum": ["mm", "um", "nm"] },
+    "Immersion": { "type": "string", "minLength": 1 },
+    "NumericalAperture": { "type": "number" },
```
mariehbourget:
```suggestion
    "NumericalAperture": { "type": "number", "exclusiveMinimum": 0 },
```

bids-validator/validators/json/schemas/microscopy.json
```diff
@@ -0,0 +1,78 @@
...
+    "Magnification": { "type": "number" },
```
mariehbourget:
```suggestion
    "Magnification": { "type": "number", "exclusiveMinimum": 0 },
```

bids-validator/validators/json/schemas/microscopy.json
```diff
@@ -0,0 +1,78 @@
...
+    "ImageAcquisitionProtocol": { "type": "string", "minLength": 1 },
+    "OtherAcquisitionParameters": { "type": "string", "minLength": 1 },
+    "ChunkTransformationMatrix": {"type": "array", "items": { "type": "array", "items":{ "type": "number"} } },
+    "ChunkTransformationMatrixAxis": { "type": "array", "items": { "type": "string" } }
```
mariehbourget:
```suggestion
    "ChunkTransformationMatrixAxis": { "type": "array", "minItems": 2, "maxItems": 3, "items": { "type": "string" } }
```

bids-validator/validators/json/schemas/microscopy.json
```diff
@@ -0,0 +1,78 @@
...
+    "SampleExtractionInstitution": { "type": "string", "minLength": 1 },
+    "TissueDeformationScaling": { "type": "number" },
+    "PixelSize": {"type": "array", "items": { "type": "number" } },
+    "PixelSizeUnits": { "type": "string", "enum": ["mm", "um", "nm"] },
+    "Immersion": { "type": "string", "minLength": 1 },
+    "NumericalAperture": { "type": "number" },
+    "Magnification": { "type": "number" },
+    "ImageAcquisitionProtocol": { "type": "string", "minLength": 1 },
+    "OtherAcquisitionParameters": { "type": "string", "minLength": 1 },
+    "ChunkTransformationMatrix": {"type": "array", "items": { "type": "array", "items":{ "type": "number"} } },
```
mariehbourget:
```suggestion
    "ChunkTransformationMatrix": {
      "anyOf": [
        {"type": "array", "minItems": 3, "maxItems": 3, "items": { "type": "array", "minItems": 3, "maxItems": 3, "items":{ "type": "number"} } },
        {"type": "array", "minItems": 4, "maxItems": 4, "items": { "type": "array", "minItems": 4, "maxItems": 4, "items":{ "type": "number"} } }
      ]
    },
```
To make sure the matrix is either 3x3 or 4x4. I'm unsure about the syntax.

mariehbourget:
Hi @bids-standard/maintainers,
I have an observation about the warning code "[38](https://github.com/bids-standard/bids-validator/blob/3f406e3f2ba328b08fb52809ca554414ecbdb1b3/bids-validator/utils/issues/list.js#L207)" .
It says:
```
  38: {
    key: 'INCONSISTENT_SUBJECTS',
    severity: 'warning',
    reason:
      'Not all subjects contain the same files. Each subject should contain the same number of files with ' +
      'the same naming unless some files are known to be missing.',
  },
```
I just wanted to point out that with the new "sample" concept in microscopy, it is likely that subjects will often have different number of files. Also, as we recommend the `sample_<label>` to be unique throughout the dataset, the naming will also often differ between subjects.

For example, this [dataset](https://zenodo.org/record/5744487#.Ycol3WjMJPY) raise the warning as all sample ids are unique, and one of the subject has 3 samples whereas the others only have 1.
I don't see it as an issue per se, because it's only a warning, but I wanted to share my observation with you.

mariehbourget:
Hi @bids-standard/maintainers,
The implementation of the microscopy validator in this PR is complete and ready for review. 😃 

Note: The ExifReader package that we use to read the XML header in OME-TIFF files seems to support only "regular TIFF" files and not "BigTIFF" files. So in cases where the files are "BigTIFF", the consistency checks between the OME metadata and the JSON metadata is skipped with a warning to the user.

Thanks to @etiennebergeron and @konantian for their help with the validator implementation! 🚀

bids-validator/utils/files/readOMEFile.js
```diff
@@ -0,0 +1,17 @@
+import ExifReader from 'exifreader'
```
rwblair, line 1: Currently exifreader in node will duplicate the buffer passed to it:
https://github.com/mattiasw/ExifReader/blob/2c35c075ec8ee8e8d852631bac91a5b0deb565e4/src/exif-reader.js#L143

This could have some problems for memory consumption if very large files are being loaded. Do we have an idea of how large the files that are going to be validated typically are? Its come up in meetings before but I forget the numbers mentioned. Examples on the ome-tiff site ranged from 100Mb to multiple Gb. Both the browser and node file api will at least read the file into memory once as is.

Going over the specification and playing around with buffer slicing/reading we could write our own code to extract ImageDescription tag from a tiff file using the offsets and sizes in the IFD to try and read only what is needed but that's another thing to test and maintain, I don't want to prematurely optimize.

Unfortunately we don't get any assurances that the actual offset for the contents of the tag will be close to the beginning of the file so just reading in X number of bytes won't always work. I don't have any intuition as to of how common it is to have this type of information at the end of the file.

Aside from the bids-examples are there real world datasets this has been tested against? If it seems to be working fine for real world usage we can merge as is, and then if issues arise in the future from it we can change it.

rwblair (changes requested):
Only one comment in, in readOEMFile, thoughts on potential memory issues, doesn't necessarily have to be fixed at this time but I wanted to see if any had any opinons/thoughts on it.

Other than that any issues are addressed in this PR:
https://github.com/neuropoly/bids-validator/pull/7

bids-validator/utils/files/readOMEFile.js
mariehbourget: Thanks @rwblair for the review!
We do have concerns about memory usage but did not find a good way around it.

Unfortunately, this may become an issue rather sooner than later because OME-TIFF files and datasets can be very large. We have this real-world [dataset](https://zenodo.org/record/5760967#.YdxJI_7MJPY) as example (each file is 592 MB for a total of 16.5 GB in the rawdata folder). It is a sub-set of this 800 GB [dataset](https://gui.dandiarchive.org/#/dandiset/000026/draft/files?location=rawdata%2Fsub-I46%2Fses-SPIM%2F).
**EDIT: The [example real-world dataset](https://zenodo.org/record/5840690#.Yd77sP7MJPY) was updated with regular TIF instead of BigTIF. Each file is now about 200-250 MB for a total of 6.5 GB, and the validator runs without errors in a few seconds.**

Reading the beginning of the file only would certainly help but I don't know how frequent the XML is at the end of the file either. For all our examples, it is at the beginning., but I think those examples from OME ([5D datasets](https://docs.openmicroscopy.org/ome-model/5.5.7/ome-tiff/data.html#d-datasets)) are at the end.

Looking at the figure on the right [here](https://docs.openmicroscopy.org/ome-model/5.5.7/ome-tiff/specification.html#ome-xml-metadata), I think maybe the location of the XML block (Y) is indicated in the offset X+8 bytes? I don't know if that makes sense, I'm a bit out of my depth with this.

rwblair: I spent quite a bit of time last week reading the OME-TIFF and TIFF standards and am pretty confident that we could extract the xml data by reading the various offsets, so if in the future this becomes a serious issue I can help who ever would be interested in implementing such a thing. That being said sticking with maintained off the shelf solutions is nice.

Since its working on real world data well keep the current implementation and worry about it if/when it becomes an issue.

mariehbourget: Perfect, thank you for your help!

PR #1362, opened 2021-10-10
FIX: remove unnecessarily early definition of variable from ASL loop (validate TSV columns)

DimitriPapadopoulos:
The initial value of `potentialSidecars` is unused, since it is always overwritten.

https://lgtm.com/projects/g/bids-standard/bids-validator/snapshot/acec0669924a08e7832df462802d51c44ab1fa42/files/bids-validator/validators/tsv/validateTsvColumns.js?sort=name&dir=ASC&mode=heatmap#xdcf5436ed4bb11bc:1


sappelhoff (approved):
Thanks

PR #1367, opened 2021-10-18
[ENH] Accept additional properties within beh.json JSON schema

### LINKED ISSUE #1364: false positive errors on a beh.json file
Some colleagues tries to use a `beh` folder to organize their data and has been getting some errors on the json content that look like false positive.

I have been able to reproduce the error here:
- https://github.com/Remi-Gau/dummy_beh_ds/tree/main/sub-01/beh

Here I copied an `events.tsv` and its side car and renamed it into `beh.tsv`

Apparently the specifications says nothing of how the content of `beh.json` should be organized.

So it looks like the validator is complaining as soon as one puts anything in the json file (an empty `{}` is ok though).

Also note that the validator is not picking up that the `beh.tsv` has both `onset` and `duration` and should then probably be an `events.tsv` - though see [this case in the `bids-example`](https://github.com/bids-standard/bids-examples/blob/master/ds000117/sub-01/ses-meg/beh/sub-01_ses-meg_task-facerecognition_events.tsv).

```
sub-01_task-TEST_beh.json
0.215 KB | application/json
Location:

dummy_beh_ds/sub-01/beh/sub-01_task-TEST_beh.json
Reason:

Invalid JSON file. The file is not formatted according the schema.
Evidence:

should NOT have additional properties
```

---

Suggestions: 

- lenient: do not put any restriction on how the content of `beh.json` files should be organized
- strict and more "agressive": assume that `beh.json` files should have the same structure as `events.json` (will require an update of the spec)

Remi-Gau:
fixes https://github.com/bids-standard/bids-validator/issues/1364 when merged

- [x] allow additional properties for beh.json
- [ ] throw warning for columns of `beh.tsv` with no data_dictionary (is this too strict???)
- [ ] throw warning if `beh.tsv` has `onset` and `duration` column (it should probably be an `events.tsv` in this case)

### DIFF
--- a/bids-validator/tests/json.spec.js
+++ b/bids-validator/tests/json.spec.js
@@ -587,4 +587,22 @@ describe('JSON', function() {
       assert(issues[1].evidence == '.CogPOID should match format "uri"')
     })
   })
+
+  it('*beh.json with extra content throws no error', function() {
+    var jsonObj = {
+      TaskName: 'stroop',
+      trial: {
+        LongName: 'Trial name',
+        Description: 'Indicator of the type of trial',
+        Levels: {
+          congruent: 'Word and color font are congruent.',
+          incongruent: 'Word and color font are not congruent.',
+        },
+      },
+    }
+    jsonDict[beh_file.relativePath] = jsonObj
+    validate.JSON(beh_file, jsonDict, function(issues) {
+      assert(issues.length === 0)
+    })
+  })
 })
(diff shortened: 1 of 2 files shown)


sappelhoff:
> throw warning for columns of beh.tsv with no data_dictionary (is this too strict???)

No, I think this is a fine level of strictness. I think we already warn for any unexplained arbitrary column in TSV files.

sappelhoff (approved):
LGTM so far. I think we can merge this and open an issue for your follow-up points. Just because I think these follow-ups are considerably more difficult to do and it'd delay the easy-but-good part you have contributed here.

sappelhoff:
Thanks @Remi-Gau! Feel free to open issues for your other two points

Issue #1358, opened 2021-10-07
Broken CI


DimitriPapadopoulos:
it looks like "**Build & publish containers / Push Docker image to Docker Hub (pull_request)**" fails consistently.

> **Error:** Username and password required

rwblair:
Hopefully #1357 fixes this. credentials weren't available to PRs from other remotes. Changed it so we only publish on release instead of on any PR against master.

DimitriPapadopoulos:
Yes, the above issue seems to have been fixed, thank you.

However, "**ci/circleci: test**" fails too, and the error seems totally unrelated to the PR:
https://circleci.com/gh/bids-standard/bids-validator/9753?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

DimitriPapadopoulos:
The reported errors don't make any sense at all, they point to `bids-validator/cli.js:81:15`:
```
 FAIL  bids-validator/tests/cli.spec.js
  ● Console

    console.log
      undefined does not exist 

      at bids-validator/cli.js:81:15

    console.log
      /root/project/bids-validator/tests/data/valid_dataset does not exist 

      at bids-validator/cli.js:81:15

    console.log
      /root/project/bids-validator/tests/data/valid_dataset does not exist 

      at bids-validator/cli.js:81:15

    console.log
      /root/project/bids-validator/tests/data/empty_files does not exist 

      at bids-validator/cli.js:81:15

    console.log
      /root/project/bids-validator/tests/data/empty_files does not exist 

      at bids-validator/cli.js:81:15

    console.log
      /root/project/bids-validator/tests/data/valid_dataset does not exist 

      at bids-validator/cli.js:81:15

    console.log
      /root/project/bids-validator/tests/data/valid_dataset does not exist 

      at bids-validator/cli.js:81:15

    console.log
      /root/project/bids-validator/tests/data/valid_dataset does not exist

      at bids-validator/cli.js:81:15
```
I'm rebasing against `master` in the hope the issue might go away.

DimitriPapadopoulos:
Rebasing doesn't help, see #1355.

rwblair:
I haven't been able to figure this one out. Rerunning the tests via SSH into circle the tests pass. Also haven't been able to recreate locally.

sappelhoff:
on an up-to-date `master`, after doing `npm ci` and `npm test`, I get the following:

```
> test
[...]

PR #1375, opened 2021-10-26
run node tests in github actions

rwblair:
See if it gets same problem as in #1358


rwblair:
I'm still clueless as to why fs.existsSync returns false in gitlab, circleci, and githubactions but true when in a proper shell. Maybe worth testing for existeance some other way and catching ENOENT code.

effigies:
> will --colors work for windows gitlab?

LOLNO

effigies:
@rwblair Assume you've seen https://stackoverflow.com/questions/66185466/npm-script-fails-is-not-recognized-as-an-internal-or-external-command? It's not obvious that there are useful suggestions there, but it does seem that for some reason we're generating unix test scripts when we should be generating Windows.

rwblair:
@effigies thank you. Made me wonder why things in the package.json were calling bins with relative paths. Turns out to be for no good reason. Now its failing for new reasons.

Also looks like eslint on ubuntu might not really be running. The command takes a minute on my local machine, but is done in seconds on ci.

effigies:
This looks good. Cleaned up in #1426 with one additional commit.

PR #1378, opened 2021-11-08
Fix missing LICENSE on PyPI

musicinmybrain:
Add a symbolic link to the `LICENSE` file in the parent directory so that it (already named in `MANIFEST.in`) appears in the PyPI source and wheel distributions.

This is particularly important because the chosen MIT license requires that the copyright statement and license text be reproduced in derivative works; to comply with the license, library users need it to be included, not just referenced.

While we’re at it, add a symbolic link for `CODE_OF_CONDUCT.md`, since it’s already named in `MANIFEST.in`, implying somebody wanted to include it in the source distribution.


sappelhoff (commented):
Thanks for catching this! Personally, I don't think we need to ship the code of conduct with the package, the CoC has a good place on GitHub, where collaboration actually takes place. (So I'd vote for removing the CoC also from the manifest file)

However, we should definitely include the license. I am just not sure whether what you did here will do the trick: Did you just create a new markdown file that contain a the text  `../LICENSE`? Or is this some GitHub rendering of something else you did?

musicinmybrain:
> Thanks for catching this! Personally, I don't think we need to ship the code of conduct with the package, the CoC has a good place on GitHub, where collaboration actually takes place. (So I'd vote for removing the CoC also from the manifest file)

Sounds good. I’ll modify the second commit to do that instead.

----

> However, we should definitely include the license. I am just not sure whether what you did here will do the trick: Did you just create a new markdown file that contain a the text `../LICENSE`? Or is this some GitHub rendering of something else you did?

I did (working on Linux, so POSIX shell syntax):

```sh
git clone https://github.com/bids-standard/bids-validator.git
cd bids-validator/bids-validator
ln -s ../LICENSE
git add LICENSE
git commit
```

In my testing, `python3 setup.py sdist` followed the symbolic link and included a copy of the `LICENSE` file from the parent directory, and `python3 setup.py bdist_wheel` copied the `LICENSE` file into the `dist-info` metadata directory. However, it’s worth testing that the symbolic link works with your PyPI upload workflow.

Or, if you prefer, I could add a separate copy of the `LICENSE` file from the root of the repository rather than using a symbolic link.

sappelhoff (approved):
Thanks @musicinmybrain - that sounds reasonable.

sappelhoff:
@all-contributors please add @musicinmybrain for platform

PR #1381, opened 2021-11-09
Allows EDF extensions to be upper case in filenames

### LINKED ISSUE #47: upper- vs lower- casing of file extensions: Validator should send informative error message
Upon uploading to OpenNeuro, I get an error for uploading an EDF file with `.EDF` extension. Apparently only `.edf` is valid, however, the error message is not so clear about this.

> Files with such naming scheme are not part of BIDS specification. This error is most commonly caused by typos in file names that make them not BIDS compatible. Please consult the specification and make sure your files are named correctly. If this is not a file naming issue (for example when including files not yet covered by the BIDS specification) you should include a ".bidsignore" file in your dataset (see https://github.com/bids-standard/bids-validator#bidsignore for details). Please note that derived (processed) data should be placed in /derivatives folder and source data (such as DICOMS or behavioural logs in proprietary formats) should be placed in the /sourcedata folder.

>Location:
/sub-kumc1/ses-extraoperative/ieeg/sub-kumc1_ses-extraoperative_task-interictal_run-01_ieeg.EDF

See: https://github.com/bids-standard/bids-validator/pull/1381#issuecomment-965253877

An improved error/warning message is desired. 

This can be marked as an easy issue.

adam2392:
Fixes: bids-standard/bids-validator#47

### DIFF
--- a/bids-validator/schema/datatypes/eeg.yaml
+++ b/bids-validator/schema/datatypes/eeg.yaml
@@ -4,6 +4,7 @@
   extensions:
     - .json
     - .edf
+    - .EDF
     - .vhdr
     - .vmrk
     - .eeg
--- a/bids-validator/schema/datatypes/ieeg.yaml
+++ b/bids-validator/schema/datatypes/ieeg.yaml
@@ -5,6 +5,7 @@
     - .mefd/
     - .json
     - .edf
+    - .EDF
     - .vhdr
     - .eeg
     - .vmrk
--- a/bids-validator/tests/tsv.spec.js
+++ b/bids-validator/tests/tsv.spec.js
@@ -180,9 +180,9 @@ describe('TSV', function() {
       '/sub-08/ses-test/eeg/sub-08_ses-test_task-linebisection_run-01_eeg.fif',
   }
   var ieegFile = {
-    name: 'sub-08_ses-test_task-linebisection_run-01_ieeg.edf',
+    name: 'sub-08_ses-test_task-linebisection_run-01_ieeg.EDF',
     relativePath:
-      '/sub-08/ses-test/ieeg/sub-08_ses-test_task-linebisection_run-01_ieeg.edf',
+      '/sub-08/ses-test/ieeg/sub-08_ses-test_task-linebisection_run-01_ieeg.EDF',
   }
 
   var btiFiles = [


sappelhoff (commented):
technically you are correct, that both upper and lowercase `.edf` extensions are fine. I am just not sure whether this is the right fix, or whether we should rather convert parsed extensions to lowercase and spare ourselves the trouble to double code each casing variant everywhere 🤔

adam2392:
So this would require a PR to the specification saying we REQUIRE lower case extension for EDF files. This imo does make everyone's life easier (even mne-bids). If this is the move, then lmk.

sappelhoff:
> So this would require a PR to the specification saying we REQUIRE lower case extension for EDF files. This imo does make everyone's life easier (even mne-bids). If this is the move, then lmk.

I also think that'd be easier ... but what I actually had in mind was bids-validator calling `extension.lower()` on whatever "extension" it validates 🤔

adam2392:
> > So this would require a PR to the specification saying we REQUIRE lower case extension for EDF files. This imo does make everyone's life easier (even mne-bids). If this is the move, then lmk.
> 
> I also think that'd be easier ... but what I actually had in mind was bids-validator calling `extension.lower()` on whatever "extension" it validates 🤔

Oh this sounds good too. Where can we implement this in the current validator?

sappelhoff:
Maybe @rwblair knows or has an opinion?

effigies:
I don't think we generally want to permit case-insensitive file extensions. Would tools handle `".NII.GZ"` with equal facility as `".nii.gz"`? Beyond this, I think treating the entities and suffix, which are case-sensitive in BIDS, separately from the extension is going to be difficult, since they're tightly coupled.

Generally, this feels like a non-problem. The specification says ".edf", the validator validates ".edf"  and raises an error when it's ".EDF". I don't see a good reason to allow multiple capitalizations of the same extension even if there are tools that will happily open either.

If the problem is that users used to being able to freely switch between `.EDF` and `.edf` see an error that only says "This file is not a valid BIDS filename." for a file that is fine except for the extension, then we just need a new error message. Perhaps before checking the full filename, we check whether the the extension is defined anywhere in BIDS and emit a specific error there.

adam2392:
the issue for EDF is that the specification explicitly states both capitalized and lower case are fine. Whereas for all other formats it specifically goes with lower case. 

Would checking the lower case extension be a good solution as @sappelhoff suggested? 

Also improving the error message sounds good to me.

effigies:
Sorry, can you quote the spec? I couldn't find it. I assume you mean BIDS, not EDF?

adam2392:
> Sorry, can you quote the spec? I couldn't find it. I assume you mean BIDS, not EDF?

https://www.edfplus.info/specs/edfplus.html

```
  EDF+ prescribes the following filename convention. Signals, recorded using the same technique and constant amplifier settings, can be stored in one file. Different techniques, or identical techniques but with different amplifier settings, must be stored in separate files. All EDF+ files must have .edf or .EDF as filename extension. See also section 2.3.
```

and then in BIDS, we don't `RECOMMEND/REQUIRE` a specific extension currently. I don't find it as an issue anymore, but it has come up for me because some source files were named differently. 

The best solution could be RECOMMEND lower case extension for EDF in the BIDS spec, and then check lower-cased extensions in the validator?

effigies:
BIDS says[[1]](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/03-electroencephalography.html#eeg-recording-data)[[2]](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/04-intracranial-electroencephalography.html#ieeg-recording-data):

> European data format (Each recording consisting of a .edf file)

This does not indicate to me that `.EDF` is permitted. We could update to specifically disallow `.EDF` if this is unclear, and it does not violate the EDF+ spec for us to select one capitalization.

> The best solution could be RECOMMEND lower case extension for EDF in the BIDS spec, and then check lower-cased extensions in the validator?

A single, case-sensitive extension for a single format seems preferable to allowing multiple capitalizations (or worse, arbitrary capitalizations, which `.lower()` would enable). If we must allow `.EDF` as well as `.edf`, then the cleanest solution would be as an explicit extension, as in this PR. Any "smarter" approach will take development effort and invite introducing bugs.

I still think that a better error message is the best way forward.

sappelhoff:
> Would tools handle ".NII.GZ" with equal facility as ".nii.gz"? Beyond this, I think treating the entities and suffix, which are case-sensitive in BIDS, separately from the extension is going to be difficult, since they're tightly coupled.

true, agreed.

> If the problem is that users used to being able to freely switch between .EDF and .edf see an error that only says "This file is not a valid BIDS filename." for a file that is fine except for the extension, then we just need a new error message.

yeah, also agreed.

+1 from me for 

1. enforcing in BIDS that only `.edf` is allowed (it has been like this in the validator since 2019), despite the EDF standard saying that both edf and EDF are fine (i.e., making a PR to the spec that makes this clear &explicit instead of implicit, as it is currently)
2. improving the BIDS validator warning

PR #1384, opened 2021-11-18
Remove repeated words

DimitriPapadopoulos:


sappelhoff (approved):

bids-validator/schema/entities.yaml
```diff
@@ -42,7 +42,7 @@ acquisition:
     user MAY use to distinguish a different set of parameters used for
     acquiring the same modality.
     For example this should be used when a study includes two T1w images - one
-    full brain low resolution and and one restricted field of view but high
+    full brain low resolution and one restricted field of view but high
```
sappelhoff: @rwblair are the schema files here in bids-validator somehow "imported" from the bids-specification repo and should thus NOT be edited "manually"? (that is, they should only be updated "upstream" in bids-specifiation?) ... or are these just copy pasted and it's fine to edit here?

Otherwise I think this can be merged

nellh: Yeah, fixes for the yaml schema files should be made against the specification repo. They are temporarily added here to make sure tests doesn't depend on external resources while we solved the schema distribution method.

sappelhoff: thanks Nell. @DimitriPapadopoulos could you please revert the change to the schema then? the rest seems good to me.

DimitriPapadopoulos: Done. Note that another merge request has already fixed that the specification repo (https://github.com/bids-standard/bids-specification/pull/934) so a synchronization should fix it here too.

PR #1386, opened 2021-11-18
Prevent CLI test failure in initial test for directory access

nellh:
fs.exists is deprecated and I'm not sure why it doesn't work but switching it out to opendirSync seems like a good idea anyways.

### DIFF
--- a/bids-validator/cli.js
+++ b/bids-validator/cli.js
@@ -53,34 +53,37 @@ export function cli(argumentOverride) {
       return filenamesOnly()
     }
 
-    if (fs.existsSync(dir)) {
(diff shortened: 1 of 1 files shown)


rwblair:
@nellh amazing.

I just now saw the fs.exists documentation recommending to try opening things instead of testing for existence.

Looks like gitlab/windows might still be unhappy. We want to fix that here or is it ok to merge?

nellh:
> @nellh amazing.
> 
> I just now saw the fs.exists documentation recommending to try opening things instead of testing for existence.
> 
> Looks like gitlab/windows might still be unhappy. We want to fix that here or is it ok to merge?

We should merge this one, I'm working on that separately.

Issue #1326, opened 2021-09-11
"n/a" for required tag generates "Invalid JSON file"; "else" error?


nicholst:
Using the validator to check a publically available PET dataset I got a bunch of "Error 1: [Code 55] JSON_SCHEMA_VALIDATION_ERROR" errors that were generated by "should be number" evidence, however, examining the flagged JSON files I found that the 'invalid' values were `n/a`.

The spec says 
> Throughout BIDS you can indicate missing values with n/a (for "not available").

Now, as it happens the errors are generated on required tags, and perhaps the spec should be clearer that "REQUIRED" isn't satisifed by supplying a `n/a` value.

But for the validator, it would be more informative if, when a required tag *has* a value but that value is "n/a", that the error reflects that (e.g. "Required tag set to n/a value"), instead of treating it as a generic low-level error.

Also, among this dataset's other errors, there is one "Invalid JASON error" that has evidence that I can't parse:
```
should match "else" schema
```
I googled around but couldn't figure what this was supposed to mean... is it just a validator error?

Dataset is: https://www.nitrc.org/frs/download.php/12247/Dataset_BIDS.zip
Full log attached

[dataset_bids_errors.txt](https://github.com/bids-standard/bids-validator/files/7148078/dataset_bids_errors.txt)

sappelhoff:
>  Perhaps the spec should be clearer that "REQUIRED" isn't satisifed by supplying a n/a value.

I agree that such a clarification would not hurt :+1: 

> But for the validator, it would be more informative if, when a required tag has a value but that value is "n/a", that the error reflects that (e.g. "Required tag set to n/a value"), instead of treating it as a generic low-level error.

I think that's a good suggestion :+1: **However**, that ties in with another feature that we have wanted for a long time now (e.g., bids-standard/bids-validator#59) --> We use https://json-schema.org/ to validate BIDS metadata, so we specify a "structure", datatype, or pattern in a JSON file, and then get a BIDS JSON file, and compare the two (the schema, and the data). Unfortunately this does not really allow us to easily customize warnings or inject custom code in the right spots. It just tells us where the data and the schema don't match. We would want some code in between that analyzes the type of mismatch, and then triages this type into a set of specific warnings or hints.

I am convinced it's somehow possible, but it's not an inbuilt feature and some level of expertise would be required to get this done. (But the impact would be big and it would be awesome).

> Also, among this dataset's other errors, there is one "Invalid JASON error" that has evidence that I can't parse:

[...]

PR #1396, opened 2021-12-14
Update pet.json to accept "n/a" for InjectedMass and SpecificRadioact…

mnoergaard:
This PR solves the issues raised in https://github.com/bids-standard/bids-validator/issues/1326#issuecomment-993344409

This PR updates InjectedMass and SpecificRadioactivity to accept "n/a" according to the PET-BIDS specification. Currently, it only accepts numbers, and therefore complains when a string is entered. Therefore, it has been modified to also accept a string for the case "n/a".


sappelhoff (commented):
thanks @mnoergaard! I pushed a commit that will ensure that the only valid string that can be passed is `"n/a"` (or instead of a str, any number as previously)

sappelhoff (approved):

mnoergaard:
Great, LGTM! Thanks @sappelhoff!

effigies (approved):
LGTM.

effigies:
We should get an example that takes advantage of this so that it's tested.

Issue #1393, opened 2021-12-07
BIDS URI support


effigies:
xref https://github.com/bids-standard/bids-specification/pull/918

Minimally we need:

- [x] Permit `dataset_description.DatasetLinks` object
   * Empty string as a key is an error
   * Values are URIs (relative paths accepted)
- [ ] Handle BIDS-URIs as part of NIfTI validation routine for `IntendedFor`
    * see https://github.com/bids-standard/bids-validator/issues/1393#issuecomment-1019520428
    * however, see also: https://github.com/bids-standard/bids-validator/pull/1409#issuecomment-1191749813

Long-term:

* Deprecation warning if BIDS-URI field is not URI
* Validate all `IntendedFor` (not just for NIfTI), BIDS URI validation only if "local"

sappelhoff:
> Empty string as a key is an error

That can probably be done in JSON schema terms via:
```json
{
 "not": {
  "anyOf": [
    { "required": [ "" ] }
  ]
}
```

https://stackoverflow.com/a/70266255/5201771

sappelhoff:
The bids-validator currently contains a check for the `IntendedFor` field as part of the NIfTI validation:

https://github.com/bids-standard/bids-validator/blob/4feb91b8aa9506c123d38327485dd5d4cd69e80b/bids-validator/validators/nifti/nii.js#L1247-L1265

This leads to issues during validation with BIDS URIs, see this bids-examples PR: https://github.com/bids-standard/bids-examples/pull/304, and the problem log:

```shell
bids-validator 7t_trt -c /home/runner/work/bids-examples/bids-examples/bidsconfig.json --ignoreNiftiHeaders
bids-validator@1.8.10-dev.0
bids-specification@disable
(node:1867) Warning: Closing directory handle on garbage collection
(Use `node --trace-warnings ...` to show where the warning was created)
	1: [ERR] 'IntendedFor' field needs to point to an existing file. (code: 37 - INTENDED_FOR)
		./sub-01/ses-1/fmap/sub-01_ses-1_run-1_phasediff.nii.gz
			Evidence: bids::sub-01/ses-1/func/sub-01_ses-1_task-rest_acq-fullbrain_run-1_bold.nii.gz
...
```

For non-NIfTI files, this is not a problem, because such a check was never implemented as far as I know.

This is something that needs to be fixed before we can merge BIDS URI support into the spec.

effigies:
Done in schema validator. Not planned for legacy validator.

PR #1404, opened 2022-01-16
Add DatasetLinks to JSON schema

sappelhoff:
addresses point one of #1393: Adding the `DatasetLinks` metadata field to the `dataset_description.json` JSON schema.

xref https://github.com/bids-standard/bids-specification/pull/918

### DIFF
--- a/bids-validator/tests/json.spec.js
+++ b/bids-validator/tests/json.spec.js
@@ -484,27 +484,42 @@ describe('JSON', function() {
     })
   })
 
-  it('dataset_description.json should raise on empty key in DatasetLinks', function() {
+  it('dataset_description.json should raise on bad keys in DatasetLinks', function() {
     var jsonObj = {
       Name: 'Example Name',
       BIDSVersion: '1.4.0',
       DatasetLinks: {
         mylink: 'https://www.google.com',
         '': 'https://www.yahoo.com',
+        'mylink!': ':/path',
+        'my link': ':/another/path',
       },
     }
     jsonDict[dataset_description_file.relativePath] = jsonObj
     validate.JSON(dataset_description_file, jsonDict, function(issues) {
-      assert(issues.length === 2)
+      assert(issues.length === 6)
       assert(
         issues[0].evidence ==
           '.DatasetLinks should NOT be shorter than 1 characters',
       )
       assert(issues[1].evidence == ".DatasetLinks property name '' is invalid")
+      assert(
+        issues[2].evidence ==
+          '.DatasetLinks should match pattern "^[a-zA-Z0-9]*$"',
+      )
+      assert(
+        issues[3].evidence ==
+          ".DatasetLinks property name 'mylink!' is invalid",
+      )
+      assert(issues[4].evidence == issues[2].evidence)
+      assert(
+        issues[5].evidence ==
+          ".DatasetLinks property name 'my link' is invalid",
+      )
     })
   })
 
-  it('dataset_description.json should raise on wrong key in DatasetLinks', function() {
+  it('dataset_description.json should raise on non-object value in DatasetLinks', function() {
     var jsonObj = {
       Name: 'Example Name',
       BIDSVersion: '1.4.0',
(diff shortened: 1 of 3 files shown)


effigies:
The error message for the empty string is hard to interpret. Can we just make the format of the key to be alphanumeric of length >=1?

sappelhoff:
It may be possible using [`property-names`](https://json-schema.org/understanding-json-schema/reference/object.html#property-names)

bids-validator/validators/json/schemas/dataset_description.json
```diff
@@ -52,6 +52,19 @@
     "DatasetDOI": {
       "type": "string"
     },
+    "DatasetLinks": {
+      "type": "object",
+      "properties": { },
+      "propertyNames": {
+        "type": "string",
+        "minLength": 1,
+        "pattern": "^[a-zA-Z0-9]*$"
```
effigies, line 61: What about:

```suggestion
        "pattern": "^[a-zA-Z0-9]+$"
```

Or would that make the error less clear?

sappelhoff: yes, that was my train of thought: better to be explicit here. Everyone can guess what `minLength` means, but few people know the difference between `*` and `+` in a regex

sappelhoff: don't know about the error message though (on re-reading, that seems to be your concern, whereas mine was with future devs)

sappelhoff: it would change `.DatasetLinks should NOT be shorter than 1 characters` to `'.DatasetLinks should match pattern "^[a-zA-Z0-9]+$"'`

effigies: My concern is that we get both:

```
'.DatasetLinks should NOT be shorter than 1 characters'
".DatasetLinks property name '' is invalid"
```

The second one seems more helpful.

bids-validator/tests/json.spec.js
```diff
@@ -467,6 +467,80 @@ describe('JSON', function() {
     relativePath: '/dataset_description.json',
   }
 
+  it('dataset_description.json should validate DatasetLinks', function() {
+    var jsonObj = {
+      Name: 'Example Name',
+      BIDSVersion: '1.4.0',
+      DatasetLinks: {
+        mylink: 'https://www.google.com',
```
effigies, line 475: Let's check that the example URIs we list in the spec PR don't cause errors:

```suggestion
        mylink: 'https://www.google.com',
        deriv1: 'derivatives/derivative1',
        phantoms: 'file:///data/phantoms',
        ds000001: 'doi:10.18112/openneuro.ds000001.v1.0.0',
```

bids-validator/validators/json/schemas/dataset_description.json
sappelhoff: > My concern is that we get both

I see what you mean, but that's also the case after your suggested fix unfortunately. It'd just be 

```
'.DatasetLinks should match pattern "^[a-zA-Z0-9]+$"'
".DatasetLinks property name '' is invalid"
```

instead.

I think both of these error messages are helpful though (much more helpful than before in any case, so thanks for your feedback)

effigies (approved):
While it would be nice not to raise two issues for a single bad key, we shouldn't hold up this PR on that basis. One suggestion though.

bids-validator/tests/json.spec.js
```diff
@@ -467,6 +467,83 @@ describe('JSON', function() {
...
+        deriv1: 'derivatives/derivative1',
```
sappelhoff, line 476: it's now failing, because `'derivatives/derivative1'` doesn't match a URI format :thinking:

bids-validator/validators/json/schemas/dataset_description.json
```diff
@@ -52,6 +52,19 @@
...
+      },
+      "additionalProperties": {
+        "type": "string",
+        "format": "uri"
```
effigies: Apparently what we actually want is a [uri-reference](https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.7.3.5).

```suggestion
        "format": "uri-reference"
```

bids-validator/tests/json.spec.js
```diff
@@ -467,6 +467,83 @@ describe('JSON', function() {
...
+  })
+
+  it('dataset_description.json should raise on invalid values in DatasetLinks', function() {
+    var jsonObj = {
+      Name: 'Example Name',
+      BIDSVersion: '1.4.0',
+      DatasetLinks: {
+        mylink1: 'https://www.google.com',
+        mylink2: 1,
+        mylink3: 'nope',
```
effigies: I guess this is now valid. Can either drop it or make it invalid.

```suggestion
        mylink3: ':/nope',
```

sappelhoff: I'll drop it and change it to a test that fails for keys like `I am a key with whitespace and not alphanumeric "§%`

PR #1409, opened 2022-01-30
Relax NIfTI IntendedFor check when BIDS-URIs are used

sappelhoff:
see:

- https://github.com/bids-standard/bids-specification/pull/918
- https://github.com/bids-standard/bids-validator/issues/1393#issuecomment-1019520428
- https://github.com/bids-standard/bids-examples/pull/304#issuecomment-1019521206

In this PR I want to relax the check for IntendedFor data to actually exist **IF** the IntendedFor field is a BIDS-URI. That is, it it's a bids-uri, we don't raise if a file doesn't exist ... else, we still do.

### DIFF
--- a/bids-validator/validators/nifti/nii.js
+++ b/bids-validator/validators/nifti/nii.js
@@ -1258,11 +1258,10 @@ export default function NIFTI(
         const intendedForFile = intendedFor[key]
         // Only check for presence of IntendedFor files if not a BIDS-URI
         // https://github.com/bids-standard/bids-validator/issues/1393
-        if (!intendedForFile.startsWith("bids:")) {
+        if (!intendedForFile.startsWith('bids:')) {
           checkIfIntendedExists(intendedForFile, fileList, issues, file)
           checkIfValidFiletype(intendedForFile, issues, file)
         }
-
       }
     }
   }
--- a/bids-validator/validators/nifti/nii.js
+++ b/bids-validator/validators/nifti/nii.js
@@ -1256,8 +1256,13 @@ export default function NIFTI(
 
       for (let key = 0; key < intendedFor.length; key++) {
         const intendedForFile = intendedFor[key]
-        checkIfIntendedExists(intendedForFile, fileList, issues, file)
-        checkIfValidFiletype(intendedForFile, issues, file)
+        // Only check for presence of IntendedFor files if not a BIDS-URI
+        // https://github.com/bids-standard/bids-validator/issues/1393
+        if (!intendedForFile.startsWith("bids:")) {
+          checkIfIntendedExists(intendedForFile, fileList, issues, file)
+          checkIfValidFiletype(intendedForFile, issues, file)
+        }
+
       }
     }
   }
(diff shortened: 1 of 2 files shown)


bids-validator/validators/nifti/nii.js
```diff
@@ -1256,8 +1256,13 @@ export default function NIFTI(
 
       for (let key = 0; key < intendedFor.length; key++) {
         const intendedForFile = intendedFor[key]
-        checkIfIntendedExists(intendedForFile, fileList, issues, file)
-        checkIfValidFiletype(intendedForFile, issues, file)
+        // Only check for presence of IntendedFor files if not a BIDS-URI
+        // https://github.com/bids-standard/bids-validator/issues/1393
+        if (!intendedForFile.startsWith("bids:")) {
```
effigies: If it starts with `bids::`, we could still do the check within the local dataset.

sappelhoff: done in https://github.com/bids-standard/bids-validator/pull/1409/commits/7c0cfabbc676d653f7156ee17fec1f0c7ebc2687

sappelhoff:
> If it starts with bids::, we could still do the check within the local dataset.

@effigies I am actually reconsidering this: 

- the `IntendedFor` validated here (NIfTI only) is **subject relative**
- [link to spec](https://bids-specification.readthedocs.io/en/latest/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#using-intendedfor-metadata)
- [link to schema](https://github.com/bids-standard/bids-specification/blob/a93d4f2a27b2b79ba569a1d02f1e7f560af3525f/src/schema/objects/metadata.yaml#L1418-L1433)
- whereas BIDS URIS (also local ones, like `bids::<relative-path>`) are always relative to the root of the dataset (see https://github.com/bids-standard/bids-specification/pull/918)

I think I'll just revert https://github.com/bids-standard/bids-validator/pull/1409/commits/7c0cfabbc676d653f7156ee17fec1f0c7ebc2687.

I would argue that we can merge this for two reasons:

1. we can extend IntendedFor checks to work with BIDS URIs in another PR. This very basic PR doesn't take away any previous validation capabilities ... it only avoids unreasonable errors for currently non-implemented features.
2. IntendedFor checks arguably need a more general overhaul (which @rwblair is starting in #1419), but this PR removes an initial blocker to add some BIDS URIs to the examples (e.g., https://github.com/bids-standard/bids-examples/pull/304#issuecomment-1019521206)

note that this PR **does not** close #1393 ... we still need proper IntendedFor validation, and we need to make it work:

- for all files (not just NII)
- for IntendedFor fields that use dataset-relative paths (soon to be deprecated in favor of BIDS URIs)
- for BIDS URIs (only local ones)


EDIT: this PR makes the examples pass: https://github.com/bids-standard/bids-examples/pull/304

bids-validator/validators/nifti/nii.js
sappelhoff: however, see https://github.com/bids-standard/bids-validator/pull/1409#issuecomment-1191749813

effigies (approved):
I'm fine with this. @rwblair @nellh  Do you think we should have a warning for "This feature is not fully validated, future versions of the validator may error."?

PR #1413, opened 2022-02-03
Schema prototyping

nellh:
Some work in progress attempts to work with updated yaml schemas from bids-specification and extend with issues definitions.


nellh:
One breaking change in the existing validator required for this branch is the [removal of a `File.slice` call in readNiftiHeader](https://github.com/bids-standard/bids-validator/pull/1413/commits/242389187816af12bdceff5adb338a4ac6cfb339). A fix for this will be needed to make this mergable.

nellh:
Status on this is that it can now run the existing validator and load the full schema. From here need to add the ability to generate the context based on the schema and apply rules. To apply rules, we need to evaluate expressions (may need to tweak the specification expressions).

nellh:
Closing in favor of #1455

PR #1418, opened 2022-02-10
Add option to throw an error if specified modalities are detected

rwblair:


bids-validator/src/options.js
```diff
@@ -21,6 +21,12 @@ export function parseOptions(argumentOverride) {
       'ignoreSubjectConsistency',
       'Skip checking that any given file for one subject is present for all other subjects.',
     )
+    .option('blacklistModalities', {
+      default: [],
+      describe: 'Array of modalities to error on if detected.',
+      array: true,
+      choices: ['MRI', 'PET', 'MEG', 'EEG', 'iEEG', 'Microscopy'],
```
rwblair, line 28: @nellh 

should this array be defined elsewhere and exported/imported, could be reused here as well:
https://github.com/bids-standard/bids-validator/blob/c3670466ef2e669e5cda7ee672c38908b86bf88b/bids-validator/utils/summary/collectModalities.js#L4

Do you think secondary modalities should be an option?

bids-validator/utils/consoleFormat.js
```diff
@@ -66,19 +66,23 @@ function logIssues(issues, color, options) {
       if (!file || !file.file) {
         continue
       }
-      output.push('\t\t.' + file.file.relativePath)
+      let indent = '\t\t'
+      if (file.file.relativePath) {
```
rwblair, line 70: Currently `.undefined` is shown if an invalid file path is passed. For this issue it seemed useful to collect modality evidence lines without a specific file, so I added logic to avoid printing undefined and to indent the following lines accordingly. Looks nicer, but not sure if its actually better overall.

bids-validator/src/options.js
nellh: I don't think we need secondary modalities for this.

It's also defined here, we should combine what we can. https://github.com/bids-standard/bids-validator/blob/c3670466ef2e669e5cda7ee672c38908b86bf88b/bids-validator/src/schemaTypes.js#L7

bids-validator/utils/consoleFormat.js
nellh: This looks good to me.

bids-validator/src/options.js
nellh: On second thought, the definition in schemaTypes is kind of its own thing since it's just mirrored from the specification schema to simplify the implementation of loading those files. I think this is good as is.

nellh (approved):
Tested this with OpenNeuro locally and it works as expected. :+1:

PR #1415, opened 2022-02-06
[ENH] Microscopy: Add validation of JSON for photo files

mariehbourget:
This PR adds validation of the new JSON file associated with microscopy photo files in:

- https://github.com/bids-standard/bids-specification/pull/1000.

Still WIP as the spec PR is not approved yet.

The new JSON file includes the `IntendedFor` field.
I used the same error code (`37`) and validation steps as in `nii.js` [here](https://github.com/bids-standard/bids-validator/blob/4feb91b8aa9506c123d38327485dd5d4cd69e80b/bids-validator/validators/nifti/nii.js#L1307).

New photos and associated JSON files are added in microscopy examples in:

- https://github.com/bids-standard/bids-examples/pull/307.

PR #1419, opened 2022-02-11
[WIP] Intendedfor checks overhaul

rwblair:
This is an attempt to centralize where checks for existing files for intendedFor fields happens. 

Motivated by #1415 and #1409 

In nifti validation after checking that the intended for target file exists a check is made to make sure the target ends with .nii(.gz). This check can be moved to json schema  validation, but unfortunately...

There are a whole mess of fieldmap json files that have no json schema files. M0Scan.json existed but was not being called.

Todo:
- [ ] add a json schema for the 4 different cases of fieldmap files.
- [ ] add logic to validators/json/json.js to load the appropriate one.
- [ ] `validTarget` that I added to M0Scan schema should probably move to common definitinons as `validNiiTarget` or something.
- [ ] Tests that all these schemas are getting used
- [ ] update existing intended for tests to look for json schema validation error code


rwblair:
related #1393

Issue #1223, opened 2021-03-10
Migrate gitlab CI + CircleCI to GitHub Actions?


sappelhoff:
~I broke gitlab ci with #1221 (just noticed after merging, because the gitlab ci job is apparently not run on PRs?), then tried to fix it with several commits to master, but failed. No idea what the issue is.~

edit to original post: GitLab CI was fixed in https://github.com/bids-standard/bids-validator/commit/5ee85764957b8b07681ac33c92b7d3052c01ce45

But in a separate meeting we agreed that it would be great to consolidate all CIs under GitHub actions. ~That's work in progress now on #1224~

**Update: #1224 was merged into https://github.com/bids-standard/bids-validator/tree/ci/migrate/gh-actions so that in the future someone may pick up that work and complete it. But it'd also be fair enough just to delete https://github.com/bids-standard/bids-validator/tree/ci/migrate/gh-actions and start from scratch, because neither @rwblair nor I have managed to finish that work in more than 6 months.**

rwblair:
After #1426 the following circleci workflows still need to be migrated to github actions:

- [ ] gh-pages_deployment
- [ ] test_docker
- [ ] githubPagesTest
- [ ] pypi_precheck
- [ ] pypi_deployment

PR #1426, opened 2022-03-16
CI: Run Node tests in GitHub actions

effigies:
This is just #1375 without a lot of the experimentation. I did a little cherry-picking and history rewriting to get a small set of commits. My only suggestion on #1375 I've included as an additional commit.


rwblair:
@effigies thank you so much, really needed to be done to fix that insane rebase I had done.

sappelhoff:
xref: this addresses some of the points in  #1223

PR #1433, opened 2022-03-22
Remove gitlabci badge, add gh actions badges

rwblair:
Removing gitlabci badge now that we can test windows in gh actions. Added what I thought were most pertinent workflow badges.


sappelhoff (approved):

sappelhoff (commented):
Oh, I think we also need to remove https://github.com/bids-standard/bids-validator/blob/da94aa9773e750cf44605b1fe039d5c9962d0807/.gitlab-ci.yml#L1

otherwise gitlab CI will keep showing up red in our checks

rwblair:
@nellh should we remove the gitlab ci file, or would you prefer to disable the github integration on the gitlab end?

nellh:
> @nellh should we remove the gitlab ci file, or would you prefer to disable the github integration on the gitlab end?

Both sounds good...

PR #1439, opened 2022-04-02
Updated bids validator link from http to https

Spectre-ak:
Updated bids validator link from http to https


effigies:
That's reasonable.

PR #1441, opened 2022-04-06
Add ordering error for eeg channels.tsv

### LINKED ISSUE #1411: Channel columns: split cases "MISSING" and "WRONG_ORDER"
As discussed in #1410 
Currently, bids-validator checks the first columns in channels.tsv and throws a **[CODE72] MISSING_TSV_COLUMNS_IEEG_CHANNELS** error if the first columns are not all required columns, i.e. if the order of the columns is wrong.

grafik

Ideally ,the error message should in this be more precise. When all required columns are present, but not in the right order, it should throw an error that is something like: **WRONG_ORDER_TSV_COLUMN_IEEG_CHANNELS**

The corresponding source code I think can be found [here](https://github.com/bids-standard/bids-validator/blob/4feb91b8aa9506c123d38327485dd5d4cd69e80b/bids-validator/validators/tsv/tsv.js#L406), where each single column name is checked to be at the correct position, e.g.
`checkheader('name', 0, file, 72)`
A solution might be to add a new method before the positional checks that checks simply if all required columns are present. If not, then we get the MISSING error. We would then also change the error code in the `checkheader` method to a different error code which would give us information about the wrong order.

Would this be the preferred way to go?

I am happy to contribute to this, but I have no prior experience in JS so it might be a bit bumpy

rwblair:
fixes #1411

### DIFF
--- a/bids-validator/tests/tsv.spec.js
+++ b/bids-validator/tests/tsv.spec.js
@@ -455,12 +455,12 @@ describe('TSV', function() {
   })
 
   it('should not allow EEG channels.tsv files with name column in wrong place', function() {
-    var tsv = 'header-one\ttype\tunits\tname\n' + 'value-one\tEEG\tmV\tvalue-name'
+    var tsv =
+      'header-one\ttype\tunits\tname\n' + 'value-one\tEEG\tmV\tvalue-name'
     validate.TSV.TSV(channelsFileEEG, tsv, [], function(issues) {
       assert(issues.length === 1 && issues[0].code === 228)
     })
   })
-
 
   it('should not allow EEG channels.tsv files without type column', function() {
     var tsv = 'name\theader-two\tunits\n' + 'value-one\tEEG\tmV'
@@ -504,10 +504,11 @@ describe('TSV', function() {
       'name\ttype\tunits\thigh_cutoff\tlow_cutoff\n' +
       'value-name\tECOG\tmV\tvalue-highcut\tvalue-lowcut'
     validate.TSV.TSV(channelsFileIEEG, tsv, [], function(issues) {
-      assert(issues.length === 2 && issues[0].code === 229 && issues[1].code === 229)
+      assert(
+        issues.length === 2 && issues[0].code === 229 && issues[1].code === 229,
+      )
     })
   })
-
 
   it('should not allow iEEG channels.tsv files without high_cutoff column', function() {
     var tsv =
--- a/bids-validator/tests/tsv.spec.js
+++ b/bids-validator/tests/tsv.spec.js
@@ -454,6 +454,14 @@ describe('TSV', function() {
     })
   })
 
+  it('should not allow EEG channels.tsv files with name column in wrong place', function() {
+    var tsv = 'header-one\ttype\tunits\tname\n' + 'value-one\tEEG\tmV\tvalue-name'
+    validate.TSV.TSV(channelsFileEEG, tsv, [], function(issues) {
+      assert(issues.length === 1 && issues[0].code === 228)
+    })
+  })
+
+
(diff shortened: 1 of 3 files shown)


bids-validator/tests/tsv.spec.js
```diff
@@ -454,6 +454,14 @@ describe('TSV', function() {
     })
   })
 
+  it('should not allow EEG channels.tsv files with name column in wrong place', function() {
+    var tsv =
+      'header-one\ttype\tunits\tname\n' + 'value-one\tEEG\tmV\tvalue-name'
+    validate.TSV.TSV(channelsFileEEG, tsv, [], function(issues) {
+      assert(issues.length === 1 && issues[0].code === 228)
```
sappelhoff:
```suggestion
      assert(issues.length === 1 && issues[0].code === 230)
```

bids-validator/validators/tsv/tsv.js
```diff
@@ -408,9 +419,9 @@ const TSV = (file, contents, fileList, callback) => {
     file.relativePath.includes('/meg/') &&
     file.name.endsWith('_channels.tsv')
   ) {
-    checkheader('name', 0, file, 71)
-    checkheader('type', 1, file, 71)
-    checkheader('units', 2, file, 71)
+    checkheader('name', 0, file, 71, 228)
+    checkheader('type', 1, file, 71, 228)
+    checkheader('units', 2, file, 71, 228)
```
sappelhoff:
```suggestion
    checkheader('name', 0, file, 71, 230)
    checkheader('type', 1, file, 71, 230)
    checkheader('units', 2, file, 71, 230)
```

bids-validator/validators/tsv/tsv.js
```diff
@@ -419,9 +430,9 @@ const TSV = (file, contents, fileList, callback) => {
     file.relativePath.includes('/eeg/') &&
     file.name.endsWith('_channels.tsv')
   ) {
-    checkheader('name', 0, file, 71)
-    checkheader('type', 1, file, 71)
-    checkheader('units', 2, file, 71)
+    checkheader('name', 0, file, 71, 228)
+    checkheader('type', 1, file, 71, 228)
+    checkheader('units', 2, file, 71, 228)
```
sappelhoff:
```suggestion
    checkheader('name', 0, file, 71, 230)
    checkheader('type', 1, file, 71, 230)
    checkheader('units', 2, file, 71, 230)
```

bids-validator/utils/issues/list.js
```diff
@@ -1125,4 +1125,16 @@ export default {
     reason:
       'A BIDS dataset MUST NOT contain more than one `README` file (with or without extension) at its root directory.',
   },
+  229: {
+    key: 'INCORRECT_ORDER_TSV_COLUMN_CHANNELS',
```
sappelhoff:
```suggestion
    key: 'INCORRECT_ORDER_TSV_COLUMN_CHANNELS_IEEG',
```

bids-validator/utils/issues/list.js
```diff
@@ -1125,4 +1125,16 @@ export default {
...
+    severity: 'error',
+    reason:
+      "The column names of the IEEG channels file must be in the following order ['name', 'type', 'units', 'low_cutoff', 'high_cutoff']",
+  },
+  230: {
+    key: 'INCORRECT_ORDER_TSV_COLUMN_CHANNELS',
```
sappelhoff:
```suggestion
    key: 'INCORRECT_ORDER_TSV_COLUMN_CHANNELS_EEG',
```

sappelhoff (approved):
I rebased this, apart from my comment about the duplicate "key" in the issue list, I think this is good to merge.

PR #1443, opened 2022-04-11
rules: add png and tif to [i|m]eeg

sappelhoff:
should be merged **after** https://github.com/bids-standard/bids-specification/pull/1049 has been merged.


bids-validator/bids_validator/rules/session_level_rules.json
```diff
@@ -77,6 +77,8 @@
         "_meg.json",
         "_coordsystem.json",
         "_photo.jpg",
```
effigies, line 79: Looks like we should have a separate PR to escape all these dots...

PR #1474, opened 2022-06-24
[SCHEMA] .bidsignore implementation for Deno

nellh:
Ended up getting the existing module working for this because my other approach using the Deno standard library globs wasn't working in some subtle ways.


rwblair (approved):
I found inverting the ignores logic using with it being called 'test' a bit confusing at first but there's nothing wrong with it. ignore library has a function called test as well that operates differently, returns an object that says if the file was ignored by a rule and also if the file was subsequently unignored by a different rule.

Maybe in the future we add a flag to print a list of ignored files (and the ignore library rules as it parses them? `this.#ignore._rules` ) to help users debug.

PR #1480, opened 2022-06-28
[SCHEMA] tests: Add integration tests for summary and empty_files using example datasets

nellh:
Most of the new tests here are failing intentionally to provide useful targets for implementation.

### DIFF
--- a/bids-validator/src/tests/local/valid_headers.test.ts
+++ b/bids-validator/src/tests/local/valid_headers.test.ts
@@ -1,5 +1,5 @@
 // Deno runtime tests for tests/data/valid_headers
-import { assert } from '../../deps/asserts.ts'
+import { assert, assertEquals } from '../../deps/asserts.ts'
 import { validatePath } from './common.ts'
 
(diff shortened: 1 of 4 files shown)


bids-validator/src/tests/local/valid_headers.test.ts
```diff
@@ -8,6 +8,33 @@ Deno.test('valid_headers dataset', async (t) => {
...
+      'NOT_INCLUDED error should not be present',
+    )
+  })
+
+  await t.step('summary has correct tasks', () => {
+    assertEquals(result.summary.tasks, ['rhyme judgment'])
+  })
+
+  await t.step('summary has correct dataProcessed', () => {
+    assertEquals(result.summary.dataProcessed, ['rhyme judgment'])
```
rwblair, line 22: I think dataProcessed is a boolean, but can update this once the new summary is complete.

nellh: Yes, thanks for catching that.

PR #1481, opened 2022-06-29
Add summary generation to schema-prototype

rwblair:
Summary generated is an object literal being exported in src/summary/summary. May be better served as a class. file counts, modalities, and participant metadata seem to be working. Task name is reliant on parsing json files. Right now summaryUpdate code is trying to access the json field in the current context to see if they have a TaskName field.

Also have a some `error: AssertionError: Test case is leaking async ops.` to figure out.


bids-validator/src/summary/summary.ts
```diff
@@ -0,0 +1,132 @@
...
+    } else if (['acdyn', 'nacdyn'].includes(context.entities.rec)) {
+      secondaryModalitiesCount.PET_Dynamic++
+    }
+  }
+
+  if (context.file.path.includes('participants.tsv')) {
+    const stream = await context.file.stream
+    const streamReader = stream.getReader()
+    const denoReader = readerFromStreamReader(streamReader)
+    const fileBuffer = await readAll(denoReader)
```
nellh: The file object should probably just have a method for this, like .text() (fetch style). Otherwise we're going to do this all over the place. I like that this reads the stream though, portable to the browser!

bids-validator/src/validators/bids.ts
```diff
@@ -33,6 +33,7 @@ export async function validate(fileTree: FileTree): Promise<ValidationResult> {
       checkLabelFormat(schema, context)
     }
     applyRules(schema, context)
+    updateSummary(context)
```
nellh: To fix the async issue, looks like you just need to await this.

bids-validator/src/tests/local/common.ts
```diff
@@ -3,13 +3,17 @@ import { FileTree } from '../../types/filetree.ts'
...
 
 export async function validatePath(
   t: Deno.TestContext,
   path: string,
 ): Promise<{ tree: FileTree; result: ValidationResult }> {
   let tree: FileTree = new FileTree('', '')
-  let result: ValidationResult = { issues: new DatasetIssues(), summary: {} }
+  let result: ValidationResult = {
+    issues: new DatasetIssues(),
+    summary: formatSummary(summary),
```
rwblair: This seems silly. should just make summary a class, look more like issues.

bids-validator/src/files/deno.ts
```diff
@@ -79,7 +93,7 @@ export async function _readFileTree(
   const tree = new FileTreeDeno(relativePath, name, parent, rootPath)
 
   for await (const dirEntry of Deno.readDir(join(rootPath, relativePath))) {
-    if (dirEntry.isFile) {
+    if (dirEntry.isFile || dirEntry.isSymlink) {
```
rwblair, line 96: Running locally on an annexed dataset I had to add this to see any files. What kinds of problems could having this cause?

rwblair:
@nellh  Testing on ds000003 the deno summary matches current summary with the exception of size and datatypes. Current validator will attempt to use filename suffix as a datatype in some cases. For ds003 this means `inplaneT2` and `events` gets listed as a datatype. Can you think of anywhere that would rely on `events` being in the datatypes?

Deno correctly calculates the size where as current valdiator outputs 0. May have something to do with this: https://github.com/bids-standard/bids-validator/pull/1481/files#r912104701

Anything I should add or update?

bids-validator/src/files/deno.ts
```diff
@@ -108,6 +122,6 @@ export async function _readFileTree(
  * Read in the target directory structure and return a FileTree
  */
 export function readFileTree(rootPath: string): Promise<FileTree> {
-  const ignore = new FileIgnoreRulesDeno([])
+  const ignore = new FileIgnoreRulesDeno(['.git**'])
```
nellh: We could make this the default value for FileIgnoreRulesDeno so that other instantiations (tests, etc) also use this default.

bids-validator/src/summary/summary.ts
```diff
@@ -0,0 +1,167 @@
...
+    } else if (context.datatype === 'pet' && 'rec' in context.entities) {
+      if (['acstat', 'nacstat'].includes(context.entities.rec)) {
+        this.secondaryModalitiesCount.PET_Static++
+      } else if (['acdyn', 'nacdyn'].includes(context.entities.rec)) {
+        this.secondaryModalitiesCount.PET_Dynamic++
+      }
+    }
+
+    if (context.file.path.includes('participants.tsv')) {
+      let tsvContents = await context.file.text()
```
nellh:
```suggestion
      const tsvContents = await context.file.text()
```

bids-validator/src/summary/summary.ts
```diff
@@ -0,0 +1,167 @@
...
+      totalFiles: this.totalFiles,
+      size: this.size,
+      dataProcessed: this.dataProcessed,
+      pet: this.pet,
+      datatypes: Array.from(this.datatypes),
+    }
+  }
+}
+
+export const summary = new Summary()
```
nellh: I think it would be better to just export the class and construct it in validate() like the Issues class.

nellh:
> @nellh Testing on ds000003 the deno summary matches current summary with the exception of size and datatypes. Current validator will attempt to use filename suffix as a datatype in some cases. For ds003 this means `inplaneT2` and `events` gets listed as a datatype. Can you think of anywhere that would rely on `events` being in the datatypes?

At least on the OpenNeuro side, nothing depends on events being listed and I think we should stick to the more spec accurate output here.

nellh (changes requested):

nellh (approved):

Issue #1478, opened 2022-06-28
Populate context data


rwblair:
Context objects have a variety of properties referenced in schema rules via selectors/checks. We need functions that produce the these objects organized as outlined in https://github.com/bids-standard/bids-specification/blob/master/src/schema/meta/context.yaml. Some like path and entities are being populated as the contexts are created. The following are not and will need to be serialized from files:
- [ ]   sidecar
- [ ]   associations
- [ ]   columns
- [x]   json
- [ ]   nifti_header

The dataset and subjects properties will be populated when the functions to generate summary data is implemented.

rwblair:
For associations we'll need to add something like this to the schema:
https://github.com/bids-standard/bids-specification/issues/1047#issuecomment-1086654227

Reviewing the associations they only ones directly accessed are expecting json/tsv. But the associated json/tsv can have their own data files that would need to be asserted to exist at some point. (Don't think schema has any rules saying what kinds of metadata files are disallowed without a corresponding data file.)

If we parsed entity/suffix/extension in the initial fileTree load and exposed them through BIDSFileDeno we could load the existing associations and their data in the Context constructor.

rwblair:
Associations are currently being loaded synchronously on context loading.

PR #1487, opened 2022-07-11
[WIP] Function to load json sidecar and functions to evaluate different rule types.

rwblair:
In an effort to forward #1478  this loads a json sidecar with inheritance. Much more straightforward with the current fileTree object compared to old validator way. The call to load the sidecar is being done right before the call to applyRules, this is not the right place for it to happen. Its async but we maybe able to do it closer to context construction?

We currently have functions to evaluate selectors and checks with respect to a context. This add functions to interpret fields and columns entries. These should help with validating sidecars etc. I'm not in love with my implementation of some of these and am happy to throw them out for anything else.

In the schema we don't really have names for all the different types of rules and what all shapes they can take. 

todo:
- [x] Use issue codes and reason text directly from schema rules if they are present.
- [x] Come up with a reasonable way to generate issue codes for when the rule itself doesn't have ~~a rule~~ an issue.
- [ ] Fix typing, implemented a stop gap GenericSchema and GenericRule, and still had to use some `@ts-expect-error` decorators.
- [ ] Add tests
  - [x] For loadSidecar - inheritance in particular
  - [ ] For eval functions
- [ ] Find a dataset that really makes use of inheritance to play with.


bids-validator/src/schema/applyRules.ts
```diff
@@ -16,12 +32,185 @@ const safeHas = () => true
...
+    optional: 'ignore',
+    prohibited: 'ignore',
+  }
+  let severity: Severity = 'ignore'
+
+  if (typeof requirement === 'string' && requirement in levelToSeverity) {
+    severity = levelToSeverity[requirement]
+  } else if (typeof requirement === 'object' && requirement.level) {
+    severity = levelToSeverity[requirement.level]
+    const addendumRegex = /(required|recommended) if \`(\w+)\` is \`(\w+)\`/
```
rwblair, line 208: This probably won't work. The addendums should be rewritten to be selectors. As a regex the semantics is wonky and not as clear to the spec writer why something is or isn't working.

rwblair: We would have more rules, but the "fields" rules types would become way simpler.

rwblair: sidecars/asl.yaml is halfway there. It has rules with level_addendum, and then has a second rule with just the condition and required field. Forget where this was talked about....

bids-validator/src/schema/context.ts
```diff
@@ -62,4 +65,67 @@ export class BIDSContext implements Context {
...
+      .then((text) => parseTSV(text))
+      .catch((error) => {
+        console.log(error)
+        return {}
+      })
+    return
+  }
+
+  async loadAssociations(): Promise<void> {
+    this.associations = await buildAssociations(this.#fileTree, this)
```
rwblair, line 122: Associations are currently rarely used within the schema, and half of the defined associations could be rewritten to not use async and thus found/resolved at access time.

bids-validator/src/schema/associations.ts
```diff
@@ -0,0 +1,124 @@
...
+ * This object describes associated files for data files in a bids dataset
+ * For any given datafile we iterate over every key/value in this object.
+ * For each entry we see if any files in the datafiles directory have:
+ *   - a suffix that matches the key
+ *   - an extension in the entry's extension array.
+ *   - that all the files entities and their values match those of the datafile
+ * If the entry allows for inheritance we recurse up the filetree looking for other applicable files.
+ * The load functions are incomplete, some associations need to read data from a file so they're
+ * returning promises for now.
+ */
```
rwblair, line 19: We could add a selectors entry to this object to indicate when a search and load of an association should ever be performed.

bids-validator/src/schema/associations.ts
```diff
@@ -0,0 +1,124 @@
...
+      associationLookup[key as keyof typeof associationLookup]
+    const paths = getPaths(fileTree, source, key, extensions, inherit)
+    if (paths.length === 0) {
+      continue
+    }
+    if (paths.length > 1) {
+      // error?
+    }
+    // @ts-expect-error
+    associations[key] = await associationLookup[key].load(paths[0])
```
rwblair, line 88: I think to get around how messy the typing of this statement, especially the load, we could pass associations object down into load and let load be responsible for setting value.

PR #1500, opened 2022-08-11
Update regexp for blood pet to require the recording entity

### LINKED ISSUE #1499: Validator not checking pet blood data
We just uploaded https://openneuro.org/datasets/ds004230 that includes blood data, but for some reason the validator did not complain about the blood data not having the recording entity, which it should have? Looking at the code for the validator this is indeed present in the schema for the pet datatype, but why is it not complaining? 

@rwblair @bendhouseart

mnoergaard:
closes #1499 

This PR updates the regexp for blood pet data to have the recording entity required in the filename if blood data is present (https://github.com/bids-standard/bids-validator/issues/1499#issue-1336399775). The update is also in line with https://github.com/bids-standard/bids-specification/pull/1005/files, where I failed to identify an update to the validator back then.

### DIFF
--- a/bids-validator/bids_validator/rules/file_level_rules.json
+++ b/bids-validator/bids_validator/rules/file_level_rules.json
@@ -525,7 +525,7 @@
   },
 
   "pet_blood": {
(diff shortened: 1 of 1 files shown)


rwblair:
- `?:` starts a non-capturing group, which for how we use regex right now may not be necessary
- the `?` at the end of the capture group parentheses is what makes the group optional or required.

sappelhoff:
@mnoergaard does this fix work on your data as expected?

rwblair:
tested locally and it works.

mnoergaard:
Thank you @rwblair and @sappelhoff !

PR #1505, opened 2022-08-16
add leading slash to empty_file check path

rwblair:


rwblair:
parens around `t` in Deno test must of been added automatically.

Do we want the map to be keyed with leading slash?

nellh:
> parens around `t` in Deno test must of been added automatically.
> 
> Do we want the map to be keyed with leading slash?

Yeah, I think we do, it lets the file tree code break these paths out and not have to special case if it's an absolute or relative path in the dataset, since the first token being '/' means it is a full path.

nellh (approved):

PR #1506, opened 2022-08-17
[SCHEMA] Human readable output support

nellh:
* Fixes the structure of DatasetIssues to not require nesting issues.issues in some situations
* Brings over the human readable output
* Makes this default and enables the `--json` flag to get JSON output


rwblair:
Damn good work

PR #1507, opened 2022-08-17
[SCHEMA] implement some of the old bids spec tests

rwblair:
More failing tests, these error codes might not even exist in the new regime.

Also generate a complete list of issue codes that show up in schema to make sure we don't use any that don't exist in the new test.


rwblair:
This isn't too useful right now so we may want to hold off on merging it, or find a another way to orchestrate mass assertions against datasets.

PR #1535, opened 2022-09-23
Schema/expression language

rwblair:
- Introduces functions for schema expression language outlined [here](https://github.com/bids-standard/bids-specification/tree/master/src/schema#Expressions).
- ~~Adds nifti header loading but should revised. Causes a considerable slow down from loading whole nifti files synchronously.~~
- Adds fully loading events.tsv for associations checks.

### DIFF
--- a/bids-validator/src/files/nifti.ts
+++ b/bids-validator/src/files/nifti.ts
@@ -3,6 +3,7 @@ import { BIDSFileDeno } from './deno.ts'
 
 export function loadHeader(file: BIDSFileDeno) {
   const buf = file.readBytes(1024)
+  // @ts-expect-error
   const header = globalThis.nifti.readHeader(buf.buffer)
   // normalize between nifti-reader and spec schema
   // https://github.com/bids-standard/bids-specification/blob/master/src/schema/meta/context.yaml#L200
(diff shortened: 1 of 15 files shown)


rwblair:
Need to add tests. Expression testing examples could come from specification itself a la bids-standard/bids-specification#1274

bids-validator/src/files/nifti.ts
```diff
@@ -0,0 +1,11 @@
+import 'https://raw.githubusercontent.com/rii-mango/NIFTI-Reader-JS/master/release/current/nifti-reader-min.js'
+import { BIDSFileDeno } from './deno.ts'
+
+export function loadHeader(file: BIDSFile) {
```
nellh:
```suggestion
export function loadHeader(file: BIDSFileDeno) {
```

bids-validator/src/files/nifti.ts
```diff
@@ -0,0 +1,11 @@
...
+  const header = nifti.readHeader(Deno.readFileSync(file._getPath()).buffer)
```
nellh:
```suggestion
  const buf = file.readBytes(1024)
  const header = globalThis.nifti.readHeader(buf.buffer)
```

Added this in #1536 to keep this from requiring the private path method.

bids-validator/src/schema/context.ts
```diff
@@ -124,8 +147,11 @@ export class BIDSContext implements Context {
   }
 
   async asyncLoads() {
-    await this.loadSidecar()
-    await this.loadColumns()
-    await this.loadAssociations()
+    await Promise.all([
```
nellh:
```suggestion
    await Promise.allSettled([
```

nellh (changes requested):

nellh (approved):

PR #1547, opened 2022-10-12
Schema/derivatives walk

rwblair:
End goal is derivative schema being applied. Includes a rewrite of the filename validation rules to handle the schema reorganization and handle arbitrarily deeply nested file rules in the schema.

- [x] Run new validation on datasets in the /derivatives directory of a dataset that we are currently validating. 
  - Currently not printing the results of these sub validations. How its integrated into output can be saved for another PR
- [x] Reduce number of files output in summary, testing against existing derivatives this has become unwieldy.
  - By default only list 3 files per error, list all of them if --verbose is used
- [x] Suss out just how spec compliant the derivatives in bids-examples are.
  - Made issue for fmriprep output. Couple of issues in the schema itself that I'm making a PR for.
- [x] Figure out how to inject derivative status into contexts to prevent having to [delete parts of the schema](https://github.com/bids-standard/bids-validator/blob/ad1c60e3c6d87df0279cc86d5e3fc259964ee283/bids-validator/src/validators/bids.ts#L41)
  - Dataset description is parsed before walking the tree and passed to the dataset wide context. If DataType isn't set in it, we set it accordingly.
 - [ ] Figure out a solution to typescript issues across applyRules.ts and validateFilenames.ts. Two different sets of types for the schema are currently being used. One that explicitly lays out every object and its sub-objects and their sub-objects and so on, but is incomplete, a pain to keep up to date, and no fun to use. It does shut typescript up. The other, GenericSchema, which is a recursive type def and vague about what its children look like, and requires a lot of type guards to make go.
 - [ ] ~~Maybe make applyRules use period separated string paths (ex schema["rules.files.raw.func.bold"]) to access nested objects in the schema like valdiateFilenames does. This lets us have error messages that show the exact location of the failing rule in the schema. Typescript really does not like overriding objects base get call to do sketchy things (see bids-validator/src/utils/objectPathHandler.ts and subsequent proxy in bids-validator/src/types/schema.ts)~~ Current objectPathHandler implementation combined with recursive descent into schema was too slow. Maybe building a Map object of the terminal schema entries would be better.


bids-validator/src/validators/filenameIdentify.ts
```diff
@@ -0,0 +1,189 @@
...
+  findRuleMatches,
+  hasMatch,
+  cleanContext,
+]
+
+export async function filenameIdentify(schema, context) {
+  for (const check of CHECKS) {
+    await check(schema as unknown as GenericSchema, context)
+  }
+  return Promise.resolve()
```
nellh:
```suggestion
```

bids-validator/src/validators/filenameIdentify.ts
```diff
@@ -0,0 +1,189 @@
...
+}
+
+async function findRuleMatches(schema, context) {
+  let schemaPath = 'rules.files'
```
nellh:
```suggestion
  const schemaPath = 'rules.files'
```

bids-validator/src/validators/filenameIdentify.ts
```diff
@@ -0,0 +1,189 @@
...
+    // first directory must be subject
+  }
+  if (parts.length < 3) {
+    return Promise.resolve()
+  }
+  const sesParts = parts[2].split('-')
+  if (sesParts.length === 2 && sesParts[0] === sesEntity) {
+    datatypeIndex = 3
+  }
+  let dirDatatype = parts[datatypeIndex]
```
nellh:
```suggestion
  const dirDatatype = parts[datatypeIndex]
```

bids-validator/src/validators/filenameIdentify.ts
```diff
@@ -0,0 +1,189 @@
...
+]
+
+export async function filenameIdentify(schema, context) {
+  for (const check of CHECKS) {
+    await check(schema as unknown as GenericSchema, context)
+  }
+  return Promise.resolve()
+}
+
+async function findRuleMatches(schema, context) {
```
nellh:
```suggestion
function findRuleMatches(schema, context) {
```

bids-validator/src/validators/filenameIdentify.ts
```diff
@@ -0,0 +1,189 @@
...
+   */
+  if (context.filenameRules.length > 1) {
+    const entExtMatch = context.filenameRules.filter((rulePath) => {
+      return entitiesExtensionsInRule(schema, context, rulePath)
+    })
+    if (entExtMatch.length > 0) {
+      context.filenameRules = [entExtMatch[0]]
+    }
+  }
+  /* If we end up with multiple rules we should generate an error? */
```
nellh, line 136: Seems like this would be useful as a warning of some kind. Maybe we should have a debug flag that enables some validator implementation warnings instead of adding this as a validation warning?

bids-validator/src/validators/filenameIdentify.ts
```diff
@@ -0,0 +1,189 @@
...
+  for (let key in schema.rules.modalities) {
+    if (schema.rules.modalities[key].datatypes.includes(dirDatatype)) {
+      context.modality = key
+      context.datatype = dirDatatype
+      return Promise.resolve()
+    }
+  }
+}
+
+export async function hasMatch(schema, context) {
```
nellh:
```suggestion
export function hasMatch(schema, context) {
```

bids-validator/src/validators/filenameValidate.ts
```diff
@@ -0,0 +1,270 @@
...
+    hasProp(schema.objects, 'entities') &&
+    hasProp(schema.objects.entities, name)
+  ) {
+    const entityObj = schema.objects.entities[name]
+    if (hasProp(entityObj, 'name')) {
+      return entityObj.name
+    }
+  }
+  // if this happens there is an issue with the schema?
+  return ''
```
nellh, line 75: Seems like another good spot for --debug level warnings.

bids-validator/src/validators/filenameValidate.ts
```diff
@@ -0,0 +1,270 @@
...
+      if (!rePattern.test(label)) {
+        context.issues.addNonSchemaIssue('INVALID_ENTITY_LABEL', [
+          {
+            ...context.file,
+            evidence: `entity: ${fileEntity} label: ${label} pattern: ${pattern}`,
+          },
+        ])
+      }
+    } else {
+      // unknown entity
```
nellh, line 132: This should probably be a validation warning right?

nellh (approved):
Testing with various datasets this seems to be working pretty well. Some general feedback and minor JavaScript style tweaks suggested.

PR #1572, opened 2022-12-16
ENH: Check for empty or whitespace-only dataset names

effigies:
This is a rare case but the schema does not verify that `dataset_description.Name` contains non-whitespace characters. On OpenNeuro it led to UI issues where the dataset links were not rendered. It seems like an obviously bad practice to warn the user about, and OpenNeuro can turn it into an error if it's a validator issue.

xref openneuroorg/openneuro#2595

### DIFF
--- a/bids-validator/validators/bids/checkDatasetDescription.js
+++ b/bids-validator/validators/bids/checkDatasetDescription.js
@@ -37,7 +37,7 @@ const checkDatasetDescription = (jsonContentsDict) => {
 const checkNameField = (name) => {
   const issues = []
   // missing name will be caught by validation (later)
-  if (name !== null) {
+  if (name !== undefined) {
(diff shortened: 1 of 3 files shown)


bids-validator/validators/bids/checkDatasetDescription.js
```diff
@@ -33,6 +34,18 @@ const checkDatasetDescription = (jsonContentsDict) => {
   return issues
 }
 
+const checkNameField = (name) => {
+  const issues = []
+  // missing name will be caught by validation (later)
+  if (name !== null) {
```
nellh:
```suggestion
  if (name !== undefined) {
```
Name will be `undefined` here if it is missing from the datasetDescription object.

nellh (approved):

Issue #1573, opened 2022-12-23
Report unescaped backslashes in JSON sidecar files


effigies:
We've found a dataset on OpenNeuro with unescaped backslashes in their `bold.json` files. The dataset validates, but attempting to read the files with `jq` or Python's `json.load` results in errors.

Here's an example file that should fail:

```JSON
{"Field": "String with \ character"}
```

xref https://github.com/poldracklab/tacc-openneuro/issues/51

effigies:
Something has changed since that dataset validated on OpenNeuro. Now we crash the HED validator:

```console
$ bids-validator --ignoreSubjectConsistency /data/bids/ds003459/
bids-validator@1.9.9
(node:49452) Warning: Closing directory handle on garbage collection
(Use `node --trace-warnings ...` to show where the warning was created)
	1: [ERR] Internal error. SOME VALIDATION STEPS MAY NOT HAVE OCCURRED (code: 0 - INTERNAL ERROR)
		Evidence: TypeError: Cannot convert undefined or null to object
    at Function.entries (<anonymous>)
    at BidsSidecar.filterHedStrings (/usr/lib/node_modules/bids-validator/dist/commonjs/cli.js:135498:43)
    at new BidsSidecar (/usr/lib/node_modules/bids-validator/dist/commonjs/cli.js:135494:18)
    at /usr/lib/node_modules/bids-validator/dist/commonjs/cli.js:153338:12
    at Array.map (<anonymous>)
    at constructSidecarData (/usr/lib/node_modules/bids-validator/dist/commonjs/cli.js:153337:30)
    at checkHedStrings (/usr/lib/node_modules/bids-validator/dist/commonjs/cli.js:153303:23)
    at Object.events_default (/usr/lib/node_modules/bids-validator/dist/commonjs/cli.js:153380:10)
    at /usr/lib/node_modules/bids-validator/dist/commonjs/cli.js:154256:28

	Please visit https://neurostars.org/search?q=INTERNAL ERROR for existing conversations about this issue.

	1: [WARN] Not all subjects/sessions/runs have the same scanning parameters. (code: 39 - INCONSISTENT_PARAMETERS)
		./sub-203/func/sub-203_task-visortho_acq-a_bold.nii.gz
		./sub-263/func/sub-263_task-audphono_acq-b_bold.nii.gz
		./sub-269/func/sub-269_task-visortho_acq-a_bold.nii.gz

[...]

PR #1574, opened 2022-12-27
FIX: Use intersection to find existing events.json files

effigies:
Resolves HED failure reported in #1573.

### DIFF
--- a/bids-validator/tests/events.spec.js
+++ b/bids-validator/tests/events.spec.js
@@ -355,9 +355,8 @@ describe('Events', function () {
         jsonFiles,
         '',
       ).then((issues) => {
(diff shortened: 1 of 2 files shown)


bids-validator/tests/events.spec.js
```diff
@@ -355,9 +355,8 @@ describe('Events', function () {
         jsonFiles,
         '',
       ).then((issues) => {
-        assert.strictEqual(issues.length, 2)
+        assert.strictEqual(issues.length, 1)
```
effigies, line 358: Looking at this test, I couldn't see why this used to be 2. Seems like it could have been a meaningless duplicate, so this fix seems sensible, but would want confirmation.

Issue #1579, opened 2023-01-04
BUG: Specify schema version with `-s "v1.6.0"` option is failing


sebastientourbier:
Dear all,

I tried to use the BIDS validator `1.9.9` with the option `-s "v1.6.0"` to use a specific schema version. However, it is failing with the error below.

**Error**

```output
Error: ENOENT: no such file or directory, open 'https://raw.githubusercontent.com/bids-standard/bids-specification/v1.6.0/src/schema/rules/top_level_files.yaml
```

**Reason**

I took some time to investigate the problem and found the cause: the `rules` folder was indeed created in `v1.7.0` but it is not existing in `v1.6.0`. So, in other words, the correct paths would be for:

- `v1.7.0`:
  - [https://raw.githubusercontent.com/bids-standard/bids-specification/**v1.7.0**/src/schema/**rules/top_level_files.yaml**](https://raw.githubusercontent.com/bids-standard/bids-specification/v1.7.0/src/schema/rules/top_level_files.yaml)
  - [https://raw.githubusercontent.com/bids-standard/bids-specification/**v1.7.0**/src/schema/**rules/entities.yaml**](https://raw.githubusercontent.com/bids-standard/bids-specification/v1.7.0/src/schema/rules/entities.yaml)

- `v1.6.0`:
  - [https://raw.githubusercontent.com/bids-standard/bids-specification/**v1.6.0**/src/schema/**top_level_files.yaml**](https://raw.githubusercontent.com/bids-standard/bids-specification/v1.6.0/src/schema/top_level_files.yaml)
  - [https://raw.githubusercontent.com/bids-standard/bids-specification/**v1.6.0**/src/schema/rules/**entities.yaml**](https://raw.githubusercontent.com/bids-standard/bids-specification/v1.6.0/src/schema/entities.yaml)
 
**Possible fixes**

While I would be happy to help and provide a fix, but before taking the initiative, I would like to hear your feedback. Here are two easy fix options that came to my mind to make this beta feature working as expected:

- **option 1:** we keep `v1.6.0` in the choices (in code shown for **option 1** but we remove `rules/` (code shown in **option 2**)
- **option 2:** change `'v1.6.0'` to `'v1.7.0'` in: 
    https://github.com/bids-standard/bids-validator/blob/bdf2e72a0a51ffda49ae02384127d05d70bd757b/bids-validator/validators/options.js#L69-L74
[...]

PR #1580, opened 2023-01-06
FIX/FEAT: Use specific schema version

sebastientourbier:
This PR addresses the following issue:
- #1579

I took the initiative to go to the option solution 2 (make the `--schema` option working for `v1.6.0` and `v1.7.0`).

It integrates the following changes:

- [x] FIX: Correct prefix paths to schema files for version `v1.6.0` as the `rules/` folder has been introduced in `v1.7.0`.
- [x] FEAT/REF: Handle schema files for version `v1.7.0` where entities in the `entities.yaml` have been re-ordered alphabetically.
- [x] FEAT: Add the option choices both `v1.6.0` and `v1.7.0`.

*Note:* I made an attempt to support `v1.8.0` as well  but this was involving quite some refactoring in `generateRegex()` as it introduced new schema files specific to `channels`, `nirs`, `task`, and `photo`. I guess this would deserve its own specific PR. What do you think?

### DIFF
--- a/bids-validator/validators/options.js
+++ b/bids-validator/validators/options.js
@@ -69,9 +69,9 @@ export function parseOptions(argumentOverride) {
     .option('schema', {
       alias: 's',
       describe:
-        'BIDS specification schema version to use for validation, e.g. "v1.7.0" (beta)',
+        'BIDS specification schema version to use for validation, e.g. "v1.6.0" (beta)',
       default: 'disable',
-      choices: ['disable', 'v1.7.0', 'master'],
+      choices: ['disable', 'v1.6.0', 'v1.7.0', 'master'],
     })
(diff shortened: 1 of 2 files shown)


sappelhoff:
Thanks for looking into this @sebastientourbier -- the fix seems fine to me. From my view, we can merge it ... but it's important to keep in mind that this is bleeding edge stuff that shouldn't be relied upon.

Personally, I'd like to also be able to specify a link to a particular schema, for example to this one:

- https://bids-specification--1319.org.readthedocs.build/en/1319/schema.json

... which is the schema based on a particular PR to `bids-specification`, but maybe that should be its own issue.

What are your thoughts @rwblair @effigies ?

PR #1583, opened 2023-01-12
BF: close open nifti file after parsing the header

yarikoptic:
reincarnated #691 

Closes  may be #675

### DIFF
--- a/bids-validator/utils/files/readNiftiHeader.js
+++ b/bids-validator/utils/files/readNiftiHeader.js
@@ -51,11 +51,15 @@ function extractNiftiFile(file, callback) {
     } else {
       fs.read(fd, buffer, 0, bytesRead, 0, function () {
         if (file.name.endsWith('.nii')) {
-          callback(parseNIfTIHeader(buffer, file))
+          const header = parseNIfTIHeader(buffer, file)
+          fs.close(fd)
+          callback(header)
         } else {
           try {
             const data = pako.inflate(buffer)
-            callback(parseNIfTIHeader(data, file))
+            const header = parseNIfTIHeader(data, file)
+            fs.close(fd)
+            callback(header)
           } catch (err) {
             callback(handleGunzipError(buffer, file))
           }


yarikoptic:
- good news -- test invocation of bids-validator on my dataset didn't crash somehow new way
- bad news --  I am afraid this particular fix was not sufficient my case of #675 , I still got

```
Singularity> less -R /tmp/bids-validator-fixed.log
bids-validator@1.9.10-dev.0
...
        2: [ERR] We were unable to read this file. Make sure it contains data (file size > 0 kB) and is not corrupted, incorrectly named, or incorrectly symlinked. (code: 44 - FILE_READ)
                ./sub-56/ses-1/fmap/sub-56_ses-1_acq-mb3_dir-pa_run-01_epi.nii.gz
                ./sub-56/ses-1/func/sub-56_ses-1_task-movie_run-04_echo-1_bold.nii.gz
                ./sub-56/ses-1/func/sub-56_ses-1_task-movie_run-04_echo-2_bold.nii.gz
                ./sub-56/ses-1/func/sub-56_ses-1_task-movie_run-04_echo-2_sbref.nii.gz
                ./sub-56/ses-1/func/sub-56_ses-1_task-movie_run-04_echo-3_bold.nii.gz
                ./sub-56/ses-1/func/sub-56_ses-1_task-movie_run-04_echo-3_sbref.nii.gz
                ./sub-56/ses-1/func/sub-56_ses-1_task-movie_run-05_echo-1_bold.nii.gz
                ./sub-56/ses-1/func/sub-56_ses-1_task-movie_run-05_echo-2_bold.nii.gz
                ./sub-56/ses-1/func/sub-56_ses-1_task-movie_run-05_echo-2_sbref.nii.gz
                ./sub-56/ses-1/func/sub-56_ses-1_task-movie_run-05_echo-3_bold.nii.gz
                ... and 2482 more files having this issue (Use --verbose to see them all).
```

effigies:
Is it possible that there's a bug in node wrt symlinks that causes unclosed file handles? What if we manually readlink and then open the linked file directly?

bids-validator/utils/files/readNiftiHeader.js
```diff
@@ -51,11 +51,15 @@ function extractNiftiFile(file, callback) {
     } else {
       fs.read(fd, buffer, 0, bytesRead, 0, function () {
         if (file.name.endsWith('.nii')) {
```
effigies, line 53: Since the buffer is read in L52, can we not immediately close it here, rather than after descending in the call stack?

```suggestion
        fs.close(fd)
        if (file.name.endsWith('.nii')) {
```

yarikoptic: I am not familiar with the codebase enough to give judgement, is there a unittest for this code - I could then blindly accept ;)

effigies: Yes, this code will be thoroughly tested. The proposed change would be instead of your current commit, not in addition to.

effigies: @nellh Might know off-hand if this will fail, though...

effigies:
@yarikoptic Are you running `lsof` during these runs and seeing if it's NIFTI files that are accumulating? Even if this worked, if we were still accumulating JSON file handles, this might still exhibit in this way.

yarikoptic:
> @yarikoptic Are you running `lsof` during these runs and seeing if it's NIFTI files that are accumulating? Even if this worked, if we were still accumulating JSON file handles, this might still exhibit in this way.

I remember looking at something... don't remember what -- could have been just `/bin/ls -l /proc/{pid}/fd/ | wc -l` or alike.  I have vague memory that it was low for quite a while and then grew rapidly at some later point in time (if I am not mixing things up)

yarikoptic:
should be superseded by #1590 or something which supersedes that one ;)

PR #1600, opened 2023-02-02
[SCHEMA] Add a development script entrypoint with deno options

nellh:
This is some initial fixes to get the browser build closer to working but the main change is to add a script entrypoint for CLI development usage.

The script is very simple for now but it could attempt loading the validator locally and remotely once we have a build published somewhere so that deno and this one file is all you need.


rwblair:
@nellh we had discussed calling it bids-validator-dev. I feel like that may still be ambiguous for people who don't know whats what so I called it bids-validator-deno instead, is this ok?

PR #1611, opened 2023-02-07
Add bids-examples test to schema based validator.

rwblair:
To run from bids-validator/bids-validator:
`deno test --allow-write --allow-env --allow-read src/tests/local/bids_examples.ts`

generates a file called `bids_examples_issues.tsv` in calling directory.

### DIFF
--- a/bids-validator/src/deps/asserts.ts
+++ b/bids-validator/src/deps/asserts.ts
@@ -4,4 +4,4 @@ export {
   assertObjectMatch,
   assertExists,
   assertRejects,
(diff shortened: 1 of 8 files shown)


rwblair:
Not adding in this PR github actions, validator wide ignore issue, fixes to make more bids-examples pass.

Table only prints first found filename of provided error code.

rwblair:
There's a bug with colors and cliffy. When max width for a column is set, and a cell that gets wrapped is using colors, That cells color bleeds for the rest of the row's cells, (or until another color is set i'd guess)

nellh (approved):

PR #1637, opened 2023-03-02
Avoid crashing when reporting config file errors

mguaypaq:
When there's a config file error (for example, invalid JSON syntax in `.bids-validator-config.json`), the `issues` only has a `.config` member, and no `.errors` or `.warnings` members. So, we need to check whether `issues.errors` is defined before taking `issues.errors.length`. Otherwise, we get an unhandled rejection traceback, instead of the desired diagnostic output:

```
bids-validator@1.10.0
[ERR]  Invalid Config File
	1: [ERR] Unexpected comma. (code: 27 - JSON_INVALID)
		..bids-validator-config.json

	Please visit https://neurostars.org/search?q=JSON_INVALID for existing conversations about this issue.
```

### DIFF
--- a/bids-validator/utils/consoleFormat.js
+++ b/bids-validator/utils/consoleFormat.js
@@ -18,7 +18,7 @@ function formatIssues(issues, options = {}) {
   var errors = issues.errors
   var warnings = issues.warnings
   var output = []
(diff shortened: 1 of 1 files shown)


effigies (approved):
Looks right. @rwblair?

PR #1644, opened 2023-03-22
CI: Reenable coverage for Deno tests due to hang

effigies:
Tracking whether we can restore coverage without hanging. We should probably bump this ~weekly with a comment.

~~I believe this depends on https://github.com/denoland/deno/pull/16959 being merged.~~

Reverts bids-standard/bids-validator#1643.


effigies:
This is now failing in 1.5 minutes instead of timing out at 5.

Issue #1639, opened 2023-03-03
[SCHEMA] Integrate HED validator


nellh:
Taking a look at how the [HED JavaScript validator is called](https://github.com/bids-standard/bids-validator/blob/master/bids-validator/validators/events/hed.js), we should find a way to provide the events and sidecar data required following schema validation, then call the HED validator. The returned events issues will need to be adapted to the new Issue type to be included in the results.

effigies:
Done in #1648.

PR #1648, opened 2023-03-27
HED Validator integration into bids schema based validator.

rwblair:
For #1639.

~~bids-validator/src/deps/hed-validator` dependency was generated by running `npm run build` on latest commit for hed-javascript:~~
~~https://github.com/hed-standard/hed-javascript/commit/bba927b7b713f6aad7864b9aabc8b9afe70456cf~~

~~Two main functions located in `bids-validator/src/validators/hed.ts` are `hedAccumulator` and `hedValidate`~~

~~`hedAccumulator` is called once per context like other checks (filename validation, schema checks) in `bids-validator/src/validators/bids.ts`. It catches any events.tsv files and json files and builds up an object to be passed to the actual validator call.~~

~~Instead of recreating the old data structures that were passed into `bids-validator/validators/events/hed.js` I tried to get everything in shape a bit closer to where the outgoing calls were made, not sure if I got it quite right. The contexts have the proper sidecar data in them already so no need to mess with `potentialSidecars`.~~

~~`hedValidate` is called once, after the context walk is complete.~~

- [x] copy over the hed-validator issue to bids-validator issue functions, modify to work with new issue objects.
- [x] Add new issues to `bids-validator/src/issues/list.ts` as needed.
- [x] Copy over `detectHed` function to skip calls when no HED data present.
- [x] Pin dependency to actual release of hed-validator.
- [x] Add tests

~~Currently am getting `"ERROR: [HED_SCHEMA_LOAD_FAILED] The supplied schema specification is invalid. Specification: no sche..."` when run on `bids-examples/eeg_ds003645s_hed`~~

Latest commits incorporate the hed validator api changes and follow the lead of https://github.com/bids-standard/bids-validator/pull/2014

Validation is done per file as a context level check instead of original accumulator/dataset level check.


VisLab:
@rwblair --- should we be nervous that this is PR is still a draft?  Is there anything that we need to prepare for?  Thx! VisLab

VisLab:
I know you have been swamped, but could you give us an update on this @nellh? Thx.

nellh (approved):
I think this is ready to merge. I ran through some test cases and was able to match the failures to the legacy validator. ds004395 runs out of memory with the default heap size but does pass increasing the size to 8GB. There could be some memory optimization to be done because this doesn't happen without the HED patches but that dataset is an exception, every other test case passed with the default heap size.

VisLab:
@nellh ... thx for update.   I am not sure what is going on with ds004395
-- I looked at it on openNeuro and it doesn't have any HED in it.  The new
HED validator works directly from the list of .tsv files and their
associated sidecars as produced from the BIDS side.  I wonder if we are
somehow copying the data inadvertently before determining whether to even
look at the files.
  @Alexander Jones ***@***.***> let's look at this more closely.

Version 3.13.3 should be released in the next day or so. It has some minor
bug fixes.

rwblair:
~~Looks like the memory issue is coming from code I wrote, via the hedAccumulator in the bids-validator. Commenting out the call to the hed-validator and to the file accumulator allows ds004395 to finish validation. Commenting out just the call to the hed-validator but leaving the accumulator active causes it to run out of memory on default heap size.~~

I should finish profiling before running my mouth.

happy5214:
I'm not sure what's going on with the Deno build. The lines in the errors do not appear in the version of that file in our code base, and we don't use `Buffer` objects directly (though some dependencies might).

happy5214:
@rwblair Has there been any further progress on this?

rwblair:
@happy5214 Merged master to get it up to date. The out of memory issue is from keeping an object around with all the sideccar and tsv data in memory while the file tree is walked. I think I was mirroring how the old validator called the hed validator. I'm going to try to:

1. Detect which files have hed data at the accumulator step, to keep the arguments for hedValidator.validator.BidsDataset leaner.
2. Have the accumulator only track file names and then reload tsv and json data at hed validation time after hopefully all the per bids file context objects have been garbage collected.
3. Generate a hedValidator.validator.BidsDataset for each events file and all of its associated sidecars. Are there any HED rules that require multiple tsv files to be loaded at the same time for validation? or is it only that we need to ensure all the sidecars are present for each separate tsv file?

VisLab:
> Are there any HED rules that require multiple tsv files to be loaded at the same time for validation? or is it only that we need to ensure all the sidecars are present for each separate tsv file?

HED processes each tsv and its sidecars separately and does not need to have multiple event files at one time.

@happy5214 will need to address the details of the other issues.  Thanks for looking at this -- we'd like to get this resolved.

rwblair:
I rewrote bids-validator/src/validators/hed.ts to work on a file per context basis in line with the 3rd option in my last post. This allows ds004395 to complete validation.

I need to rewrite/add tests for this way of invoking HED validation.

VisLab (commented):
Thx!    In the longer term, we'll probably want to build the HED schema object from the dataset description before validation and pass it in as it doesn't change.  We'll get back to you on this. @happy5214 comment?

happy5214:
> Thx! In the longer term, we'll probably want to build the HED schema object from the dataset description before validation and pass it in as it doesn't change. We'll get back to you on this. @happy5214 comment?

To be honest, hearing that is a little annoying, because that's how it was *originally* implemented, and then we implemented a design philosophy to move as much of the code as possible to hed-validator. The schema-loading function is still exported (`hedValidator.validator.buildSchemas`), but there's nowhere to pass it right now in the external API in order to run BIDS validation. I'll discuss more with @VisLab when we meet today.

rwblair:
For now its still passing the dataset_description to each validation call for each file. Not sure how inefficient this is CPU wise, but the memory issue is gone for now.

VisLab:
That is fine for now.   We will work on the underlying implementation.

bids-validator/src/validators/bids.ts
```diff
@@ -15,17 +15,21 @@ import { emptyFile } from './internal/emptyFile.ts'
...
 /**
  * Ordering of checks to apply
  */
-const CHECKS: CheckFunction[] = [
+const perContextChecks: ContextCheckFunction[] = [
   emptyFile,
   filenameIdentify,
   filenameValidate,
   applyRules,
+  hedValidate
```
effigies:
```suggestion
  hedValidate,
```

happy5214:
I believe the failing test is because of an off-by-one error (you're converting the TSV data to the old format, which you shouldn't (`BidsTsvFile`'s constructor now expects either a string or the new `ColumnMap`-style format), but you're forgetting to add the header as the first row).

bids-validator/LICENSE
effigies, line 1: This is still needed for Python packaging. It can be removed with the rest of the package after the next release.

bids-validator/src/validators/hed.ts
```diff
@@ -0,0 +1,117 @@
...
+  schema: GenericSchema,
+  context: BIDSContext,
+) {
+  const sidecarFile = new hedValidator.bids.BidsSidecar(
+    context.path,
+    context.json,
+    context.file,
+  )
+  return sidecarFile
+}
```
effigies: Unused parameter (schema):

```suggestion
async function buildHedTsvFile(context: BIDSContext)  {
  const tsvData = await context.file.text()
  const eventFile = new hedValidator.bids.BidsTsvFile(
    context.path,
    tsvData,
    context.file,
    [],
    context.sidecar,
  )
  return eventFile
}

function buildHedSidecarFile(context: BIDSContext) {
  const sidecarFile = new hedValidator.bids.BidsSidecar(
    context.path,
    context.json,
    context.file,
  )
  return sidecarFile
}
```

bids-validator/src/validators/hed.ts
```diff
@@ -0,0 +1,117 @@
...
+    if (context.extension == '.tsv' && context.columns) {
+      if (!('HED' in context.columns) && !sidecarHasHed(context.sidecar)) {
+        return
+      }
+      hedValidationIssues = await setHedSchemas(context.dataset.dataset_description)
+
+      file = await buildHedTsvFile(schema, context)
+    } else if (context.extension == '.json' && sidecarHasHed(context.json)) {
+      hedValidationIssues = hedValidationIssues = await setHedSchemas(context.dataset.dataset_description)
+      file = buildHedSidecarFile(schema, context)
```
effigies:
```suggestion
      file = await buildHedTsvFile(context)
    } else if (context.extension == '.json' && sidecarHasHed(context.json)) {
      hedValidationIssues = hedValidationIssues = await setHedSchemas(context.dataset.dataset_description)
      file = buildHedSidecarFile(context)
```

effigies (approved):
Overall looks reasonable.

bids-validator/LICENSE
rwblair: Not sure why I had removed it to begin with...

bids-validator/src/validators/hed.ts
```diff
@@ -0,0 +1,111 @@
...
+      )
+    }
+  })
+}
+
+async function buildHedTsvFile(context: BIDSContext)  {
+  const tsvData = await context.file.text()
+  const eventFile = new hedValidator.bids.BidsTsvFile(
+    context.path,
+    tsvData,
```
happy5214: Would passing the `ColumnMap` (or a copy) work here, or would that expose too much internal data? The parameter type in hed-validator is `Map<string, string[]>`, which appears to be `ColumnMap`'s parent type.

rwblair: Using the columns map here generates an error:
`TypeError: Cannot read properties of undefined (reading 'forEach')`

This is coming from `convertParsedTSVData`:
https://github.com/hed-standard/hed-javascript/blob/3be833f929cd3db5e259b315ba6cff6e1c677f59/bids/tsvParser.js#L45
via:
https://github.com/hed-standard/hed-javascript/blob/3be833f929cd3db5e259b315ba6cff6e1c677f59/bids/types/tsv.js#L53

I think this is because Map objects pass the check meant to catch the old style tsv data:
```
> let x = new Map()
undefined
> x === Object(x)
true
```

effigies: @happy5214 Does it make sense to go ahead as-is and then submit a PR once hed-validator can handle a `ColumnMap` directly?

happy5214: #2041 should fix that issue. The type check was a bug that proved easy to fix.

PR #1655, opened 2023-04-10
Fix issues in expression language

rwblair:
type now returns null if operand is undefined.

Length works on arrays and strings.

### DIFF
--- a/bids-validator/src/schema/expressionLanguage.ts
+++ b/bids-validator/src/schema/expressionLanguage.ts
@@ -10,6 +10,9 @@ export const expressionFunctions = {
     if (Array.isArray(operand)) {
       return 'array'
     }
+    if (typeof operand === 'undefined') {
+      return 'null'
+    }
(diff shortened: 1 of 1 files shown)


nellh (approved):
Tested and this fixes everything I was seeing except for the incorrect duplicate files issue.

rwblair:
`rules.checks.general.DuplicateFiles` depends on `exists` and `substr` in the expression language.  Current typing on it makes it look like its looking for an element in a list both of which are passed to it, while the rule is using it to check for the existence of a file.

I'm not sure how to make exists work from where it [lives at the moment](https://github.com/bids-standard/bids-validator/blob/master/bids-validator/src/schema/expressionLanguage.ts#L27) if its the latter.

effigies:
Can we pass in the context as an implicit parameter?

rwblair:
wow I totally forgot... yeah it should be there. 💀

PR #1669, opened 2023-05-03
Bundle CLI entrypoint with main() call

nellh:
This adds a new `bids-validator.js` build output that can be run with Deno without any additional code. This avoids the need to inject a caller like `await main()` and lets you run the validator directly from the raw script URL.

I wasn't sure if I should name it bids-validator.js or something more or less descriptive. If you download this file, it's fairly intuitive to run it with this name: `deno run ./bids-validator.js`

### DIFF
--- a/.github/workflows/deno_tests.yml
+++ b/.github/workflows/deno_tests.yml
@@ -74,7 +74,8 @@ jobs:
       - name: Commit to new branch
         run: |
           mv main/main.js .
-          git add main.js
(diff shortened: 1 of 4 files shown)


rwblair (approved):
Only thing I could think of is putting bids-validator.ts in a `bin` directory. I could imagine thinking it would be silly to have a directory with a single file in it. I don't know how I feel about it.

PR #1675, opened 2023-05-17
fix(schema): Improve performance with memoization for evalConstructor

nellh:
This saves a decent amount of time constructing the evalConstructor function when the same function has already been parsed.

On a large dataset with 200k files, before:
```
real    4m14.717s
user    4m15.705s
sys     0m6.777s
```

After:
```
real    3m44.549s
user    3m46.593s
sys     0m6.151s
```


nellh:
I've also added related memoization for readEntities and a performance test for it. This is another 20 seconds or so off the time for this dataset.

rwblair:
Looks great!

PR #1430, opened 2022-03-17
Migrate web site from Next to Vite

rwblair:
Help match the openneuro build environment more closely.

I didn't need to use any of the workarounds in https://github.com/OpenNeuroOrg/openneuro/blob/master/packages/openneuro-app/vite.config.js to allow vite build to go off. Mosttly stayed with defaults from vite.

PR #1685, opened 2023-06-14
Bump hed-validator dependency to v3.9.0

happy5214:
This version bundles the SCORE library schema and includes support for `Onset` and `Offset` tag validation.


sappelhoff:
Thanks, the failure in the githubPagesTest seems to be related to HED, at first glance.

happy5214:
It doesn't seem to like the instance variable declarations. Do you have a solution that doesn't involve a complete rewrite of the hed-validator classes? (Honestly, I'd rather learn TypeScript and rewrite the whole codebase in that than strip the classes of their instance variable declarations.)

rwblair:
I may be able to fix this by upgrading the nextjs dependency for the site. Will push to this PR when I get something that works.

sappelhoff:
> I may be able to fix this by upgrading the nextjs dependency for the site. Will push to this PR when I get something that works.

is that related to

- #1422
- #1430 

?

VisLab:
We'd really appreciate some help on this -- v 3.10.0 of the hed-validator has just been released -- supported partnered schema and there are some other key updates.  Thanks!

happy5214:
I just pushed an update to the `package.json` and `package-lock.json` to the PR.

sappelhoff:
cc @nellh @rwblair

rwblair:
I haven't been able to figure out why exactly the web build is broken. Once I clear away any commonjs directories that webpack is finding the web build errors become about babel plugins not being importable from parts of the validator code. 

I have a partially working commit with nextjs upgraded, which requires webpack5. It also requires an upgraded react which requires some of our components to be fixed. I'm seeing this web build work ok in firefox but not in chrome:
https://github.com/rwblair/bids-validator/tree/bump-hed-validator-3.9.0

There are issues with the module reload used by nextjs. It doesn't play well with workspaces, its trying to load `module.hot` from bids-validator which isn't usable in the browser, but it also doesn't know to use `import.meta.webpackHot` instead which is provided. (And if we aren't going to have hot reloading of the validator in the web dev server maybe we should just switch to vite, with module reloading being the sticking point there.)

rwblair:
@happy5214 @VisLab With the help of @nellh we got the web builds upgraded and working, along with the new hed-validator. Thank you for your patience.

VisLab:
Thanks to all --- we appreciate the help!

PR #1689, opened 2023-06-16
CI: Test Python over supported range

effigies:
Also set supported range to currently alive Pythons. (3.7 is EOL within days if not already.)


effigies:
This is triggered by seeing bids-validator in a [list of packages that Fedora had to patch](https://discuss.python.org/t/experience-with-python-3-11-in-fedora/12911) to support Python 3.11. The actual issue was versioneer, which was upgraded a while back, but we should be testing anyway.

PR #1705, opened 2023-06-29
fix jest tests that broke from earlier upgraes.

rwblair:


.circleci/config.yml
```diff
@@ -81,7 +81,7 @@ jobs:
             - ./image.tar
   githubPagesTest:
     docker:
-      - image: mhart/alpine-node:14
+      - image: node:18
```
effigies: Looks like we still need apk for rsync.

```suggestion
      - image: node:18-alpine
```

effigies:
@rwblair Built the docker image locally and got it fixed up. Some small cleanups to ci. Pushed both to your branch, and I think this is good to go, if you want to have a quick look.

sappelhoff (approved):

PR #1710, opened 2023-06-30
Return 1 on most failures in cli try/except

rwblair:
bids-examples run_tests.sh was returning 0 when using an old version of node was causing validator to crash.


rwblair:
https://github.com/rwblair/bids-validator/actions/runs/5425420299/jobs/9866141135

PR #1738, opened 2023-07-14
FIX: Permit electrodes.tsv in meg directories

effigies:
Implements https://github.com/bids-standard/bids-specification/pull/1555 for the legacy validator. RECOMMENDED status can wait on the schema.


sappelhoff (changes requested):
placing this as a visual block, as we are still waiting for:

- https://github.com/bids-standard/bids-specification/pull/1555

sappelhoff (approved):
see reasoning in:

- https://github.com/bids-standard/bids-specification/issues/1550#issuecomment-1686032272

PR #1770, opened 2023-08-04
DOC: Add usage instructions for running from deno.land

effigies:
The current README is overly developer-focused. This adds docs that will appear on the deno.land landing page (https://deno.land/x/bids_validator) that tell users how to use what's there.

### DIFF
--- a/bids-validator/src/README.md
+++ b/bids-validator/src/README.md
@@ -8,9 +8,25 @@ Deno is a JavaScript and TypeScript runtime that is used to run the schema based
 
 At the root of the repository there are two directories, `bids-validator` and `bids-validator-web`. These are separate npm packages, the Deno validator lives within the bids-validator package within the `src` directory.
 
-## Schema validator examples
+## Usage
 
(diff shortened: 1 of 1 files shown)


sappelhoff (approved):
thanks!

bids-validator/src/README.md
```diff
@@ -8,9 +8,25 @@ Deno is a JavaScript and TypeScript runtime that is used to run the schema based
...
-Deno by default sandboxes applications like a web browser. To validate datasets located on your local system, you need to use the --allow-read flag to read local files. --allow-env is also required to allow for detection of OS specific features. These flags are included in the script line of the `./bids-validator-deno` script.
+To use the latest validator hosted at https://deno.land/x/bids_validator, use the following command:
+
+```console
+$ deno run --allow-read --allow-env https://deno.land/x/bids_validator/bids-validator.ts path/to/dataset
+```
+
+Deno by default sandboxes applications like a web browser. `--allow-read` allows the validator to read local files, and `--allow-env` enables OS-specific features.
+
+By default only schema derived validation rules are run. The legacy validator can be run at the same time with `--legacy` and this may improve coverage for issues not yet implemented in the schema variant.
```
nellh: This is stale, the legacy option no longer works and we should remove the references to it.
```suggestion
```

nellh (changes requested):
Looks good except we should update the --legacy option docs since it was removed.

nellh (approved):

PR #1772, opened 2023-08-08
[WIP] Participants data always null. Issue 49 never thrown.

### LINKED ISSUE #1755: Participant data not being returned
I'm currently trying to add support for HED validation in `participants.tsv` files, and I'm running into an issue getting the participant data passed to HED. First of all, it seems the `participants` object doesn't include the full row, but I can't even use that data because I'm just getting `null` when I try to access it (the main point of this issue). I've tried it on the `fmri_ds002790s_hed_aomic` hed-examples dataset, which clearly has a `participants.tsv`, to no avail.

rwblair:
- closes #1755 

~~Participants was being defined as a const and assigned null. Usual use case for consts is to give it an actual array or object with reference and then later on push to that array or object. For whatever reason tsv validation attempts to reassign the participants reference it gets.~~


rwblair:
We now get code 67  `NO_VALID_DATA_FOUND_FOR_SUBJECT` for any bids-examples with a participants.tsv.

`ieeg_motorMiller2007` has rows in participants.tsv that have no directories in the dataset. Throws code 49 `PARTICIPANT_ID_MISMATCH`

sappelhoff:
> ieeg_motorMiller2007 has rows in participants.tsv that have no directories in the dataset. Throws code 49 PARTICIPANT_ID_MISMATCH

@dorahermes could you look into this, please?

rwblair:
Suggestion from maintainers meeting: for legacy validator automatically add 67 to ignored codes if --ignore-nifti-headers used, prevent us from having to edit ignored codes in bids-examples tests.

dorahermes:
It seems that 3 subjects are missing folders (jf, gf, jp). Data from these subjects is located in the original [ECoG library](https://exhibits.stanford.edu/data/catalog/zk881ps0522). Let me know if you prefer option 1 or 2:

1. Remove these three subjects from the BIDS example participants.tsv file 
2. I could convert these 3 subjects to BIDS and do a pull request to update accordingly

(I am not sure why I missed these subjects in the conversion. One thing I vaguely remember is that there was a mismatch in electrodes and channels, but the current version of the ECoG library does not have any mismatches. There was an update in 2022 though.)

effigies:
IMO if the usefulness of the dataset as an example isn't damaged by dropping the subjects, removing them from `participants.tsv` seems sensible to me.

sappelhoff:
> IMO if the usefulness of the dataset as an example isn't damaged by dropping the subjects, removing them from participants.tsv seems sensible to me.

agreed

PR #1776, opened 2023-08-18
Reorganize HED wrapper

happy5214:
This PR allows for supporting HED validation in TSV and sidecar file types other than events. `hed.js` is moved out of `validators/events` as a result. hed-validator is bumped to 3.12.0 to include the new class `BidsTabularFile`, which provides this support. Also, all object data is deep-cloned before being passed to hed-validator, in order to prevent accidental data corruption.


VisLab:
This looks good.  @rwblair Could you merge if there are no issues.  Thx

PR #1796, opened 2023-08-31
Fix context.columns typescript issues

nellh:
Implement ColumnsMap class for preserving ordered keys when creating a parseTSV object. ColumnsMap can be constructed to return a proxy which allows the underlying map to be accessed with a property accessor (`map[key]`) like it is a regular object. A limitation of the proxy is that it only allows key access and iteration, other methods on the underlying map are not passed through and some object functions will not work.

### DIFF
--- a/bids-validator/src/schema/applyRules.ts
+++ b/bids-validator/src/schema/applyRules.ts
@@ -211,7 +211,7 @@ function evalColumns(
       ])
     }
     if (headers.includes(name)) {
-      for (const value of context.columns[name]) {
+      for (const value of context.columns.get(name) as string[]) {
         if (
           !schemaObjectTypeCheck(columnObject as SchemaTypeLike, value, schema)
         ) {
@@ -316,11 +316,11 @@ function evalIndexColumns(
     ])
     return
   }
-  const rowCount = context.columns[index_columns[0]].length
+  const rowCount = context.columns.get(index_columns[0])?.length || 0
   for (let i = 0; i < rowCount; i++) {
     let indexValue = ''
     index_columns.map((col: string) => {
-      indexValue = indexValue.concat(context.columns[col][i])
+      indexValue = indexValue.concat(context.columns.get(col)?.[i] || '')
     })
     if (uniqueIndexValues.has(indexValue)) {
       context.issues.addNonSchemaIssue('TSV_INDEX_VALUE_NOT_UNIQUE', [
--- a/bids-validator/src/schema/applyRules.ts
+++ b/bids-validator/src/schema/applyRules.ts
@@ -197,7 +197,7 @@ function evalColumns(
   schemaPath: string,
 ): void {
   if (!rule.columns || context.extension !== '.tsv') return
-  const headers = [...Object.keys(context.columns)]
+  const headers = [...Map.prototype.keys.bind(context.columns)()]
   for (const [ruleHeader, requirement] of Object.entries(rule.columns)) {
     // @ts-expect-error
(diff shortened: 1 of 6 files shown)


bids-validator/src/types/columns.test.ts
```diff
@@ -0,0 +1,19 @@
...
+    columns[2] = ['1']
+    columns[1] = ['2']
+    columns['b'] = ['3']
+
+    // Check that the order is still correct with iteration
+    let iteration = 0
+    for (const [_, val] of columns) {
+      assertEquals(val, [iteration.toString()])
+      iteration += 1
+    }
```
effigies, line 22: Do we also have property access, or is it just square-bracket access `[]`?

```suggestion
    }

    // Check property access
    assertEquals(columns.a, '0')
```

Is there a way to get property access, if not?

nellh: Square bracket and dot access to properties is equivalent in JavaScript but it did point out a bug with how the name is looked up. aba7edc includes a fix and your suggested test change.

effigies (approved):
I don't fully understand `columns.ts`, but everything else LGTM.

effigies:
Oh, we are getting a lot of:

```
 error: TypeError: context.columns.keys is not a function or its return value is not iterable
  const headers = [...context.columns.keys()]
```

Do we need to find an alternative way to do that, or can we implement `keys`?

nellh:
> Oh, we are getting a lot of:
> 
> ```
>  error: TypeError: context.columns.keys is not a function or its return value is not iterable
>   const headers = [...context.columns.keys()]
> ```
> 
> Do we need to find an alternative way to do that, or can we implement `keys`?

Keys can be passed from the underlying map. Let me fix that...

bids-validator/src/schema/applyRules.ts
rwblair, line 1: Line 200 is failing for me:
`error: Uncaught (in promise) TypeError: context.columns.keys is not a function or its return value is not iterable`

nellh:
Added support for columns.keys() and a test for that but that breaks if you have a "keys" column, fixed it instead by changing to Object.keys(columns).

bids-validator/src/types/columns.ts
```diff
@@ -0,0 +1,24 @@
...
+    super()
+    const columns = new Map<string, string[]>() as ColumnsMap
+    return new Proxy<ColumnsMap>(columns, columnMapAccessorProxy)
+  }
+}
+
+// Proxy handler to implement ColumnsMapType
+export const columnMapAccessorProxy = {
+  get: function (target: ColumnsMap, prop: any) {
+    if (prop === Symbol.iterator) {
```
rwblair: I had to add a `Reflect.get(target, prop)` and test return of that for typeof function then bind it to get calls to `keys` working. Not sure if there is a more elegant approach.

effigies:
The failing tests are the expected failures at this point.

rwblair:
I can't get object.keys to work:
```
> import { ColumnsMap } from './columns.ts'
undefined
> let x = new ColumnsMap()
undefined
> x.a = 3
3
> Object.keys(x)
[]
```

nellh:
> I can't get object.keys to work:
> 
> ```
> > import { ColumnsMap } from './columns.ts'
> undefined
> > let x = new ColumnsMap()
> undefined
> > x.a = 3
> 3
> > Object.keys(x)
> []
> ```

Fixed that with the right prototype method, we could also just add a function with a less likely to conflict name for this.

Issue #1783, opened 2023-08-24
Add cordsystem to associations for optodes.tsv.


rwblair:
https://github.com/bids-standard/bids-specification/blob/02d97b0c75b57e5f1a528253d4348fa236b115b1/src/schema/rules/checks/nirs.yaml#L91

failing for https://github.com/bids-standard/bids-examples/blob/master/fnirs_tapping/sub-02/nirs/sub-02_optodes.tsv

effigies:
Validator implementation side. Defined in meta.context and meta.associations.

effigies:
Fixed in #1797

PR #1797, opened 2023-09-01
ENH: Finish populating associations

effigies:
This depends on Nell's #1796, but so does everything.

### DIFF
--- a/bids-validator/src/schema/associations.ts
+++ b/bids-validator/src/schema/associations.ts
@@ -79,7 +79,7 @@ const associationLookup = {
       const columns = parseBval(contents)
       return {
         path: file.path,
-        n_cols: parseBval.length,
+        n_cols: columns.length,
       }
     },
   },
@@ -104,6 +104,7 @@ const associationLookup = {
       return {
         path: file.path,
         type: columns.get('type') || [],
+        short_channel: columns.get('short_channel') || [],
       }
     },
   },
--- a/bids-validator/src/schema/associations.ts
+++ b/bids-validator/src/schema/associations.ts
@@ -24,6 +24,7 @@ import { parseBval, parseBvec } from '../files/dwi.ts'
  */
 const associationLookup = {
   events: {
+    suffix: 'events',
     extensions: ['.tsv'],
     inherit: true,
     load: async (file: BIDSFile): Promise<ContextAssociations['events']> => {
@@ -36,6 +37,7 @@ const associationLookup = {
     },
   },
   aslcontext: {
+    suffix: 'aslcontext',
     extensions: ['.tsv'],
     inherit: true,
     load: async (
@@ -51,6 +53,7 @@ const associationLookup = {
     },
   },
   m0scan: {
+    suffix: 'm0scan',
     extensions: ['.nii', '.nii.gz'],
     inherit: false,
     load: (file: BIDSFile): Promise<ContextAssociations['m0scan']> => {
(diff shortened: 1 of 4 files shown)


bids-validator/src/schema/associations.ts
```diff
@@ -60,26 +72,46 @@ const associationLookup = {
...
+    extensions: ['.bval'],
     inherit: true,
-    load: (file: BIDSFile): Promise<ContextAssociations['bval']> => {
-      return Promise.resolve({ path: file.path, n_cols: 0 })
+    load: async (file: BIDSFile): Promise<ContextAssociations['bval']> => {
+      const contents = await file.text()
+      const columns = parseBval(contents)
+      return {
+        path: file.path,
+        n_cols: parseBval.length,
```
rwblair: `columns.length`?

bids-validator/src/schema/associations.ts
```diff
@@ -60,26 +72,46 @@ const associationLookup = {
     },
   },
   bval: {
```
rwblair, line 79: bvals and bvecs aren't being loaded because this key is being used as the targetSuffix to determine which files are valid paths. bval and bvec never appearing as a suffix.

Quickest way to load them is to add something like this to buildAssociations:
`
    let ts = key
    if (key == 'bval' || key == 'bvec') {
      ts = 'dwi'
    }
    const paths = getPaths(fileTree, source, ts, extensions)
`

In the case of ds114 with bvals and bvecs in root I think it loads these as an association for every file.

We could change the key to be `dwi` and load both files at the same time, but then we would need to update the schema rules to access them appropriately. Reference in checks from `associations.bval.n_cols` to this `associations.dwi.bval.n_cols`

Or there should be a way to encode this information better in the associationLookup object itself.

bids-validator/src/schema/associations.ts
```diff
@@ -60,26 +72,46 @@ const associationLookup = {
...
-    load: (file: BIDSFile): Promise<ContextAssociations['events']> => {
-      return file
-        .text()
-        .then((text) => parseTSV(text) as ContextAssociationsEvents)
+    load: async (file: BIDSFile): Promise<ContextAssociations['channels']> => {
+      const contents = await file.text()
+      const columns = parseTSV(contents)
+      return {
+        path: file.path,
+        type: columns.get('type') || [],
```
rwblair, line 114: One check for channels requires a `short_channel` field in associations, this is probably a mistake in the schema.
https://github.com/bids-standard/bids-specification/pull/1598

bids-validator/src/schema/associations.ts
```diff
@@ -60,26 +72,46 @@ const associationLookup = {
...
+      }
+    },
+  },
+  coordsystem: {
```
rwblair, line 119: Looks good, fixes  #1783

bids-validator/src/schema/associations.ts
rwblair: Can be for a different issue, but the contexts for bvals and bvecs themselves will need custom logic in context.ts `loadColumns`, currently only works for files with the `.tsv` extension.

bids-validator/src/schema/associations.ts
effigies: Yeah, we need to add `short_channel` to `meta.context`

bids-validator/src/schema/associations.ts
effigies: Just added a `suffix` property to `associationLookups`, and that should cover these cases and any others that crop up in the future.

> In the case of ds114 with bvals and bvecs in root I think it loads these as an association for every file.

Just need to implement `meta.associations` filtering.

effigies:
@rwblair No rush, just want to make sure you know this is ready for review.

PR #1810, opened 2023-09-20
Reduce bids-validator-web pack size.

rwblair:
npm pack uses gitignore relative to the workspace directory. Removed leading path for .next in gitignore.


effigies:
@rwblair Is this good to go?

Issue #1821, opened 2023-10-02
[BUG]? Python bids-validator returns true on non-bids folder names 


bendhouseart:
Was playing around a bit last week and noticed that the python version of the validator seems to misclassify non-bids folder names as compliant.

See:

```python
>>> from bids_validator import BIDSValidator
>>> validator = BIDSValidator()
>>> validator.is_bids('/sub_07/eeg/sub-07_task-matchingpennies_events.tsv')
True
>>>
```

Python and validator version(s):
```python
Python 3.10.9 (main, Mar  1 2023, 12:20:14) [Clang 14.0.6 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import bids_validator
>>> bids_validator.__version__
'1.12.0'
```

effigies:
I'm a little inclined to mark this wontfix. If you would like to do filename validation in Python, the bidsschematools validator is going to be better supported.

effigies:
Actually, I think the better fix is to depend on `bidsschematools` and write shims for the current API. We should strip the Python validator out of the JS validator repository and free up JS developers to reorganize without potentially breaking the Python packaging.

sappelhoff:
+1 to remove the python validator from the JS repo and potentially even deprecating it in favor of the new schema based python validator.

effigies:
This will be fixed with #1967 
```python
In [1]: from bids_validator import BIDSValidator

In [2]: BIDSValidator().is_bids('/sub_07/eeg/sub-07_task-matchingpennies_events.tsv')
Out[2]: False
```

PR #1822, opened 2023-10-04
MNT: Update Python package configuration to use pyproject.toml

effigies:
This PR is one in a series to factor out the Python validator from the Javascript validator.

Here we update the configuration to use `pyproject.toml`, which will simplify any transitions we want to make away from setuptools+versioneer. We also are able to use `versioneer` as a build-time dependency instead of vendored, reducing a bit of noise in the repository.

A future PR will add `bidsschematools` as a dependency and replace the current functions with shims around that.


effigies:
Could I request a review from a Pythonista? Pinging @bids-standard/maintainers, but anybody should be able to review this.

sappelhoff (approved):
LGTM, thanks!

- xref: https://github.com/bids-standard/bids-validator/issues/1821#issuecomment-1743312143

PR #1823, opened 2023-10-04
FIX: Un-define the unspecified Context.data

effigies:
Introduced in #1797, but broken and made unnecessary by #1813 and bids-standard/bids-specification#1622.

This should fix one of the failing CI tests.


effigies:
@rwblair I think you're best placed to review this one.

PR #1837, opened 2023-11-02
ENH: Enable structured Level descriptions for TSV columns

effigies:
Follow-up to bids-standard/bids-specification#1603. Adds support for levels in legacy validator.

xref https://github.com/bids-standard/bids-examples/pull/413

### DIFF
--- a/bids-validator/validators/json/schemas/data_dictionary.json
+++ b/bids-validator/validators/json/schemas/data_dictionary.json
@@ -27,7 +27,21 @@
           "description": "For  categorical variables: a dictionary of possible values (keys) and their descriptions (values).",
           "patternProperties": {
             "^.+$": {
-              "type": "string"
+              "anyOf": [
+                { "type": "string" },
+                {
+                  "type": "object",
+                  "properties": {
+                    "Description": { "type": "string" },
+                    "TermURL": {
+                      "title": "TermURL",
+                      "description": "URL pointing to a formal definition in an ontology available on the web.",
+                      "type": "string",
+                      "format": "uri"
+                    }
+                  }
+                }
+              ]
             }
           }
         },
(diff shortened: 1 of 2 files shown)


sappelhoff (approved):
Thanks @effigies

sappelhoff:
One additional thought:

MOTION allows for this:

```json
{
  "reference_frame": {
    "Levels": {
      "global": {
        "SpatialAxes": "ALS",
        "RotationOrder": "ZXY",
        "RotationRule": "right-hand"
      },
      "local": {
        "Description": "Joint angles are described following [...]"
      }
    }
  }
}
```

([source](https://github.com/bids-standard/bids-specification/blob/47d912de10634f3baa9bbf36aff1f8a1d06597c2/src/modality-specific-files/motion.md?plain=1#L220-L235))

Would we have to add:

- `SpatialAxes`
- `RotationOrder`
- `RotationRule`

to the general `data_dictionary.json` JSON schema (`Description` and `TermURL` are permitted via this PR)? Because this general `data_dictionary.json` would be called upon for a potential `channels.json` file in a dataset, right? (because we don't have a specific channels.json schema)

Is there something in the BIDS schema we should adjust for this?

So far in the BIDS spec, we have a markdown table for the above metadata: https://github.com/bids-standard/bids-specification/blob/47d912de10634f3baa9bbf36aff1f8a1d06597c2/src/modality-specific-files/motion.md?plain=1#L211-L216

And the rows in that table have individual entries in the schema, for example: https://github.com/bids-standard/bids-specification/blob/47d912de10634f3baa9bbf36aff1f8a1d06597c2/src/schema/objects/metadata.yaml#L2725-L2736

... but I see no schemafied "rules" how in a `channels.json` you can have something like in the JSON example above.

sappelhoff:
opened a new issue for this. Merging here -- thanks.

PR #1839, opened 2023-11-03
ENH: Add column definitions for channels.json

### LINKED ISSUE #1838: motion: adding SpatialAxes etc to JSON schema
MOTION allows for this:

```json
{
  "reference_frame": {
    "Levels": {
      "global": {
        "SpatialAxes": "ALS",
        "RotationOrder": "ZXY",
        "RotationRule": "right-hand"
      },
      "local": {
        "Description": "Joint angles are described following [...]"
      }
    }
  }
}
```

([source](https://github.com/bids-standard/bids-specification/blob/47d912de10634f3baa9bbf36aff1f8a1d06597c2/src/modality-specific-files/motion.md?plain=1#L220-L235))

Would we have to add:

- `SpatialAxes`
- `RotationOrder`
- `RotationRule`

to the general `data_dictionary.json` JSON schema (`Description` and `TermURL` are permitted via this PR)? Because this general `data_dictionary.json` would be called upon for a potential `channels.json` file in a dataset, right? (because we don't have a specific channels.json schema)

Is there something in the BIDS schema we should adjust for this?

So far in the BIDS spec, we have a markdown table for the above metadata: https://github.com/bids-standard/bids-specification/blob/47d912de10634f3baa9bbf36aff1f8a1d06597c2/src/modality-specific-files/motion.md?plain=1#L211-L216

And the rows in that table have individual entries in the schema, for example: https://github.com/bids-standard/bids-specification/blob/47d912de10634f3baa9bbf36aff1f8a1d06597c2/src/schema/objects/metadata.yaml#L2725-L2736

... but I see no schemafied "rules" how in a `channels.json` you can have something like in the JSON example above.

_Originally posted by @sappelhoff in https://github.com/bids-standard/bids-validator/issues/1837#issuecomment-1792159597_

effigies:
Closes #1838.


effigies:
Do we have examples to test against yet?

sappelhoff:
> Do we have examples to test against yet?

We do, see:

https://github.com/bids-standard/bids-examples/blob/b6e52348ab2b917a417906f0811c98827e4da26a/motion_dualtask/sub-07/ses-walk/motion/sub-07_ses-walk_task-dualWalking_tracksys-PhaseSpace1_channels.json#L1-L14

added in:

- https://github.com/bids-standard/bids-examples/pull/418

effigies:
Yup, found it. Now I just need to figure out JSON schema. This is a pain.

effigies:
Some notes for anybody who wants to take this over (or me in the future):

* Test with `npm run test bids-validator/tests/bids.spec.js`
* The problem with it as it stands is that you can't override the references like I wanted to. `{"$ref": "X", "override": "value"}` needs to become `{"allOf": [{"$ref": "X"}, {"override": "value"}]}` or something. Except that doesn't seem to successfully override.
* I think what needs to be done is to write the whole thing out without references and see what can be shortened by using references and what we're stuck with duplicating.

sappelhoff (approved):
thanks for the start, Chris, and thanks for the fix, Ross.

effigies:
Changing `bids-validator/tests/data/bids-examples/motion_dualtask/sub-07/ses-stand/motion/sub-07_ses-stand_task-dualWalking_tracksys-PhaseSpace1_channels.json` to have invalid fields does not fail the validator:

```json
{
        "reference_frame":{
                "LongName":"reference_frame",
                "Description":"reference frame in which the channels is represented.",
                "Levels":{
                        "global":{
                                "Description":"room-fixed global reference frame",
                                "SpatialAxes":"ARS",
                                "RotationOrder":1,
                                "RotationRule":1
                        }
                }
        }
}
```

sappelhoff (approved):
Checked Chris' example and also manipulating the values for the levels --> raises errors as expected now.

effigies:
Interesting. I'm not sure why my using `npm test` was failing to find this error. Explicitly running the validator on the broken example did work.

PR #1841, opened 2023-11-09
FIX: Pass dataset-relative paths to .bidsignore checker

effigies:
It looks like ignore wants dataset-relative paths, but we collect them with a leading slash. I'm not sure what the splash damage of removing the leading slash globally is, so instead I'm just dropping it in the ignore check.

Also dropping a vendored ignore package because apparently that's no longer needed.


effigies:
@nellh Are we still doing the `build.ts`? Or does publishing to deno.land make that irrelevant?

nellh:
It's still useful to build for Node targets like the OpenNeuro frontend.

effigies:
Sounds good. Just spent a bit of time digging to resolve:

```
Could not resolve "npm:ignore@5.2.4"

    src/files/ignore.ts:2:34:
      2 │ import { default as ignore } from 'npm:ignore@5.2.4'
        ╵                                   ~~~~~~~~~~~~~~~~~~

  You can mark the path "npm:ignore@5.2.4" as external to exclude it from the bundle, which will remove this error.
```

Was curious if I'd wasted my time. :-)

PR #1869, opened 2023-12-20
Bump hed-validator to 3.13.1

happy5214:
This PR updates the hed-validator dependency to 3.13.1 and resolves the compatibility issue raised in #1867. This should replace that PR.

### DIFF
--- a/bids-validator/package.json
+++ b/bids-validator/package.json
@@ -45,7 +45,7 @@
     "date-fns": "^2.30.0",
     "events": "^3.3.0",
     "exifreader": "^4.17.0",
-    "hed-validator": "^3.12.0",
+    "hed-validator": "^3.13.1",
     "ignore": "^5.3.0",
     "is-utf8": "^0.2.1",
     "jest": "^29.7.0",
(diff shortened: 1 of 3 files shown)


sappelhoff:
thanks!

There seems to be a failure in the HED validation of one example: https://github.com/bids-standard/bids-validator/actions/runs/7281187870/job/19841192263?pr=1869#step:8:1692

(`eeg_ds003645s_hed_library`) 

cc @VisLab

VisLab:
> There seems to be a failure in the HED validation of one example: https://github.com/bids-standard/bids-validator/actions/runs/7281187870/job/19841192263?pr=1869#step:8:1692
> 
> (`eeg_ds003645s_hed_library`)
> 
> cc @VisLab

The error message here should have been better on the hed-validator side (@happy5214 we should look at this) --- but the problem was that the `dataset_description` file seemed to have been mistakenly changed to the one from eeg_ds003645s_hed.  The JSON `events.json` sidecar was also from that dataset.  Sorry that happened.  I have done the PR [https://github.com/bids-standard/bids-examples/pull/421](https://github.com/bids-standard/bids-examples/pull/421) to correct.

Issue #1875, opened 2024-01-03
HED column in tsv file should not give a warning


VisLab:
bids-validator 1.14.0 gives a warning if a HED column appears in a `.tsv` file other than an `events.tsv` file:

>**Example:**    3: [WARN] Tabular file contains custom columns not described in a data dictionary (code: 82 - CUSTOM_COLUMN_WITHOUT_DESCRIPTION)
                ./participants.tsv
                        Evidence: Columns: HED not defined, please define in: /participants.json

HED can be usefully included in any `.tsv` file, but particularly `participants.tsv` and `scans.tsv`, and it is our understanding that the BIDS specification allows this.  However, HED cannot syntactically appear as a top-level key in the JSON file without triggering an error.

**Would it be possible to exclude HED from the listing of missing custom column definitions?**

P.S.  HED currently in the bids-validator appears to validate the `participants.tsv` and `scans.tsv` files okay. We are still working on `motion`, `phenotype` and `samples` related `.tsv` files. 

Our demo example is available at: [eeg_ds003645s_hed_demo](https://github.com/hed-standard/hed-examples/tree/develop/datasets/eeg_ds003645s_hed_demo). Once everything is working, we'll be able to replace a number of HED examples on `bids-examples` with a single demo that shows a lot of different uses.

  @sappelhoff @effigies @rwblair @happy5214

sappelhoff:
> However, HED cannot syntactically appear as a top-level key in the JSON file without triggering an error.

I don't understand this point. If you would include an `HED` column in a `participants.tsv` file, why would you not be able to place a `HED` key inside a `participants.json` file?

I opened https://github.com/bids-standard/bids-validator/pull/1876 anyhow, to show how the request here could be fulfilled.

VisLab:
The warning refers to HED as a column name.  This would require a json file:

{
    "HED": "Blah blah"
}

but the syntax for HED in BIDS requires that HED be at the same level as
"Levels" under a column name:

{
     "Column_name1":  {
           "Levels":  { .....}
           "HED": { .....}
     }
     "Column_name2": {
          "HED":  "Label/#"
      }
}

If the first example is used and "HED" appears at the first level, the HED
validator will rightfully detect it as an error.

sappelhoff:
Thanks for the explanation @VisLab! may you please review #1876 then?

PR #1876, opened 2024-01-04
add HED as optional column for other TSV files

sappelhoff:
- some steps towards: https://github.com/bids-standard/bids-validator/issues/1875

### DIFF
--- a/bids-validator/bids_validator/tsv/non_custom_columns.json
+++ b/bids-validator/bids_validator/tsv/non_custom_columns.json
@@ -23,7 +23,8 @@
     "tracking_system",
     "tracked_point",
     "component",
-    "placement"
+    "placement",
+    "HED"
   ],
   "electrodes": [
     "name",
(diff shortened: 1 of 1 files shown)


VisLab (commented):
This looks good --- except I don't see `samples.tsv` and `samples.json` in this list. Thx!

VisLab:
Also, I am wondering about the phenotype entry.  The phenotype is a
directory and the requirement is that phenotype/*.tsv should have a
participant_id. Thx

sappelhoff:
> Also, I am wondering about the phenotype entry.  The phenotype is a directory and the requirement is that phenotype/*.tsv should have a participant_id. 

You are right about this. There is no `phenotype.tsv` ... only a `phenotype/<whatever-name>.tsv`.

I will merge this here and open a new issue.

sappelhoff:
> This looks good --- except I don't see samples.tsv and samples.json in this list. Thx!

- fixes here: https://github.com/bids-standard/bids-validator/pull/1896

PR #1891, opened 2024-02-08
Skip checkheader for motion.tsv in legacy validator

### LINKED ISSUE #1889: motion.tsv files raise warnings and can raise errors due to a lack of column headers
via a collaborator who is preparing a dataset with motion data, the following issues were identified by the validator:

> 1. Warning - related to the fact that the *motion.tsv files have numeric column headers. In reality these files don't have any header.
> 3. Error - related to the fact that one *motion.tsv file had the same value in the first row for two different columns. This is just a coincidence and is related to (1). 

The examples don't exhibit it because they are empty. We should also update the examples to have a couple lines of data.

rwblair:
fixes #1889

### DIFF
--- a/bids-validator/validators/tsv/tsv.js
+++ b/bids-validator/validators/tsv/tsv.js
@@ -55,7 +55,10 @@ const TSV = (file, contents, fileList, callback) => {
   let emptyCells = false
   let NACells = false
 
-  checkHeaders(headers, file, issues)
+  // motion tsvs don't have headers
+  if (!file.name.endsWith('_motion.tsv')) {
+    checkHeaders(headers, file, issues)
+  }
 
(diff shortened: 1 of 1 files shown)


effigies (approved):
I think this is okay, but note that parsing this data is very inefficient. We can have files that are hundreds of MB.

If we need to improve treatment in the legacy validator, I think the more thorough fix would be to treat `motion.tsv` files as special in the same way that we split out continuous recordings here:

https://github.com/bids-standard/bids-validator/blob/5942782a1175c24f1d70dc3eba8edc97df81158f/bids-validator/validators/bids/groupFileTypes.js#L57-L58

Then we just parse the first line and verify that it has the same number of columns as an associated `channels.tsv` has rows.

Issue #1366, opened 2021-10-14
Permit bidsignore'd files in scans.tsv


effigies:
There are unstandardized files that might nevertheless be appropriate to put in a scans.tsv. By putting them in `.bidsignore`, we get  `[ERR] The filename in scans.tsv file does not match what is present in the BIDS dataset. (code: 129 - SCANS_FILENAME_NOT_MATCH_DATASET)`, and thus have to ignore `scans.tsv` as well.

appukuttan-shailesh:
Facing the same problem. I am willing to have a go at implementing this if it is approved that we can ignore files in `scans.tsv` that correspond to filters in `.bidsignore`

effigies:
This may be done. Adding a test case should resolve it.

PR #48, opened 2015-11-17
[WIP] EchoTimeDifference becomes replaced with EchoTime1 and EchoTime2

chrisgorgo:

PR #1914, opened 2024-03-09
Ignore files in _scans.tsv that correspond to entries in .bidsignore (#1366)

appukuttan-shailesh:
PR for handling bids-standard/bids-validator#48 

Makes use of session storage to save the contents of .bidsignore when it is read originally. This is retrieved later while evaluating _scans.tsv files for ignoring the matching files. All tests with `npm run test` are passing (except two which were failing even on the unchanged master branch), and two new tests have been added. 

For testing purposes, a mock session storage was required to be implemented as Jest is currently configured with `"testEnvironment": "node"`. Could possibly be avoided if set to `"testEnvironment": "jsdom"`. Another option is to employ `jest-localstorage-mock` package. But the current mock implementation seemed the simplest, and so went with it. 

Two tests have been added:
- `should ignore files in scans.tsv that correspond to entries in .bidsignore`
This checks that files to be ignored, based on .bidsignore, are being ignored when evaluating the file list in _scans.tsv

- `should not allow missing files listed in scans.tsv and not accounted for by .bidsignore`
This catches files that are listed in _scans.tsv but are actually missing, and not accounted for by .bidsignore

First PR here; apologies if I have missed anything.
@effigies: Could you take a look?

### DIFF
--- a/bids-validator/utils/files/readDir.js
+++ b/bids-validator/utils/files/readDir.js
@@ -5,6 +5,29 @@ import fs from 'fs'
 import * as child_proccess from 'child_process'
 import isNode from '../isNode'
 
+function createSessionStorageMock() {
+  const sessionStorage = {}
+
+  return {
+    getItem: (key) => sessionStorage[key],
+    setItem: (key, value) => {
+      sessionStorage[key] = value
+    },
+    removeItem: (key) => {
+      delete sessionStorage[key]
+    },
+    clear: () => {
+      Object.keys(sessionStorage).forEach((key) =>
+        sessionStorage.removeItem(key),
+      )
+    },
+  }
+}
+
+const sessionStorage = isNode
+  ? createSessionStorageMock()
+  : window.sessionStorage
+
 /**
  * Read Directory
  *
--- a/bids-validator/utils/files/readDir.js
+++ b/bids-validator/utils/files/readDir.js
@@ -4,29 +4,9 @@ import path from 'path'
 import fs from 'fs'
 import * as child_proccess from 'child_process'
 import isNode from '../isNode'
+import getSessionStorage from '../getSessionStorage'
 
-function createSessionStorageMock() {
-  const sessionStorage = {}
-
-  return {
-    getItem: (key) => sessionStorage[key],
-    setItem: (key, value) => {
-      sessionStorage[key] = value
-    },
-    removeItem: (key) => {
-      delete sessionStorage[key]
-    },
-    clear: () => {
-      Object.keys(sessionStorage).forEach((key) =>
-        sessionStorage.removeItem(key),
-      )
-    },
-  }
-}
-
-const sessionStorage = isNode
-  ? createSessionStorageMock()
-  : window.sessionStorage
+const sessionStorage = isNode ? getSessionStorage() : window.sessionStorage
 
 /**
  * Read Directory
--- a/bids-validator/utils/files/readDir.js
+++ b/bids-validator/utils/files/readDir.js
@@ -331,7 +331,7 @@ async function getBIDSIgnore(dir) {
     const content = await readFile(bidsIgnoreFileObj)
     ig.add(content)
     // Store the .bidsignore content in session storage
-    sessionStorage.setItem('bidsignoreContent', JSON.stringify(content));
+    sessionStorage.setItem('bidsignoreContent', JSON.stringify(content))
   }
   return ig
 }
--- a/bids-validator/utils/files/readDir.js
+++ b/bids-validator/utils/files/readDir.js
@@ -330,6 +330,8 @@ async function getBIDSIgnore(dir) {
   if (bidsIgnoreFileObj) {
     const content = await readFile(bidsIgnoreFileObj)
     ig.add(content)
+    // Store the .bidsignore content in session storage
+    sessionStorage.setItem('bidsignoreContent', JSON.stringify(content));
   }
   return ig
 }
--- a/bids-validator/validators/tsv/tsv.js
+++ b/bids-validator/validators/tsv/tsv.js
@@ -614,8 +614,8 @@ const TSV = (file, contents, fileList, callback) => {
       )
     } else {
       // Retrieve the .bidsignore content (if any) from session storage
-      const content = sessionStorage.getItem('bidsignoreContent');
-      const ig = content ? ignore().add(JSON.parse(content)) : null;
+      const content = sessionStorage.getItem('bidsignoreContent')
+      const ig = content ? ignore().add(JSON.parse(content)) : null
 
       // check scans filenames match pathList
       const filenameColumn = headers.indexOf('filename')
@@ -625,7 +625,7 @@ const TSV = (file, contents, fileList, callback) => {
         const scanFullPath = scanDirPath + '/' + scanRelativePath
 
         // check if file should be ignored based on .bidsignore content
-        if (ig && ig.ignores((path.relative('/', scanRelativePath)))) {
+        if (ig && ig.ignores(path.relative('/', scanRelativePath))) {
           continue
         }
 
--- a/bids-validator/validators/tsv/tsv.js
+++ b/bids-validator/validators/tsv/tsv.js
@@ -6,6 +6,7 @@ import checkStatusCol from './checkStatusCol'
 import checkTypecol from './checkTypeCol'
 import parseTSV from './tsvParser'
 import checkMotionComponent from './checkMotionComponent'
+import ignore from 'ignore'
 var path = require('path')
 
 /**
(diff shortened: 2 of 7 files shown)


marcelzwiers:
In bidscoin, I have added extra lines of code to remove those bidsignore files from the scans.tsv file because the bids-validator was complaining about them. Since the specs don't mention anything about this (at least not that I'm aware of), I simply took the validator output as the definition of the standard. So does this mean that I should not have added those extra lines of code to bidscoin and should just ignore the validator here?

bids-validator/utils/files/readDir.js
```diff
@@ -330,6 +330,8 @@ async function getBIDSIgnore(dir) {
   if (bidsIgnoreFileObj) {
     const content = await readFile(bidsIgnoreFileObj)
     ig.add(content)
+    // Store the .bidsignore content in session storage
+    sessionStorage.setItem('bidsignoreContent', JSON.stringify(content))
```
effigies, line 337: This is causing an error.

```
Unhandled rejection (
  reason: ReferenceError: sessionStorage is not defined
    at getBIDSIgnore (/home/runner/work/bids-validator/bids-validator/bids-validator/bids-validator/utils/files/readDir.js:334:5)
    at Object.readDir (/home/runner/work/bids-validator/bids-validator/bids-validator/bids-validator/utils/files/readDir.js:23:14)
    at /home/runner/work/bids-validator/bids-validator/bids-validator/bids-validator/validators/bids/start.js:40:21
).
```

effigies:
@marcelzwiers I would call this undefined behavior. The spec didn't say what to do, and bidsignore is not a spec-level concept, so the interaction turned out to be a problem.

As a practical matter, I think following the validator when it exceeds the specification is a good idea unless your goal is to force the issue and aim your users to complain to the validator issue tracker. I wouldn't consider that a bug, but you could stop removing the lines, once this or a similar solution is released.

effigies:
```
 	1: [ERR] Internal error. SOME VALIDATION STEPS MAY NOT HAVE OCCURRED (code: 0 - INTERNAL ERROR)
		Evidence: ReferenceError: sessionStorage is not defined
    at TSV (/home/runner/work/bids-validator/bids-validator/bids-validator/validators/tsv/tsv.js:617:23)
    at /home/runner/work/bids-validator/bids-validator/bids-validator/validators/tsv/validate.js:33:9
```

I'm not sure if global storage is the right tool for this job. @rwblair @nellh Your insight would be appreciated here. I don't think I'm qualified to review this PR.

appukuttan-shailesh:
I did have my doubts about using sessionStorage for solving this problem, and it was infact a fallback approach to get the web app working. I believe the JS web app and the Python package might be using partly the same codebase (?), and if so this will not suffice. My first approach was to re-open the `.bidsignore` file when analyzing the _scans.tsv file, but this was problematic as `.bidsignore` is no longer accessible at that stage of the code (i.e. inside `tsv.js`), and couldn't force read owing to browser security related issues.

Looking forward to some feedback (@rwblair @nellh ) and suggestions on what would be a useful way to tackle this. Happy to try out a different approach.

effigies:
Please ignore the Python package. It is only useful for packaging regular expressions for validating filenames. It will be replaced entirely in the future.

appukuttan-shailesh:
Just trying to debug the remaining errors with the current approach while waiting for further feedback.

Locally, I am testing the following (based on instructions [here](https://github.com/bids-standard/bids-validator?tab=readme-ov-file#testing)) before pushing the changes:
- `npm run test`
- `npm run lint`

But the errors that have been reported above (through the workflows) are never caught/tested via the above. Are there additional instructions for running the entire testing pipeline locally?

I have manually ran the following to ensure previous workflow errors do not arise again:
- `bids-validator/bin/bids-validator bids-validator/tests/data/valid_headers/ --ignoreNiftiHeaders`
- `../../../bin/bids-validator 7t_trt -c /home/shailesh/gits/bids-validator/bids-validator/tests/data/bids-examples/bidsconfig.json --ignoreNiftiHeaders`

5 workflows await approval from maintainers to check if all is clear this time around.

bids-validator/validators/tsv/tsv.js
```diff
@@ -612,13 +616,22 @@ const TSV = (file, contents, fileList, callback) => {
...
         const scanRelativePath = row[filenameColumn]
         const scanFullPath = scanDirPath + '/' + scanRelativePath
 
+        // check if file should be ignored based on .bidsignore content
+        if (ig && ig.ignores(path.relative('/', scanRelativePath))) {
+          continue
+        }
+
         // check if scan matches full dataset path list
         if (!pathList.includes(scanFullPath)) {
```
effigies, line 636: Okay, I get what you're doing here. Thanks for this!

The issue here is that we want to error if a listed file doesn't exist, but we're currently erroring if the file doesn't exist or exists but is ignored. If you're going to follow the strategy you're using here, instead of storing the ignore patterns, you should be storing the ignored files. We can then say

```javascript
 if (!(pathList.includes(scanFullPath) || ignoreList.includes(scanFullPath))) {

Comment (appukuttan-shailesh) [reply bids-validator/validators/tsv/tsv.js:636]: Thanks for the feedback @effigies 

That should also work, just that we would need to store much more data in the session storage (storing ignore patterns vs complete list of ignored files across all subjects and sessions). I'm also curious why the current approach of storing patterns would not suffice; maybe I didn't follow your comment entirely.

As per #1366 , the intention was to allow `.bidsignore` files to be listed in `_scans.tsv`.
The plan then was to: 
> doing a match of scans.tsv files against ignore entries

The PR, in its current form, was intending to just add that one extra check, i.e. for each file listed in `_scans.tsv`, check if the file is to be ignored (based on `.bidsignore`), if yes then ignore that file entry with no error. This is being done irrespective of whether the file actually exists in the directory or not.

I added two tests for checking the following cases:
- file is listed in _scans.tsv but isn't part of output of `readDir()`  -> ERROR 129
- file is listed in _scans.tsv; isn't part of output of `readDir()`, but is to be ignored -> no error

and both these tests are passing.

_Note that there is a difference between the file actually existing (on disk) vs being present in the output of `readDir()` as the latter function already filters out the files to be ignored, before returning the list of required files (i.e. existing and non-ignored) to other steps in the workflow (e.g. TSV() in tsv.js)_

**Comment by appukuttan-shailesh** [comment]
In addition to the discussions above about the nature of the changes, I think the PR is passing most of the checks. 3 of them that it is failing, is found to have failed with similar status in a recent [merged PR](https://github.com/bids-standard/bids-validator/pull/1915) as well, so not sure if it's something specific to this PR that I should look into. I have made a small change to address the 4th failed test which dealt with code coverage; the current change should marginally improve that.

**Comment by appukuttan-shailesh** [comment]
Could we run the remaining 5 workflows to see if they pass?
> 5 workflows awaiting approval

Also, happy to receive any feedback regarding the changes here.

**Comment by appukuttan-shailesh** [comment]
As mentioned above, I believe it is now passing all the checks that a [previous merged PR](https://github.com/bids-standard/bids-validator/pull/1915) conformed to. So not sure if the 3 that failed, and 1 skipped, are something that I should look into.

**Comment by appukuttan-shailesh** [comment]
I was wondering if we had any feedback on this PR.

**Comment by effigies** [comment]
I still think we want the following cases:

| File exists | File is valid | File is ignored | Result |
|--|--|--|--|
| T | T | * | Pass |
| T | F | T | Pass |
| T | F | F | Fail |
| F | * | * | Fail |

It seems like a regression to me to stop erroring when a user mentions a file that does not exist. The goal is to

I spent a little time trying to figure out how to pass ignored files around. I got this far:

```diff
❯ git --no-pager diff
diff --git a/bids-validator/validators/bids/fullTest.js b/bids-validator/validators/bids/fullTest.js
index edf0b405..efcd6204 100644
--- a/bids-validator/validators/bids/fullTest.js
+++ b/bids-validator/validators/bids/fullTest.js
@@ -76,8 +76,10 @@ const fullTest = (fileList, options, annexed, dir, schema, callback) => {
   }
 
   // remove ignored files from list:
+  const ignoredFiles = {}
   Object.keys(fileList).forEach(function (key) {
     if (fileList[key].ignore) {
+      ignoredFiles[key] = fileList[key]
       delete fileList[key]
     }
   })
@@ -123,6 +125,7 @@ const fullTest = (fileList, options, annexed, dir, schema, callback) => {
         participants,
         phenotypeParticipants,
         stimuli,
+        ignoredFiles,
       )
     })
     .then(({ tsvIssues, participantsTsvContent }) => {
diff --git a/bids-validator/validators/tsv/tsv.js b/bids-validator/validators/tsv/tsv.js
index 0b76e6cd..153b62b7 100644
--- a/bids-validator/validators/tsv/tsv.js
+++ b/bids-validator/validators/tsv/tsv.js
@@ -36,7 +36,7 @@ const filenameEvidence = (filename) => `Filename: ${filename}`
  * specification.
  */
 
-const TSV = (file, contents, fileList, callback) => {
+const TSV = (file, contents, fileList, ignoredFiles, callback) => {
   const issues = []
   const stimPaths = []
   if (contents.includes('\r') && !contents.includes('\n')) {
@@ -628,7 +628,7 @@ const TSV = (file, contents, fileList, callback) => {
         const scanFullPath = scanDirPath + '/' + scanRelativePath
 
         // check if file should be ignored based on .bidsignore content
-        if (ig && ig.ignores(path.relative('/', scanRelativePath))) {
+        if (ignoredFiles.includes(scanFullPath)) {
           continue
         }
 
diff --git a/bids-validator/validators/tsv/validate.js b/bids-validator/validators/tsv/validate.js
index dba18e7a..dfcd725c 100644
--- a/bids-validator/validators/tsv/validate.js
+++ b/bids-validator/validators/tsv/validate.js
@@ -9,6 +9,7 @@ const validate = (
   participants,
   phenotypeParticipants,
   stimuli,
+  ignoredFiles,
   annexed,
   dir,
 ) => {
@@ -34,6 +35,7 @@ const validate = (
           file,
           contents,
           fileList,
+          ignoredFiles,
           function (tsvIssues, participantList, stimFiles) {
             if (participantList) {
               if (file.name.endsWith('participants.tsv')) {
```

Something's not working right, but I don't understand the validator enough to know what. Perhaps this will help you, though?

PR #1915, opened 2024-03-12
Add vite web validator for the schema validator

nellh:
Adds a vite project to run the schema validator. Decided to put this at the top level under web, it really doesn't depend on anything except a build path for the schema validator.

The browser bundle target is not watched for changes but run once when you start dev or build modes.

```
# Live reload for development
cd web && deno task dev
# Build for static deployments
cd web && deno task build
```


effigies:
Looks great!

image

Missing lists of affected files, though.

nellh:
> Missing lists of affected files, though.

If the issue generated includes them they are displayed as a list of paths but this seems to be missing from a number of the issues generated.

PR #1924, opened 2024-04-02
FIX: Add minimum Python requirement to package metadata

effigies:
Python 3.8 is the oldest Python that is not EOL, so we only test that far back.
People using older Pythons could still get this installed, and it turns out we
use a Python 3.8 feature.

After the next release, we should yank PyPI packages newer than 1.14.0 so that
Python 3.6 and 3.7 users will install working versions by default.

### DIFF
--- a/bids-validator/pyproject.toml
+++ b/bids-validator/pyproject.toml
@@ -20,6 +20,7 @@ classifiers = [
     "Programming Language :: Python",
     "Topic :: Scientific/Engineering",
 ]
+requires-python = ">=3.8"
 
 [project.urls]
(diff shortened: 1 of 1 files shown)


effigies:
xref https://github.com/bids-standard/pybids/issues/1048

effigies:
@rwblair Could we get a patch release whenever it's convenient for you?

rwblair:
https://github.com/bids-standard/bids-validator/releases/tag/v1.14.5

effigies:
Versions 1.14.1-1.14.4 have been yanked from PyPI, so there should be no problems with installation through normal methods now.

PR #1939, opened 2024-04-19
Have filenameIdentify use datasetContext to determine which filename rules to use.

rwblair:
Currently is applying derived filename rules when it shouldn't be. This makes it difficult to notify a user that has a derivatives dataset mistakenly listed in dataset_description as a derivative. Something that may have never happened before. I'm fine overlooking that for now.

Should we have deriv datasets not apply any schema.rules.files.raw rules? I think thats why we went through all the trouble of making $refs work for filenames in the schema.

### DIFF
--- a/bids-validator/src/schema/applyRules.ts
+++ b/bids-validator/src/schema/applyRules.ts
@@ -31,7 +31,7 @@ export function applyRules(
     rootSchema = schema
   }
   if (!schemaPath) {
-    schemaPath = 'schema'
+    schemaPath = 'schema.rules'
   }
   Object.assign(context, expressionFunctions)
   // @ts-expect-error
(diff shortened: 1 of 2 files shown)


effigies:
> Should we have deriv datasets not apply any schema.rules.files.raw rules?

Raw files are valid in derivative datasets, so there shouldn't be a need to prune them. They may be redundant with the more lax derivatives rules, however, assuming we're thorough enough in keeping derivative counterparts to all raw rules.

effigies (approved):
This LGTM. Don't know if your question and my answer impact whether you want to do anything else. Go ahead and merge, if not.

rwblair:
Given how marginal the performance gains from pruning the deriv branch were I'll leave it how it is right now.

PR #1940, opened 2024-04-19
Pass on multiple rules.files matches from filenameIdentify to filenameValidate.

rwblair:
filenameValidate.ts handles multiple rule matches properly, no need to truncate rule matches unnecessarily in filenameIdentify.ts

fixes longstanding ds000248 validation error.

### DIFF
--- a/bids-validator/src/validators/filenameIdentify.ts
+++ b/bids-validator/src/validators/filenameIdentify.ts
@@ -129,12 +129,9 @@ export function hasMatch(schema, context) {
     const entExtMatch = context.filenameRules.filter((rulePath) => {
       return entitiesExtensionsInRule(schema, context, rulePath)
     })
-    if (entExtMatch.length > 0) {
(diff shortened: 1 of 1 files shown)


bids-validator/src/validators/filenameIdentify.ts
```diff
@@ -129,13 +129,10 @@ export function hasMatch(schema, context) {
     const entExtMatch = context.filenameRules.filter((rulePath) => {
       return entitiesExtensionsInRule(schema, context, rulePath)
     })
-    if (entExtMatch.length > 0) {
-      context.filenameRules = [entExtMatch[0]]
+    if (entExtMatch.length == 1) {
```
effigies: Don't you still want this to be `> 0`, as at least you're paring down the rules?

rwblair: Yes, good catch.

effigies (approved):

PR #1941, opened 2024-04-19
have exists return false on bids-uri check if string doesn't start with 'bids:'

rwblair:
instead of always returning true lets at least check for a sensible 'bid:' at the start of string. allows exists(x, 'bids-uri) + exists(x, 'dataset') == 1 style tests to pass for non bids-uris


effigies:
This bumps up the number of failing examples. Should they be failing where they haven't been?

effigies:
Okay, I'm finding legit errors in ds000117's `coordsystem.json` files.

effigies:
Opened https://github.com/bids-standard/bids-examples/issues/437.

This works for me.

Issue #1947, opened 2024-04-20
IntendedFor links pointing into `derivatives/` directory fail


effigies:
The pruning of derivatives from `context.fileTree` mean we can't verify the existence of links pointing into the `derivatives/` subdirectory. This is a source of failures for `ieeg_motorMiller2007`.

effigies:
Similar issue for when a "file" referenced is a directory like [sub-emptyroom/ses-18901014/meg/sub-emptyroom_ses-18901014_task-noise_run-01_meg.ds](https://github.com/bids-standard/bids-examples/tree/master/ds000247/sub-emptyroom/ses-18901014/meg/sub-emptyroom_ses-18901014_task-noise_run-01_meg.ds).

I think we want `fileTree.contains()` these files to pass while iterating over files fails. Also, in this case a directory needs to return `true`, not simply recurse.

effigies:
This issue covers examples:

* ds000247 (.ds file)
* ieeg_motorMiller2007 (derivatives/)
* ieeg_visual (derivatives/)

PR #1950, opened 2024-04-22
Fix exists for derivative paths in raw datasets.

rwblair:
Keep derivatives in filetree for intendedfor check. Make exists checkaware of leading slashes.
#1947

### DIFF
--- a/bids-validator/src/schema/expressionLanguage.ts
+++ b/bids-validator/src/schema/expressionLanguage.ts
@@ -27,6 +27,17 @@ function exists(list: string[], rule: string = 'dataset'): number {
   } else {
     // dataset, subject and stimuli
     return list.filter((x) => {
+      /* dataset relative paths must start with slash. This leading slash will 
+       * trip up filetree.contains though so chomp if its present and fail if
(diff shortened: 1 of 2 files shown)


bids-validator/src/schema/expressionLanguage.ts
```diff
@@ -27,6 +27,15 @@ function exists(list: string[], rule: string = 'dataset'): number {
...
     return list.filter((x) => {
+      /* local paths must start with slash. This leading slash will 
+       * trip up filetree.contains though so chomp if its present and fail if
+       * it isn't present.
+       */
+      if (x[0] == '/') {
+        x = x.substr(1)
+      } else {
+        return false
+      }
```
effigies: This will break all the other datasets. Relative paths without initial slashes are extremely common. If we want to make this change, I would suggest getting input from a larger group first.

```suggestion
      // Leading slashes will trip up filetree.contains so chomp if its present
      if (x[0] == '/') {
        x = x.substr(1)
      }
```

effigies (approved):
Does the job.

PR #1966, opened 2024-05-04
fix(legacy): Add missing top-level rule for UNIT1 sidecars

bpinsard:
To be reviewed by qMRI BIDS experts.

### DIFF
--- a/bids-validator/bids_validator/rules/top_level_rules.json
+++ b/bids-validator/bids_validator/rules/top_level_rules.json
@@ -140,7 +140,7 @@
   },
 
   "unit1_top": {
-    "regexp": "^[\\/\\\\](?:ses-[a-zA-Z0-9]+_)?(?:acq-[a-zA-Z0-9]+_)?(?:ce-[a-zA-Z0-9]+_)?(?:rec-[a-zA-Z0-9]+_)?(?:run-[0-9]+_)?(?:chunk-[0-9]+_)?(@@@_unit1_suffixes_@@@)\\.json$",
+    "regexp": "^[\\/\\\\](?:ses-[a-zA-Z0-9]+_)?(?:task-[a-zA-Z0-9]+_)?(?:acq-[a-zA-Z0-9]+_)?(?:ce-[a-zA-Z0-9]+_)?(?:rec-[a-zA-Z0-9]+_)?(?:run-[0-9]+_)?(?:chunk-[0-9]+_)?(@@@_unit1_suffixes_@@@)\\.json$",
     "tokens": {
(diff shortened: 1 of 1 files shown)


mathieuboudreau:
@agahkarakuzu

agahkarakuzu:
@bpinsard is the purpose to allow adding UNIT1 under the anat folder? If so, green light on my end, as this was already intended, but probably not implemented.

bpinsard:
@agahkarakuzu not in the anat folder, but at the top level (eg. if one wants to factor shared metadata across the whole dataset). 
I was not certain about simply adding it to the suffixes in the pattern, or if a new pattern needs to be crafted if some entities in this pattern do not conform to the spec for UNIT1 files. 
After carefully reading the spec with a rested brain, it seems that `[_[echo]-<[index]>][_[part-<mag|phase|real|imag>]]` do not apply to UNIT1, so it requires crafting a separate pattern. Will amend this PR accordingly.

bids-validator/bids_validator/rules/top_level_rules.json
```diff
@@ -25,6 +25,7 @@
         "TB1RFM\\.json",
         "TB1SRGE\\.json",
         "RB1COR\\.json",
+        "UNIT1\\.json",
```
effigies: This will be covered by your rule, below.

```suggestion
```

bids-validator/bids_validator/rules/top_level_rules.json
```diff
@@ -139,6 +139,13 @@
     }
   },
 
+  "unit1_top": {
+    "regexp": "^[\\/\\\\](?:ses-[a-zA-Z0-9]+_)?(?:acq-[a-zA-Z0-9]+_)?(?:ce-[a-zA-Z0-9]+_)?(?:rec-[a-zA-Z0-9]+_)?(?:run-[0-9]+_)?(?:chunk-[0-9]+_)?(@@@_unit1_suffixes_@@@)\\.json$",
```
effigies: Extremely unlikely, but `task` is permitted for UNIT1s since https://github.com/bids-standard/bids-specification/pull/1185:

```suggestion
    "regexp": "^[\\/\\\\](?:ses-[a-zA-Z0-9]+_)?(?:task-[a-zA-Z0-9]+_)?(?:acq-[a-zA-Z0-9]+_)?(?:ce-[a-zA-Z0-9]+_)?(?:rec-[a-zA-Z0-9]+_)?(?:run-[0-9]+_)?(?:chunk-[0-9]+_)?(@@@_unit1_suffixes_@@@)\\.json$",
```

effigies (approved):

PR #1973, opened 2024-05-22
Add support for the XPCT suffix in microscopy

chourroutm:
This PR will ensure the new XPCT suffix for microscopy (see https://github.com/bids-standard/bids-specification/pull/1822 and https://github.com/bids-standard/bids-specification/issues/1646) passes the BIDS validator check.


effigies:
Thanks! I just noticed that we didn't cover `.json` files in the file-level rules. Could you also update:

https://github.com/bids-standard/bids-validator/blob/6a32b718015eff54c2847eb4915d52e6a05e8c97/bids-validator/bids_validator/rules/file_level_rules.json#L622-L643

effigies (approved):

PR #1977, opened 2024-05-24
feat(schema): Load subject data

### LINKED ISSUE #1978: Implement allequal in expression language.
bids-standard/bids-specification/pull/1837

effigies:
This fixes the implementation of:

* `rules.checks.dataset.SubjectFolders`
* `rules.checks.dataset.ParticipantIDMismatch`
* `rules.checks.dataset.PhenotypeSubjectsMissing`

These started failing with https://github.com/bids-standard/bids-specification/pull/1833, which activated previously disabled checks.

@rwblair

Closes #1978.

### DIFF
--- a/bids-validator/src/schema/context.test.ts
+++ b/bids-validator/src/schema/context.test.ts
@@ -88,7 +88,7 @@ rootFileTree.files = [
 ]
 rootFileTree.directories = [subjectFileTree]
 
-const context = new BIDSContext(anatFileTree, dataFile, issues)
+let context = new BIDSContext(anatFileTree, dataFile, issues)
 
 Deno.test('test context LoadSidecar', async (t) => {
   await context.loadSidecar(rootFileTree)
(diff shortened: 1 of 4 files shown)


effigies:
The implementation here is really suboptimal. I don't know if we have a way of creating a baseline `context.dataset` object that just gets passed around, but if so, this should go there. If not, we should make it so we're not potentially parsing many TSV files every time we look at another file.

effigies:
Apparently Javascript doesn't allow you to do array equality with `==` or anything else, so checks like `sorted(columns.participant_id) == sorted(dataset.subjects.sub_dirs)` will always fail. I think the only way to handle this will be to create an `isEqual(x,y)` function in the expression language.

rwblair:
> The implementation here is really suboptimal. I don't know if we have a way of creating a baseline `context.dataset` object that just gets passed around, but if so, this should go there. If not, we should make it so we're not potentially parsing many TSV files every time we look at another file.

I believe that a single reference to the dsContext is used in normal validation. Its instantiated here and then passed to the context generator:
https://github.com/bids-standard/bids-validator/blob/master/bids-validator/src/validators/bids.ts#L48

'll look at if it has the information needed to populate the subjects at creation time. If not we can add a check to the file context to only run load subjects if its undefined. I also need to remember why and when the default dsContext is used.

effigies:
Added a commit to leave `.subjects` undefined until the first call, and then only create and populate it if undefined. That way we only duplicate effort if the context is regenerated.

rwblair:
Looking good. 

A heads up the subjects listed in summary aren't dependent on the dsContext. Every context's filename gets checked for  a sub entity and that is added as a subject to the summary:
https://github.com/bids-standard/bids-validator/blob/master/bids-validator/src/summary/summary.ts#L106 

I think its trying to capture a case in which a malformed dataset has more subjects files in it than it has proper sub-<label> directories. Another unlikely case is that subjects with completely empty subject dirs won't be counted in the summary.

I can see this potentially as a point of confusion, but also don't think it will matter 99.9% of the time. We can change it in a future PR if need be.

effigies:
I think if there are no files in a subject dir, it won't make it into the fileTree, but I'm okay with a deviation between these if it does.

PR #1987, opened 2024-06-07
(deno) TSV type check overhaul.

rwblair:
Modify tsv column value type checks to use user defined types if specifiedified in sidecar and schema rules for that column are in the 'json schema' defined style outlined in bids-standard/bids-specification#1838

For traditional schema defined columns, if they're required the type check produces an error, but if it is not required it generates a warning.

I ran `deno fmt` on the files and its got opinions on quotes and semicolons, thats where most the changes come from.

### DIFF
--- a/bids-validator/src/schema/applyRules.ts
+++ b/bids-validator/src/schema/applyRules.ts
@@ -3,12 +3,12 @@ import {
   GenericSchema,
   SchemaFields,
   SchemaTypeLike,
-} from '../types/schema.ts'
-import { Severity } from '../types/issues.ts'
-import { BIDSContext } from './context.ts'
-import { expressionFunctions } from './expressionLanguage.ts'
-import { logger } from '../utils/logger.ts'
-import { memoize } from '../utils/memoize.ts'
+} from "../types/schema.ts";
+import { Severity } from "../types/issues.ts";
+import { BIDSContext } from "./context.ts";
+import { expressionFunctions } from "./expressionLanguage.ts";
+import { logger } from "../utils/logger.ts";
+import { memoize } from "../utils/memoize.ts";
 
 /**
  * Given a schema and context, evaluate which rules match and test them.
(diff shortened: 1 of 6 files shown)


bids-validator/src/validators/bids.ts
```diff
@@ -33,77 +33,83 @@ export async function validate(
...
+    const description = await ddFile.text().then((text) => JSON.parse(text));
+    summary.dataProcessed = description.DatasetType === "derivative";
+    dsContext = new BIDSContextDataset(options, description);
   } else {
+<<<<<<< HEAD
+    dsContext = new BIDSContextDataset(options);
+=======
     dsContext = new BIDSContextDataset(options)
     issues.addNonSchemaIssue('MISSING_DATASET_DESCRIPTION', [] as IssueFile[])
+>>>>>>> 5b7a3648d8cdb825660b1f3e5d9d847862e30402
```
effigies:
```suggestion
    dsContext = new BIDSContextDataset(options);
    issues.addNonSchemaIssue('MISSING_DATASET_DESCRIPTION', [] as IssueFile[]);
```

effigies:
Fixed the CI failures. Back to green on Ubuntu!

bids-validator/src/schema/applyRules.ts
```diff
@@ -168,61 +168,123 @@ function schemaObjectTypeCheck(
...
+): boolean {
+  if (
+    "Levels" in rule && rule["Levels"] && typeof (rule["Levels"]) == "object"
+  ) {
+    return value in rule["Levels"];
+  }
+  if ("Units" in rule) {
+    schemaObjectTypeCheck({ "type": "number" }, value, schema);
+  }
+  return true;
```
effigies: I think this is missing an `n/a` exception for `Levels`, and you aren't returning the value of the schemaObjectTypeCheck. I think if we use an `if/else if/else` block, typescript would be able to detect bad fallthrough behavior.

```suggestion
  if (
    "Levels" in rule && rule["Levels"] && typeof (rule["Levels"]) == "object"
  ) {
    return value == 'n/a' || value in rule["Levels"];
  } else if ("Units" in rule) {
    return schemaObjectTypeCheck({ "type": "number" }, value, schema);
  } else {
    return true;
  }
```

effigies (commented):
I think I read through all the real changes.

effigies (approved):

PR #1988, opened 2024-06-08
fix(tests): Remove assumptions about tmpfile location, clean up

effigies:
OSX tests are failing because we assume temporary files exist in /tmp, while OSX has different defaults. This creates a temporary file and splits the returned path to ensure we can point to existing files.


effigies:
Closed in favor of #1989.

PR #1989, opened 2024-06-08
fix: OSX and Windows tests

effigies:
Builds on #1988, which resolves OSX failures.


effigies:
~~This resolves the tests in the most straightforward way, but it's possible that instead of accommodating backslashes in more places, we want to more aggressively convert some components to always use forward slashes. Otherwise we may find ourselves with datasets that pass on Windows because selectors like `path == '/dataset_description.json'` fail and rules that should fail get skipped.~~

I switched to using POSIX joins to ensure that the internal representation of paths is POSIX. Added a test.

nellh (approved):

PR #1990, opened 2024-06-09
fix(expr): Use a sensible sorting function for sorted()

effigies:
Array.sort(), by default, converts numbers to unicode strings and sorts them lexically. This uses `(a > b) - (a < b)` to implement a standard comparison operator. Greater results in 1, equal in 0, lesser in -1.

### DIFF
--- a/bids-validator/src/schema/expressionLanguage.ts
+++ b/bids-validator/src/schema/expressionLanguage.ts
@@ -88,8 +88,8 @@ export const expressionFunctions = {
     return arg.substr(start, end - start)
   },
   sorted: <T>(list: T[]): T[] => {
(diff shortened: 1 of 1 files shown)


effigies:
I added `sorted([9, 81, 729, 6561])` to the expression tests upstream. This PR fixes it:

```
  sorted([3, 2, 1])                      1,2,3         1,2,3         pass  
  allequal(sorted([3, 2, 1]), [1, 2, 3]) true          true          pass  
  sorted([9, 81, 729, 6561])             9,81,729,6561 9,81,729,6561 pass  
```

nellh (approved):

PR #2014, opened 2024-07-05
Completely rewrite HED support

happy5214:
This commit completely rewrites the HED integration module to only build most objects when necessary (due to memory issues).

This depends on code in hed-validator that has not yet been released to npm, and thus it currently does not build. Please test using the `modify-bids-interface` branch of hed-validator and `npm link`.

### DIFF
--- a/bids-validator/package.json
+++ b/bids-validator/package.json
@@ -45,7 +45,7 @@
     "date-fns": "^3.6.0",
     "events": "^3.3.0",
     "exifreader": "^4.23.2",
-    "hed-validator": "^3.14.0",
+    "hed-validator": "^3.15.0",
     "ignore": "^5.3.1",
     "is-utf8": "^0.2.1",
     "jest": "^29.7.0",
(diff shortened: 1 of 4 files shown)


VisLab (commented):
This looks great --- and follows as discussed in the code walk through. 

@rwblair -- this update should also allow the HED validator to be used more easily in the deno version.  (The current deno version may not actually validate HED.)

This version will require package lock to hed-validator v3.15.0 (which @happy5214 is in the processing of releasing on npm).

effigies:
The `ieeg_epilepsy_ecog` example is failing, presumably because it has `"HEDVersion": "n/a"` in its `dataset_description.json`. Will open a PR to remove that, unless you'd rather it stay?

VisLab:
Yes, it should be removed.  If a `dataset_description.json` has a `HedVersion` field, it must be a valid HED version or it is considered to be an error.  Thx!

PR #2015, opened 2024-07-10
fix(legacy): Prevent prototype pollution (global property) from crashing sidecar merges

nellh:
This avoids the loop iterating over non-enumerable properties of jsonObject.


effigies:
ChatGPT's suggestion:

```js
describe('Object pollution test', () => {
  beforeAll(() => {
    // Simulate the polluting library
    Object.defineProperty(Object.prototype, 'global', {
      get: function() {
        return globalThis;
      },
      configurable: true
    });
  });

  afterAll(() => {
    // Clean up the pollution
    delete Object.prototype.global;
  });

  test('should handle maps with global key correctly', () => {
    const myMap = new Map();
    const globalKey = 'global';

    expect(() => {
      myMap.set(globalKey, 'someValue');
    }).not.toThrow();

    expect(myMap.get(globalKey)).toBe('someValue');
  });

  // Additional tests to verify your patch
});
```

PR #2020, opened 2024-07-12
Generate JSON schema's from spec schema to validate values in sidecars.

rwblair:


rwblair:
This isn't using AJV to validate column value types yet. Currently all values read in from a tsv are treated as strings. The schema specifies the type of some columns as numbers. For now the code looks up the format associated with the type and runs that regex on the values. I'll need to attempt coercion of values if we want type checking there to be similar to json files.

bids-validator/src/schema/context.ts
```diff
@@ -25,13 +30,22 @@ export class BIDSContextDataset implements ContextDataset {
...
-  constructor(options?: ValidatorOptions, description = {}) {
+  constructor(options?: ValidatorOptions, schema?: Schema, description = {}) {
     this.dataset_description = description
     this.files = []
     this.tree = {}
     this.ignored = []
     this.modalities = []
+    this.ajv = new Ajv({strictSchema: false})
+    // @ts-expect-error
+    this.ajv.compile = memoize(this.ajv.compile)
```
rwblair, line 42: Haven't been able to figure out how to make these types mesh. Could store it as another entry in the dataset context/.

bids-validator/src/schema/applyRules.ts
```diff
@@ -442,6 +444,43 @@ function evalJsonCheck(
         ]);
       }
     }
+    let originFileKey = ''
+    if (keyName in context.sidecarKeyOrigin) {
+      originFileKey = `${context.sidecarKeyOrigin[keyName]}.${keyName}`
```
effigies: Since `.` is a part of the path, might be a little clearner to use a different separator:

```suggestion
      originFileKey = `${context.sidecarKeyOrigin[keyName]}:${keyName}`
```

bids-validator/src/schema/applyRules.ts
```diff
@@ -442,6 +444,43 @@ function evalJsonCheck(
...
+    }
+    if (context.dataset.sidecarKeyValidated.has(originFileKey)) {
+      return
+    }
+
+    if (keyName in context.sidecar && schema.objects.metadata && keyName in schema.objects.metadata) {
+      // @ts-expect-error
+      const validate = context.dataset.ajv.compile(schema.objects.metadata[keyName])
```
effigies: Instead of memoizing on the contents of the `objects.metadata.*`, wouldn't it make sense to have something like:

```js
const cache = {}
function getValidator(key, schema) {
    if (!(key in cache)) {
        cache[key] = ajv.compile(schema.objects.metadata[key])
    }
    return cache[key]
}
```

Then:

```suggestion
      const validate = getValidator(keyName, schema)
```

effigies:
> I'll need to attempt coercion of values if we want type checking there to be similar to json files.

I think the regex checking is probably going to be faster than coercion and validation, and it gets us where we want to be.

bids-validator/src/issues/list.ts
```diff
@@ -47,6 +47,10 @@ export const filenameIssues: IssueDefinitionRecord = {
     severity: "warning",
     reason: "A data files JSON sidecar is missing a key listed as recommended.",
   },
+  JSON_SCHEMA_VALIDATION_ERROR: {
```
rwblair, line 45: We should pull this code from schema.rules.errors (and make all other errors there available. Internally represent with code as the key and rest of object as contents. add `addSchemaIssue` to src/issues/schemaIssues.ts or something.

bids-validator/src/deps/ajv.ts
```diff
@@ -0,0 +1,4 @@
+export {Ajv} from 'https://esm.sh/ajv@8.16.0';
+import ValidateFunction from 'https://esm.sh/ajv@8.16.0';
+import JSONSchemaType from 'https://esm.sh/ajv@8.16.0';
+export { ValidateFunction, JSONSchemaType }
```
effigies: Maybe?

```suggestion
export {Ajv, ValidateFunction, JSONSchemaType } from 'https://esm.sh/ajv@8.16.0';
```

bids-validator/src/schema/context.ts
nellh: Maybe AJV itself should be a private field and compile is just a method on the context?

bids-validator/src/schema/applyRules.ts
```diff
@@ -442,6 +444,44 @@ function evalJsonCheck(
         ]);
       }
     }
+    let originFileKey = ''
+    if (keyName in context.sidecarKeyOrigin) {
+      originFileKey = `${context.sidecarKeyOrigin[keyName]}:${keyName}`
+    }
```
effigies, line 460: else emit warning or error?

bids-validator/src/schema/applyRules.ts
```diff
@@ -442,6 +444,44 @@ function evalJsonCheck(
...
+    if (context.dataset.sidecarKeyValidated.has(originFileKey)) {
+      return
+    }
+
+    if (keyName in context.sidecar && schema.objects.metadata && keyName in schema.objects.metadata) {
```
effigies: Always true

```suggestion
    if (keyName in context.sidecar) {
```

bids-validator/src/schema/applyRules.ts
```diff
@@ -442,6 +444,44 @@ function evalJsonCheck(
...
+      // @ts-expect-error
+      const validate = context.dataset.ajv.compile(schema.objects.metadata[keyName])
+      const result = validate(context.sidecar[keyName])
+      if (result === false) {
+        const evidenceBase = `Failed for this file.key: ${originFileKey} Schema path: ${schemaPath}`
+        let messages = []
```
nellh: Unused?
```suggestion
```

bids-validator/src/schema/applyRules.ts
```diff
@@ -442,6 +444,44 @@ function evalJsonCheck(
...
+        if (!validate.errors) {
```
effigies: Does this get cleared on re-calling validate? If not, can we clear it explicitly?

bids-validator/src/schema/applyRules.ts
```diff
@@ -442,6 +444,44 @@ function evalJsonCheck(
...
+    if (context.dataset.sidecarKeyValidated.has(originFileKey)) {
+      return
+    }
+
+    if (keyName in context.sidecar && schema.objects.metadata && keyName in schema.objects.metadata) {
+      // @ts-expect-error
+      const validate = context.dataset.ajv.compile(schema.objects.metadata[keyName])
+      const result = validate(context.sidecar[keyName])
+      if (result === false) {
+        const evidenceBase = `Failed for this file.key: ${originFileKey} Schema path: ${schemaPath}`
```
rwblair: Errors could take schema path as a seperate argument and print the entire rule. https://deno.land/x/json2yaml@v1.0.1

bids-validator/src/schema/applyRules.ts
rwblair: Confirmed that errors does get reset on each call to validate. Only the new error message(s) or null if there are no errors on latest validation call.

rwblair:
~~ds000117, ds000246, ds000247 failing because this pr can't handle these two entries:~~
https://bids-specification.readthedocs.io/en/stable/glossary.html#digitizedheadpoints-sense-1-metadata
https://bids-specification.readthedocs.io/en/stable/glossary.html#digitizedheadpoints-sense-2-metadata

~~Should be merged into an anyOf?~~

effigies:
Actually a schema bug. Should be:

```diff
diff --git a/src/schema/rules/sidecars/meg.yaml b/src/schema/rules/sidecars/meg.yaml
index 33d851f6..5391c829 100644
--- a/src/schema/rules/sidecars/meg.yaml
+++ b/src/schema/rules/sidecars/meg.yaml
@@ -205,7 +205,7 @@ MEGCoordsystemDigitizedHeadPoints:
     - datatype == "meg"
     - suffix == "coordsystem"
   fields:
-    DigitizedHeadPoints: optional
+    DigitizedHeadPoints__coordsystem: optional
     DigitizedHeadPointsCoordinateSystem: optional
     DigitizedHeadPointsCoordinateUnits: optional
     DigitizedHeadPointsCoordinateSystemDescription:
```

effigies:
Should be able to test the above patch with:

```
BIDS_SCHEMA=https://bids-specification--1876.org.readthedocs.build/en/1876/schema.json deno test ...
```

effigies:
@rwblair I pushed a small fix in 6b61737431e9d781fb905e205e28a9ddc12fc523, but otherwise I think we're good to go on this.

effigies (approved):
This LGTM. I have a potential refactor that pulls the ajv out of `context.dataset`, but I don't want to hold this up on that. I'll clean it up and make a PR after this is in, if it seems worth the effort.

PR #2027, opened 2024-07-29
feat(json): Warn on missing RECOMMENDED metadata

effigies:
This dramatically increases the number of warnings emitted, since BIDS has been very liberal with its RECOMMENDATIONS.

The alternative is to go add issues to all of the sidecar fields that previously warned.


rwblair:
People seem to ignore warnings as is, but not reporting them is against the spirit of BIDS. I'm neutral either way.

PR #2030, opened 2024-07-30
fix(context): Add file size to context, fixes README_FILE_SMALL

effigies:
We were getting false positives on README_FILE_SMALL:

```
ReadmeFileSmall:
  issue:
    code: README_FILE_SMALL
    message: |
      The recommended file '/README' is very small.
      Please consider expanding it with additional information about the dataset.
    level: warning
  selectors:
    - match(path, '^/README')
  checks:
    - size > 150
```

Turns out size wasn't implemented, but the fix is quick.

### DIFF
--- a/bids-validator/src/schema/context.ts
+++ b/bids-validator/src/schema/context.ts
@@ -107,6 +107,10 @@ export class BIDSContext implements Context {
     this.associations = {} as ContextAssociations
   }
 
+  get size(): number {
+    return this.file.size
+  }
+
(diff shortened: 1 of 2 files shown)


rwblair:
We should double check this in the browser to make sure a `size` is reported on that api's file object.

effigies:
Confirmed that this issue doesn't falsely show up in the web browser now. Truncating a `README` to 149 characters does cause the issue to be raised.

PR #2031, opened 2024-07-30
feat(eval): Distinguish JSON rules from sidecar rules

effigies:
We currently duplicate a lot of sidecar errors for both the data file and the sidecar. This PR skips sidecar checks for sidecar files, avoids loading sidecar data for JSON files, and creates separate issue codes for missing keys in standalone JSON files (`dataset_description.json`, `*_coordsystem.json`, etc.) vs sidecar files.


effigies:
This needs https://github.com/bids-standard/bids-specification/pull/1878. Not filling in `context.sidecar` for JSON files helped catch some checks that were triggered on sidecar files as well as their data files.

PR #2035, opened 2024-07-31
rf(context): Refactor inheritance principle to walk from leaf to root

effigies:
This PR proposes a single implementation of the inheritance principle.

* BIDSFile gains a `.parent` reference to a `FileTree` instance. For the moment it was simplest to treat it as optional, because the browser implementation constructs the `BIDSFile` objects before the containing `FileTree` objects.
* `src/files/inheritance.ts` exposes a `walkBack` generator function that finds at most one matching file in each level. If `inherit = False`, it only looks at the leaf level.
* Loading sidecars and building associations is trivial from here.


bids-validator/src/files/inheritance.ts
```diff
@@ -0,0 +1,47 @@
...
+        )
+      })
+      if (exactMatch) {
+        yield exactMatch
+      } else {
+        console.warn(`
+Multiple candidates detected for '${source.path}'
+
+${candidates.map((file) => `* ${file.path}`).join('\n')}
+`)
```
effigies, line 39: This should really be an issue, as the presence of multiple applicable files at a level violates the spec.

PR #2039, opened 2024-07-31
feat(context): Detect duplicate columns when parsing TSV files

effigies:
This is hacky, but it works. @rwblair LMK how you would integrate this into the existing issue framework.


rwblair:
That throw is great. You can use `addNonSchemaIssue` instead of `add`. You'll have to add `TSV_COLUMN_HEADER_DUPLICATE` to https://github.com/bids-standard/bids-validator/blob/master/bids-validator/src/issues/list.ts

PR #2042, opened 2024-07-31
fix(expr): Escape backslashes in checks before interpolating

effigies:
Was testing https://github.com/bids-standard/bids-specification/pull/1879 and getting odd results. Turns out that our creating a function from strings can collapse escapes. This only appears to affect `match()` expressions right now.

Will work on a test.


effigies:
Re-pushed just the test to show that it fails. Now pushing the fix to show that it resolves the problem.

Added some argumentation into the comments of the test to explain to future us why the fix is here, not in the schema.

rwblair:
Far out, didn't know about `String.raw...`. Looks good to me.

PR #2051, opened 2024-08-02
fix(channels): Add needed fields to context.associations.channels

effigies:
Partners with https://github.com/bids-standard/bids-specification/pull/1880 to provide the fields needed for the NIRS checks.

I got rid of the `|| []` guards. Undefined is the more intuitive representation of an absent column.

### DIFF
--- a/bids-validator/src/schema/associations.ts
+++ b/bids-validator/src/schema/associations.ts
@@ -111,8 +111,9 @@ const associationLookup = {
       const columns = parseTSV(contents)
       return {
         path: file.path,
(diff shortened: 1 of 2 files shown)


rwblair:
Looks good to me. I'll merge once bids-standard/bids-specification/pull/1880 gets signed off on, or if you don't think any changes will be made to spec that affect this you can merge it now.

effigies:
The only changes I expect there would be requirement levels or error messages, so they shouldn't affect this.

PR #2052, opened 2024-08-02
feat(tsv): Memoize TSV loading with a tiered cache, deleting after leaving a directory

### LINKED ISSUE #2044: JSON/TSV caching
Sidecars and associated files may be loaded from disk once for each file they apply to and in themselves. It would be good to cache these to avoid both the IO and parsing costs of re-execution. However, in a dataset with many files and sidecars in the leaf directories instead of the root, a full memoization would use an excessive amount of memory.

Consider a two-level cache:

```python
class Cache[T]:
    func: Callable[Path, T]
    cache: dict[Path, dict[Path, T]]

    def __call__(self, path: Path) -> T:
        subcache = self.cache.setdefault(path.parent, {})
        val = subcache.get(path)
        if val is None:
            subcache[path] = val = self.func(path)
        return val

    def clear(path: Path):
        del self.cache[path]
```

As we walk the file tree, we could clear, for example `/sub-01` when we move on to `/sub-02`, but `/T1w.json` will be in the `/` subcache, so it would not be reloaded.

effigies:
This reduces the runtime of ds004395 from 3m49s to 2m28s, while having negligible memory impact.

In contrast, a naive memoize that never clears hits process limits in ~30s.

Doing the same with JSON does not produce a measurable reduction (actually 2m43s, so it's an increase, but could be within error bars).

The main other things we parse are bvals/bvecs and NIFTI headers. Caching headers could be a good idea, but will need to test with a large, fully downloaded dataset.

Closes #2044.

### DIFF
--- a/bids-validator/src/files/browser.ts
+++ b/bids-validator/src/files/browser.ts
@@ -10,15 +10,16 @@ export class BIDSFileBrowser implements BIDSFile {
   #file: File
   name: string
   path: string
-  parent?: FileTree
+  parent: FileTree
 
-  constructor(file: File, ignore: FileIgnoreRules) {
+  constructor(file: File, ignore: FileIgnoreRules, parent?: FileTree) {
     this.#file = file
     this.#ignore = ignore
     this.name = file.name
     const relativePath = this.#file.webkitRelativePath
     const prefixLength = relativePath.indexOf('/')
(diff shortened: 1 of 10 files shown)


effigies:
This is ready to go.

PR #2056, opened 2024-08-04
feat: I/O checks for NIfTI and JSON files

effigies:
This implements standard issues:

* INVALID_JSON_ENCODING
* JSON_INVALID
* NIFTI_HEADER_UNREADABLE

Because `BIDSFile.text()` is intended to mimic the behavior of `File.text()` and replace unknown characters, it's unsuited to validating JSON, which needs to be actual UTF-8. Further, piping `BIDSFile.stream` through a `TextDecoderStream` can reliably cause resource leaks on UTF16 data (https://github.com/denoland/deno/issues/24872). Therefore I've reimplemented reading with `readBytes()` followed by an all-at-once decoding, which seems to be safer.

~~The main bit of interest is in passing options to `TextDecoderStream` to ensure we catch bad unicode. This creates a divergence with the browser implementation, so we would need to decide what the best API is. I could switch to using `file.stream` directly.~~

~~There appears to be a bug in `TextDecoderStream` (https://github.com/denoland/deno/issues/24872) that prevents us cleaning up properly. So the subtests all pass, but the whole JSON UTF-16 test fails due to a detected leak.~~

### DIFF
--- a/bids-validator/src/files/json.test.ts
+++ b/bids-validator/src/files/json.test.ts
@@ -1,0 +1,64 @@
+import { assert, assertObjectMatch } from '../deps/asserts.ts'
+import { BIDSFileDeno, UnicodeDecodeError } from './deno.ts'
+import { BIDSFile } from '../types/filetree.ts'
+import { FileIgnoreRules } from './ignore.ts'
+
+import { loadJSON } from './json.ts'
+
+function encodeUTF16(text: string) {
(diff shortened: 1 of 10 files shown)


rwblair (approved):
Looks good.

At some point, for all the throws and catches that get turned into issues, we should either put the error message from the initial throw into the issue, or pass it off to the logger.

Issue #2059, opened 2024-08-06
v1.14.7 npm package broken on bids-examples


effigies:
The stable tests on the examples repo are failing:

https://github.com/bids-standard/bids-examples/actions/runs/10259615491/job/28384370437?pr=461

The `master` tests have not failed at all, before or after the release. It seems to be something related to what gets pushed to npm.

The errors are HED-related:

```
Validating dataset asl001/: Running  bids-validator asl001 -c /home/runner/work/bids-examples/bids-examples/bidsconfig.json --ignoreNiftiHeaders
bids-validator@1.14.7
	1: [ERR] An internal error occurred during HED validation. (code: 106 - HED_INTERNAL_ERROR)

        Summary:                Available Tasks:        Available Modalities: 
        8 Files, 94.63KB                                MRI                   
        1 - Subject                                                           
        1 - Session                                                           


	If you have any questions, please post on https://neurostars.org/tags/bids.
```

cc @happy5214

sappelhoff:
We see the exact same issue at https://github.com/mne-tools/mne-bids

cc-ing @VisLab, too

VisLab:
I was going to running the `bids-validator` on my local machine to test the issue with the more comprehensive error messages that @happy5214 just pushed up to the `hed-validator`.  

However, when I looked at the newly installed npm bids-validator node-modules there was no `@hedvalidator` in it. When I tried to link to my local copy of the `hed-validator` this is verified by the following message:

> bids-validator> npm link hed-javascript
> npm error code E404
> npm error 404 Not Found - GET https://registry.npmjs.org/hed-javascript - Not found
> npm error 404
> npm error 404  'hed-javascript@*' is not in this registry.

I can do an npm install hed-validator with no problem so I think this problem might be in the bids-validator release on npm? @rwblair @nellh

effigies:
@VisLab Should that be `npm link hed-validator`, not `hed-javascript`?

rwblair:
Published a new release, v1.14.8 @VisLab @sappelhoff let me know if this new version fixes the issue.

1.14.7 did have a version of the hed validator bundled in its pack, but ~~`hedValidator.bids.buildBidsSchema`~~ `hedValidator.bids.buildBidsSchemas` was undefined. I'm still not sure what went wrong in the build or pack step to cause it.

happy5214:
[...]

PR #2061, opened 2024-08-06
Bump hed-validator to v3.15.4

happy5214:
This should provide the actual exception messages for the internal errors in #2059.

### DIFF
--- a/bids-validator/package.json
+++ b/bids-validator/package.json
@@ -45,7 +45,7 @@
     "date-fns": "^3.6.0",
     "events": "^3.3.0",
     "exifreader": "^4.23.3",
-    "hed-validator": "^3.15.3",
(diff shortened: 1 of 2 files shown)


effigies:
Will we need to re-release in order to observe these issues?

happy5214:
> Will we need to re-release in order to observe these issues?

That depends on how strictly npm follows the `package-lock.json` on downloaded npm packages (as opposed to cloning a git repo). If it only uses `package.json`, v3.15.4 will be pulled automatically by the semver matching.

effigies:
Rerunning on bids-examples shows the same result: https://github.com/bids-standard/bids-examples/actions/runs/10259615491/job/28414339269

I suspect it's following the `package-lock.json` then.

Issue #2043, opened 2024-08-01
Rethinking issues


effigies:
While triaging a failure to report a missing recommended key, I discovered that only one key is shown per file, presumably to avoid excess noise. I'd like to rethink the Issue structure. I am starting to think of it as a table:

| Issue | Sub-issue | Severity | Location | Message | Affects | 
|--|--|--|--|--|--|
| EMPTY_FILE | None | Error | `<path>` | ... | `<path>` |
| SIDECAR_KEY_RECOMMENDED | SidecarKey | Warning | `<path>.json` | ... | `<datafile1> <datafile2> ...` |

Here `(Issue, Sub-issue, Location)` are lookup keys to a specific problem in a file. The severity and location are the main things needed to determine whether and where to fix the problem, with message and affects being useful for understanding.

To mock up a structure in Python:

```python
class Issue:
    code: str
    sub: str | None
    severity: Literal["Error", "Warning"]
    location: Path
    message: str
    affects: list[Path]

class Issues:
    table: dict[tuple[str, str | None, Path], Issue]

    def add(self, code, sub, severity, location, msg, affected_file):
        key = (code, sub, location)
        issue = self.table.get(key) or Issue(code, sub, severity, location, msg, [])
        issue.affects.append(affected_file)
```

Thoughts? Anything you think we should add, or something we currently have that this wouldn't handle?

effigies:
Updated issue structure:

```python
class Issue:
    code: str
    sub: str | None  # Mostly applies to missing/invalid JSON keys or TSV columns
    location: Path | None  # Likely be absent for missing sidecar keys
    severity: Literal["Error", "Warning"]
    message: str | None
    suggestion: str | None
    affects: list[Path]
    rule: str | None  # Schema path, if applicable (maybe)
```

Issue #2007, opened 2024-06-22
Validate dataset_description.json contents


yarikoptic:
I am preparing example for the

- https://github.com/bids-standard/bids-specification/pull/1861

and upon initial validation receiving NO error although it should be there since I entered (for extra check) a completely unknown value.

```shell
❯ bids-validator-deno --version
bids-validator v1.14.7-dev.0-68-g53fdba69

❯ grep DatasetType dataset_description.json
    "DatasetType": "prlkjasdfoject",
```

```shell
❯ bids-validator-deno $PWD
This dataset appears to be BIDS compatible.

          Summary:                         Available Tasks:        Available Modalities:
          1 Files, 336 B                                                                
          0 - Subjects 1 - Sessions                                                     

	If you have any questions, please post on https://neurostars.org/tags/bids.

❯ bids-validator $PWD

bids-validator@1.14.5
	1: [ERR] Invalid JSON file. The file is not formatted according the schema. (code: 55 - JSON_SCHEMA_VALIDATION_ERROR)
		./dataset_description.json
			Evidence: .DatasetType should be equal to one of the allowed values

	Please visit https://neurostars.org/search?q=JSON_SCHEMA_VALIDATION_ERROR for existing conversations about this issue.

	2: [ERR] There are no subject folders (labeled "sub-*") in the root of this dataset. (code: 45 - SUBJECT_FOLDERS)

	Please visit https://neurostars.org/search?q=SUBJECT_FOLDERS for existing conversations about this issue.

	1: [WARN] The recommended file /README is missing. See Section 03 (Modality agnostic files) of the BIDS specification. (code: 101 - README_FILE_MISSING)

	Please visit https://neurostars.org/search?q=README_FILE_MISSING for existing conversations about this issue.

        Summary:            Available Tasks:        Available Modalities: 
        1 File, 336B                                                      
        0 - Subjects                                                      
        1 - Session                                                       


	If you have any questions, please post on https://neurostars.org/tags/bids.
```

effigies:
```console
❯ deno run -A src/bids-validator.ts /tmp/empty      
	[ERROR] A dataset_description.json file is required in the root of the dataset (MISSING_DATASET_DESCRIPTION)


		0 more files with the same issue

	Please visit https://neurostars.org/search?q=MISSING_DATASET_DESCRIPTION for existing conversations about this issue.


          Summary:                         Available Tasks:        Available Modalities:
          0 Files, 0 B                                                                  
[...]

PR #2067, opened 2024-08-06
Refactor DatasetIssues and Issue classes

rwblair:
Towards #2043

This PR flattens the structure of the DatasetIssues and Issue types. It adds a new `subCode` and `rule` field for further differentiation of issues. The issues of DatasetIssues are an array instead of as a map on . The nested files maps were removed in favor of creating a separate issue for each infraction file, code and all. With that there was no need to do any look ups of existing issues during validation.

`addNonSchemaIssue` was removed, its functionality combined with the regular `add`. Both took arrays of files as an argument, but in practice only one file was ever passed to it. 

The old outputs would have these file objects with path, filenam, and relative path. It wasn't clear to me that those were really necessary. I tried to get away with just using the path to the file as a string for simplicity. But I may be missing something and having a proper File object in there would be better, when we have one.

I renamed `reason` to `codeMessage` since the text was always being drawn from where the code is defined be it schema or nonschema.
I renamed `evidence` to `issueMessage` since these were always generated at issue creation time during validation. 
I don't know that either of these is actually clearer.

`filter` and `groupBy` have new DatasetIssues objects as part of their returns,
while get returns just an array of issues. should `get` return a new DatasetIssues?

You can chain calls to `filter`, is that useful? Chaining calls to `groupBy` isn't so easy, should it be easy? I had a version of groupBy that accepted an array of Issue keys to nest that would nest the groups of issues in a object for all the unique values of the keys in the order they were given, it was a mess so I removed it. 

I had played with array of Issue keys being passed into the DatasetIssues constructor and used to set up a bunch of independent maps of values to be indexed on stored in this.groups (`Map<keyof Issue, Map<Issue[keyof Issue]>, Issue[]>`) but then I couldn't imagine it ever being used so I gave up on it.

Should DatasetIssues extend array? Push issues to this instead of this.issues. At its heart it is an array with some helper functions.

Currently issues store `codeMessage`s which leads to quite a bit of duplication, they are consistently the longest field in an issue object and make this issues object 10x larger than old one.  The`codeMessages` map in DatasetIssues could be used to keep a single copy of each message so they could be dropped from the Issue object, but then the issue object needs some assemblly when accessed.

If only we could store a reference to fields in codeMessages in issues.


effigies:
@rwblair Just a note that I've opened https://github.com/rwblair/bids-validator/pull/10 against this. Issues discovered while attempting to address #2007, but the conflict with this PR was too high to be worth addressing separately.

bids-validator/src/types/issues.ts
```diff
@@ -1,6 +1,24 @@
...
+/**
+ * Updated internal Issue structure for schema based validation
+ */
+export interface Issue {
+  code: string
+  subCode?: string
+  severity?: Severity
+  location?: string
+  issueMessage?: string
+  codeMessage?: string
```
effigies:
```suggestion
```

bids-validator/src/issues/datasetIssues.ts
```diff
@@ -11,137 +10,129 @@ import {
...
-    reason,
-    severity = 'error',
-    files = [],
-  }: DatasetIssuesAddParams): Issue {
-    const existingIssue = this.get(key)
-    // Handle both the shorthand BIDSFile array or full IssueFile
-    if (existingIssue) {
-      for (const f of files) {
-        existingIssue.files.set(f.path, f)
+  add(issue: Issue) {
```
effigies:
```suggestion
  add(issue: Issue, codeMessage?: string) {
```

bids-validator/src/schema/applyRules.ts
```diff
@@ -250,12 +251,12 @@ export function evalColumns(
...
-      context.dataset.issues.addNonSchemaIssue('TSV_COLUMN_MISSING', [
-        {
-          ...context.file,
-          evidence: `Column with header ${name} listed as required. ${schemaPath}`,
-        },
-      ])
+      context.dataset.issues.add({
+        code: 'TSV_COLUMN_MISSING',
+        location: context.path,
+        issueMessage: `Column with header ${name} listed as required.`,
```
effigies:
```suggestion
        subCode: name,
        location: context.path,
```

bids-validator/src/schema/applyRules.ts
```diff
@@ -276,10 +277,12 @@ export function evalColumns(
...
         errorObject = context.sidecar[name]
       } else {
-        context.dataset.issues.addNonSchemaIssue('TSV_COLUMN_TYPE_REDEFINED', [{
-          ...context.file,
-          evidence: `'${name}' redefined with sidecar ${inspect(context.sidecar[name])}`,
-        }])
+        context.dataset.issues.add({
+          code: 'TSV_COLUMN_TYPE_REDEFINED',
+          location: context.path,
+          issueMessage: `'${name}' redefined with sidecar ${inspect(context.sidecar[name])}`,
```
effigies:
```suggestion
          code: 'TSV_COLUMN_TYPE_REDEFINED',
          subCode: name,
          location: context.path,
          issueMessage: `Sidecar definition: ${inspect(context.sidecar[name])}`,
```

bids-validator/src/schema/applyRules.ts
```diff
@@ -291,12 +294,12 @@ export function evalColumns(
...
-        context.dataset.issues.addNonSchemaIssue(error_code, [
-          {
-            ...context.file,
-            evidence: `'${value}' ${inspect(columnObject)}`,
-          },
-        ])
+        context.dataset.issues.add({
+          code: error_code,
+          location: context.path,
+          issueMessage: `'${value}' ${inspect(columnObject)}`,
```
effigies:
```suggestion
          code: error_code,
          subCode: name
          location: context.path,
          issueMessage: `'${value}'`,
```

We could probably add row number if we enumerate the loop.

bids-validator/src/schema/applyRules.ts
```diff
@@ -324,17 +327,23 @@ function evalInitialColumns(
...
-        `Column with header ${ruleHeaderName} not found, indexed from 0 it should appear in column ${ruleIndex}. ${schemaPath}`
-      context.dataset.issues.addNonSchemaIssue('TSV_COLUMN_MISSING', [
-        { ...context.file, evidence: evidence },
-      ])
+      const message =
+        `Column with header ${ruleHeaderName} not found, indexed from 0 it should appear in column ${ruleIndex}.`
+      context.dataset.issues.add({
+        code: 'TSV_COLUMN_MISSING',
+        location: context.path,
+        issueMessage: message,
```
effigies:
```suggestion
      context.dataset.issues.add({
        code: 'TSV_COLUMN_MISSING',
        subCode: ruleHeaderName,
        location: context.path,
        issueMessage: `Column ${ruleIndex} (starting from 0) not found.`,
```

bids-validator/src/schema/applyRules.ts
```diff
@@ -324,17 +327,23 @@ function evalInitialColumns(
...
-        `Column with header ${ruleHeaderName} found at index ${contextIndex} while rule specifies, indexed from 0, it should be in column ${ruleIndex}. ${schemaPath}`
-      context.dataset.issues.addNonSchemaIssue('TSV_COLUMN_ORDER_INCORRECT', [
-        { ...context.file, evidence: evidence },
-      ])
+      const message =
+        `Column with header ${ruleHeaderName} found at index ${contextIndex} while rule specifies, indexed from 0, it should be in column ${ruleIndex}.`
+      context.dataset.issues.add({
+        code: 'TSV_COLUMN_ORDER_INCORRECT',
+        location: context.path,
+        issueMessage: message,
```
effigies:
```suggestion
      context.dataset.issues.add({
        code: 'TSV_COLUMN_ORDER_INCORRECT',
        subCode: ruleHeaderName,
        location: context.path,
        issueMessage: `Column ${ruleIndex} (starting from 0) found at index ${contextIndex}.`,
```

bids-validator/src/schema/applyRules.ts
```diff
@@ -360,9 +369,12 @@ function evalAdditionalColumns(
...
     if (extraCols.length) {
-      context.dataset.issues.addNonSchemaIssue('TSV_ADDITIONAL_COLUMNS_NOT_ALLOWED', [
-        { ...context.file, evidence: `Disallowed columns found ${extraCols}` },
-      ])
+      context.dataset.issues.add({
+        code: 'TSV_ADDITIONAL_COLUMNS_NOT_ALLOWED',
+        location: context.path,
+        issueMessage: `Disallowed columns found ${extraCols}`,
+        rule: schemaPath,
+      })
```
effigies: Needs more than I can do in a suggestion, due to diff, but what about:

```ts
for (const col of extraCols) {
  context.dataset.issues.add({
    code: 'TSV_ADDITIONAL_COLUMNS_NOT_ALLOWED',
    subCode: col,
    location: context.path,
    rule: schemaPath,
  })
}
```

bids-validator/src/schema/applyRules.ts
```diff
@@ -389,12 +401,12 @@ function evalIndexColumns(
...
-        ...context.file,
-        evidence: `Columns cited as index columns not in file: ${missing}. ${schemaPath}`,
-      },
-    ])
+    context.dataset.issues.add({
+      code: 'TSV_COLUMN_MISSING',
+      location: context.path,
+      issueMessage: `Columns cited as index columns not in file: ${missing}.`,
+      rule: schemaPath,
+    })
```
effigies: Again, I think a foreach using subcodes would be good here.

bids-validator/src/schema/applyRules.ts
```diff
@@ -441,31 +456,30 @@ function evalJsonCheck(
...
-          key: requirement.issue.code,
-          reason: requirement.issue.message,
+          code: requirement.issue.code,
+          subCode: keyName,
+          location: context.path,
           severity,
-          files: [{ ...context.file }],
+          codeMessage: requirement.issue.message,
+          rule: schemaPath,
         })
```
effigies: Moving `codeMessage` out. Let deno reformat.

```suggestion
          rule: schemaPath,
        }, requirement.issue.message)
```

bids-validator/src/issues/datasetIssues.ts
```diff
@@ -11,137 +10,129 @@ import {
...
+    if (!issue.codeMessage) {
+      if (issue.code in nonSchemaIssues) {
+        issue.codeMessage = nonSchemaIssues[issue.code].reason
+        issue.severity ??= nonSchemaIssues[issue.code].severity
+      } else {
+        throw new Error(
+          `key: ${issue.code} does not exist in non-schema issues definitions`,
+        )
```
effigies:
```suggestion
    if (!codeMessage) {
      if (issue.code in nonSchemaIssues) {
        codeMessage = nonSchemaIssues[issue.code].reason
        issue.severity ??= nonSchemaIssues[issue.code].severity
      } else {
        throw new Error(
          `key: ${issue.code} does not exist in non-schema issues definitions`,
        )
```

bids-validator/src/issues/datasetIssues.ts
```diff
@@ -11,137 +10,129 @@ import {
...
-        severity,
-        reason,
-        files,
-      })
-      this.set(key, newIssue)
-      return newIssue
     }
+    issue.severity ??= 'error'
+    if (!this.codeMessages.has(issue.code)) {
+      this.codeMessages.set(issue.code, issue.codeMessage)
```
effigies:
```suggestion
      this.codeMessages.set(issue.code, codeMessage)
```

effigies (commented):
Stopped making `codeMessage` suggestions. Grabbing food, so submitting now.

effigies (approved):

PR #2071, opened 2024-08-07
Enforce uniqueness of data files in a directory.

### LINKED ISSUE #1107: Validator passed corrupted BIDS EEG
This BIDS dataset contains both .edf and .bdf file (which are very small)

https://openneuro.org/datasets/ds002034/versions/1.0.1

sub-01/ses-01/eeg/sub-01_ses-01_task-offline_run-01_eeg.edf
sub-01/ses-01/eeg/sub-01_ses-01_task-offline_run-01_eeg.bdf

I believe it should not have passed the validator since there are 2 types of binary files and the BDF file is obviously corrupted.

### LINKED ISSUE #2215: Validator MUST NOT accept identical files under different extensions
This BIDS dataset contains both .edf and .bdf file (which are very small)

https://openneuro.org/datasets/ds002034/versions/1.0.1

sub-01/ses-01/eeg/sub-01_ses-01_task-offline_run-01_eeg.edf
sub-01/ses-01/eeg/sub-01_ses-01_task-offline_run-01_eeg.bdf

I believe it should not have passed the validator since there are 2 types of binary files and the BDF file is obviously corrupted.

rwblair:
https://bids-specification.readthedocs.io/en/stable/common-principles.html#uniqueness-of-data-files
fixes bids-standard/bids-specification#2215

Added in src/valdiators/filenameValidate.ts. This check sees if another file exists in the parent directory of the context that has the contexts exact entities and sufiix but a different valid (non-sidecar) extension from the `rules.files.*` rule that the context matched.

Unfortunately there is a problem with relying on the `rules.files.*`. It does not state which extensions are mutually exclusive. One ieeg format for instance states `Each recording consists of a .vhdr, .vmrk, .eeg file triplet` as valid:
https://bids-specification.readthedocs.io/en/stable/modality-specific-files/intracranial-electroencephalography.html#ieeg-recording-data
https://github.com/bids-standard/bids-specification/blob/master/src/schema/rules/files/raw/ieeg.yaml

Maintaing list of sidecar extensions in filenameValidate.ts is no good and should probably be pulled from the schema if possible.

### DIFF
--- a/bids-validator/src/issues/list.ts
+++ b/bids-validator/src/issues/list.ts
@@ -42,6 +42,10 @@ export const bidsIssues: IssueDefinitionRecord = {
     severity: 'error',
     reason: 'Extension used by file does not match allowed extensions for its suffix',
   },
+  NON_UNIQUE_DATAFILE: {
+    severity: 'error',
+    reason: 'Multiple files found with the same entities, datatype and suffix.'
+  },
   JSON_KEY_REQUIRED: {
     severity: 'error',
     reason: 'A JSON flle is missing a key listed as required.',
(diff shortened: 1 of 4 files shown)


effigies (commented):
We should check that `.eeg+.vhdr+.vmrk` don't trigger this.

Also, there is https://github.com/bids-standard/bids-specification/pull/1492. Might be worth getting that working again and using that? I believe the schema was in good shape and the rendering code just gathered merge conflicts.

effigies:
@rwblair Please reopen on bids-validator.

PR #2078, opened 2024-08-13
feat: Separate output issues by subcodes

rwblair:
Includes #2075

### DIFF
--- a/bids-validator/src/issues/datasetIssues.test.ts
+++ b/bids-validator/src/issues/datasetIssues.test.ts
@@ -8,7 +8,6 @@ Deno.test('DatasetIssues management class', async (t) => {
     new DatasetIssues()
   })
   await t.step('add Issue throws an error with bad error code', () => {
-    // This mostly tests the issueFile mapping function
     const issues = new DatasetIssues()
     assertThrows(() => {
       issues.add({ code: '__NOT_A_REAL_CODE__' })
@@ -16,7 +15,6 @@ Deno.test('DatasetIssues management class', async (t) => {
   })
 
   await t.step('add Issue with several kinds of files', () => {
-    // This mostly tests the issueFile mapping function
     const issues = new DatasetIssues()
     const testStream = new ReadableStream()
     const text = () => Promise.resolve('')
--- a/bids-validator/src/issues/datasetIssues.test.ts
+++ b/bids-validator/src/issues/datasetIssues.test.ts
@@ -71,12 +71,4 @@ Deno.test('DatasetIssues management class', async (t) => {
     assert(code1 !== undefined)
     assertEquals(code1.size, 2)
   })
-  await t.step('test multiple groupBy', () => {
-    const issues = new DatasetIssues()
-    issues.add({ code: 'NOT_INCLUDED', subCode: 'sub1', location: '/file_1' })
-    issues.add({ code: 'NOT_INCLUDED', subCode: 'sub2', location: '/file_1' })
-    issues.add({ code: 'NOT_INCLUDED', subCode: 'sub1', location: '/file_2' })
-    issues.add({ code: 'NOT_INCLUDED', subCode: 'sub2', location: '/file_2' })
-    issues.add({ code: 'EMPTY_FILE', location: '/file_1' })
-  })
 })
--- a/bids-validator/src/issues/datasetIssues.test.ts
+++ b/bids-validator/src/issues/datasetIssues.test.ts
@@ -78,6 +78,5 @@ Deno.test('DatasetIssues management class', async (t) => {
     issues.add({ code: 'NOT_INCLUDED', subCode: 'sub1', location: '/file_2' })
     issues.add({ code: 'NOT_INCLUDED', subCode: 'sub2', location: '/file_2' })
     issues.add({ code: 'EMPTY_FILE', location: '/file_1' })
-    console.log(issues._groupBy(['subCode', 'code']))
   })
 })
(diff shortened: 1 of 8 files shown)


bids-validator/src/issues/datasetIssues.test.ts
```diff
@@ -73,4 +71,13 @@ Deno.test('DatasetIssues management class', async (t) => {
...
     assertEquals(code1.size, 2)
   })
+  await t.step('test multiple groupBy', () => {
+    const issues = new DatasetIssues()
+    issues.add({ code: 'NOT_INCLUDED', subCode: 'sub1', location: '/file_1' })
+    issues.add({ code: 'NOT_INCLUDED', subCode: 'sub2', location: '/file_1' })
+    issues.add({ code: 'NOT_INCLUDED', subCode: 'sub1', location: '/file_2' })
+    issues.add({ code: 'NOT_INCLUDED', subCode: 'sub2', location: '/file_2' })
+    issues.add({ code: 'EMPTY_FILE', location: '/file_1' })
+    console.log(issues._groupBy(['subCode', 'code']))
```
effigies: We don't want to log this, correct?

effigies (commented):
An extra newline after each error will make it easier to read.

image

I'm also not sure that including the schema path in the pretty output is helpful. We could possibly add `--detail` or `--verbose` flags for people who do want it, but I think it's cluttering.

Similarly, I don't think the sidecar definition is helpful:

image

I'd be inclined to drop it or hide it behind a detail flag.

bids-validator/src/issues/datasetIssues.test.ts
rwblair: No we don't

bids-validator/src/issues/datasetIssues.test.ts
```diff
@@ -73,4 +71,12 @@ Deno.test('DatasetIssues management class', async (t) => {
...
     assertEquals(code1.size, 2)
   })
+  await t.step('test multiple groupBy', () => {
+    const issues = new DatasetIssues()
+    issues.add({ code: 'NOT_INCLUDED', subCode: 'sub1', location: '/file_1' })
+    issues.add({ code: 'NOT_INCLUDED', subCode: 'sub2', location: '/file_1' })
+    issues.add({ code: 'NOT_INCLUDED', subCode: 'sub1', location: '/file_2' })
+    issues.add({ code: 'NOT_INCLUDED', subCode: 'sub2', location: '/file_2' })
+    issues.add({ code: 'EMPTY_FILE', location: '/file_1' })
+  })
```
effigies: I think you're missing assertions here.

bids-validator/src/schema/applyRules.ts
```diff
@@ -446,92 +446,90 @@ function evalJsonCheck(
...
+          }
+          if (code) {
+            context.dataset.issues.add({
+              code,
+              subCode: keyName,
+              location: context.path,
+              severity,
+              rule: schemaPath,
+            })
+          }
```
effigies: Just to shave off a few lines:

```suggestion
          const keyType = sidecarRule ? 'SIDECAR_KEY' : 'JSON_KEY'
          const level = severity === 'error' ? 'REQUIRED' : 'RECOMMENDED'
          context.dataset.issues.add({
            `${keyType}_${level}`,
            subCode: keyName,
            location: context.path,
            severity,
            rule: schemaPath,
          })
```

rwblair:
Needs the help url per issue.

bids-validator/src/schema/applyRules.ts
```diff
@@ -446,92 +446,90 @@ function evalJsonCheck(
...
-            issueMessage: `${messageBase} ${message}`,
+            location: location,
+            issueMessage: message,
             rule: schemaPath,
-            affects: [originFile],
+            affects: affects,
           })
         }
       }
     }
```
effigies: Nit-pick (feel free to defer to another PR): I don't think the `!validate.errors` branch is possible to hit. (Coverage indicates it is never hit in our tests.)

Following the lead of https://ajv.js.org/api.html#error-parameters, we could simplify this section:

```ts
const validate = compile(metadataDef)
if (!validate(value)) {
  for (const err of validate.errors as DefinedError[]) {
    context.dataset.issues.add({
      code: 'JSON_SCHEMA_VALIDATION_ERROR',
      subCode: keyName,
      issueMessage: err['message']
      rule: schemaPath,
      location,
      affects,
    })
  }
}

**Review by effigies — APPROVED** [review]
Only a couple remaining points. LGTM.

File: bids-validator/src/schema/applyRules.ts [hunk 5]
```diff
@@ -446,92 +446,85 @@ function evalJsonCheck(
...
+            severity,
+            rule: schemaPath,
+          }, requirement.issue.message)
+        } else {
+          let code
+          const keyType = sidecarRule ? 'SIDECAR_KEY' : 'JSON_KEY'
+          const level = severity === 'error' ? 'REQUIRED' : 'RECOMMENDED'
           context.dataset.issues.add({
-            code,
+            `${keyType}_${level}`,
```
effigies: Sorry.

```suggestion
            code: `${keyType}_${level}`,
```

bids-validator/src/schema/applyRules.ts
rwblair: `!validate.errors` was a hedge because I couldn't be positive that validate would always return an entry in errors going off the ValidationFunction type from AJV. I've never seen it not return something. Since the AJV documentation seems confident something will always be there I'll update it to your suggestion.

bids-validator/src/issues/datasetIssues.test.ts
rwblair: For a feature I reneged on in this PR. I removed them.

PR #2080, opened 2024-08-14
Throw error on multiple valid inheritable files in the same directory...

### LINKED ISSUE #2037: Raise validation error if multiple applicable files are found in one directory via inheritance
From the spec:

> ### Rules
> 
> 1.  Any metadata file (such as `.json`, `.bvec` or `.tsv`) MAY be defined at any directory level.
> 
> 1.  For a given data file, any metadata file is applicable to that data file if:
>     1.  It is stored at the same directory level or higher;
>     1.  The metadata and the data filenames possess the same suffix;
>     1.  The metadata filename does not include any entity absent from the data filename.
> 
> 1.  A metadata file MUST NOT have a filename that would be otherwise applicable
>     to some data file based on rules 2.b and 2.c but is made inapplicable based on its
>     location in the directory structure as per rule 2.a.
> 
> 1.  There MUST NOT be multiple metadata files applicable to a data file at one level
>     of the directory hierarchy.
> 
> 1.  If multiple metadata files satisfy criteria 2.a-c above:
> 
>     1.  For [tabular files](#tabular-files) and other simple metadata files
> 	(for instance,
> 	[`bvec` / `bval` files for diffusion MRI](modality-specific-files/magnetic-resonance-imaging-data.md#required-gradient-orientation-information)),
> 	accessing metadata associated with a data file MUST consider only the
> 	applicable file that is lowest in the filesystem hierarchy.
> 
>     1.  For [JSON files](#key-value-files-dictionaries), key-values are loaded
> 	from files from the top of the directory hierarchy downwards, such that
> 	key-values from the top level are inherited by all data files at lower
> 	levels to which it is applicable unless overridden by a value for the
> 	same key present in another metadata file at a lower level
> 	(though it is RECOMMENDED to minimize the extent of such overrides).
> 
> ### Corollaries
> 
> 1.  As per rule 3, metadata files applicable only to a specific participant / session
>     MUST be defined in or below the directory corresponding to that participant / session;
>     similarly, a metadata file that is applicable to multiple participants / sessions
>     MUST NOT be placed within a directory corresponding to only one such participant / session.
> 
> 1.  It is permissible for a single metadata file to be applicable to multiple data
>     files at that level of the hierarchy or below. Where such metadata content is consistent
>     across multiple data files, it is RECOMMENDED to store metadata in this
>     way, rather than duplicating that metadata content across multiple metadata files.
> 
> 1.  Where multiple applicable JSON files are loaded as per rule 5.b, key-values can
>     only be overwritten by files lower in the filesystem hierarchy; the absence of
>     a key-value in a later file does not imply the "unsetting" of that field
>     (indeed removal of existing fields is not possible).

----

              This should really be an issue, as the presence of multiple applicable files at a level violates the spec.

_Originally posted by @effigies in https://github.com/bids-standard/bids-validator/pull/2035#discussion_r1697793907_

rwblair:
...and catch and store as a dataset issue.

fixes #2037

### DIFF
--- a/bids-validator/src/files/inheritance.test.ts
+++ b/bids-validator/src/files/inheritance.test.ts
@@ -6,9 +6,19 @@ import { nullFile } from '../tests/nullFile.ts'
 
 Deno.test('walkBack throws multiple inheritance error', async (t) => {
   const rootFileTree = new FileTree('/', '')
-  rootFileTree.files.push({ path: '/', name: 'sub-01_T1w.json', ...nullFile, parent: rootFileTree})
-  rootFileTree.files.push({ path: '/', name: 'T1w.json', ...nullFile, parent: rootFileTree})
-  const dataFile =   { path: '/', name: 'sub-01_acq-test_T1w.nii', ...nullFile, parent: rootFileTree}
+  rootFileTree.files.push({
+    path: '/',
+    name: 'sub-01_T1w.json',
+    ...nullFile,
+    parent: rootFileTree,
+  })
+  rootFileTree.files.push({ path: '/', name: 'T1w.json', ...nullFile, parent: rootFileTree })
+  const dataFile = {
+    path: '/',
+    name: 'sub-01_acq-test_T1w.nii',
+    ...nullFile,
+    parent: rootFileTree,
+  }
   rootFileTree.files.push(dataFile)
   assertThrows(() => {
     try {
--- a/bids-validator/src/files/inheritance.test.ts
+++ b/bids-validator/src/files/inheritance.test.ts
@@ -1,0 +1,24 @@
+import { assertEquals, assertThrows } from '../deps/asserts.ts'
+import { FileTree } from '../types/filetree.ts'
+import { BIDSFileDeno } from '../files/deno.ts'
+import { walkBack } from './inheritance.ts'
+import { nullFile } from '../tests/nullFile.ts'
+
+Deno.test('walkBack throws multiple inheritance error', async (t) => {
+  const rootFileTree = new FileTree('/', '')
+  rootFileTree.files.push({ path: '/', name: 'sub-01_T1w.json', ...nullFile, parent: rootFileTree})
+  rootFileTree.files.push({ path: '/', name: 'T1w.json', ...nullFile, parent: rootFileTree})
+  const dataFile =   { path: '/', name: 'sub-01_acq-test_T1w.nii', ...nullFile, parent: rootFileTree}
+  rootFileTree.files.push(dataFile)
+  assertThrows(() => {
+    try {
+      const sidecars = walkBack(dataFile)
+      for (const f of sidecars) {
+        continue
+      }
+    } catch (error) {
+      assertEquals(error.code, 'MULTIPLE_INHERITABLE_FILES')
+      throw error
+    }
+  })
+})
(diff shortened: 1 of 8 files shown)


bids-validator/src/files/inheritance.ts
```diff
@@ -23,22 +23,11 @@ export function* walkBack(
...
     if (candidates.length > 1) {
-      const exactMatch = candidates.find((file) => {
-        const { suffix, extension, entities } = readEntities(file.name)
-        return Object.keys(sourceParts.entities).every((entity) =>
-          entities[entity] === sourceParts.entities[entity]
-        )
-      })
-      if (exactMatch) {
-        exactMatch.viewed = true
-        yield exactMatch
```
effigies: I think we do want the `exactMatch` path still. Especially in derivatives. In derivatives, you might have the same file with `space-<label>` and without, each with their own JSON. It's perfectly unambiguous following the exact match rule.

bids-validator/src/files/inheritance.test.ts
```diff
@@ -0,0 +1,34 @@
+import { assertEquals, assertThrows } from '../deps/asserts.ts'
+import { FileTree } from '../types/filetree.ts'
+import { BIDSFileDeno } from '../files/deno.ts'
+import { walkBack } from './inheritance.ts'
+import { nullFile } from '../tests/nullFile.ts'
+
+Deno.test('walkBack throws multiple inheritance error', async (t) => {
+  const rootFileTree = new FileTree('/', '')
+  rootFileTree.files.push({
```
effigies, line 9: This whole thing makes me think I should go ahead and finish my branch making a `FileTree` from a list of filenames. This should be a lot easier.

bids-validator/src/schema/context.ts
```diff
@@ -170,13 +170,21 @@ export class BIDSContext implements Context {
       return
     }
     const sidecars = walkBack(this.file)
```
effigies: I like to keep try/catches very small what about this instead:

```suggestion
    try {
      const sidecars = [...walkBack(this.file)]
    } catch (error) {
      if (error.code === 'MULTIPLE_INHERITABLE_FILES') {
        this.dataset.issues.add(error)
      } else {
        throw error
      }
    }
```

bids-validator/src/schema/context.ts
```diff
@@ -170,13 +170,21 @@ export class BIDSContext implements Context {
...
+        })
+        this.sidecar = { ...json, ...this.sidecar }
+        Object.keys(json).map((x) => this.sidecarKeyOrigin[x] ??= file.path)
+      }
+    } catch (error) {
+      if (error.code === 'MULTIPLE_INHERITABLE_FILES') {
+        this.dataset.issues.add(error)
+      } else {
+        throw error
+      }
```
effigies: Then this whole section can be reverted.

bids-validator/src/schema/context.ts
```diff
@@ -211,7 +219,15 @@ export class BIDSContext implements Context {
   }
 
   async loadAssociations(): Promise<void> {
-    this.associations = await buildAssociations(this.file, this.dataset.issues)
+    try {
+      this.associations = await buildAssociations(this.file, this.dataset.issues)
```
effigies: We pass `this.dataset.issues` to `buildAssociations`, so, if we catch in there, we could provide an error for each association that has multiple inheritable files.

rwblair: I totally missed that. x_x

effigies (approved):

PR #2082, opened 2024-08-15
feat: Utilities for constructing BIDSFiles and FileTrees

effigies:
We were starting to excessively reimplement creation of BIDSFiles, and creating FileTrees is cumbersome. I had started to abstract out the FileTree generation from the browser implementation last week, so I finished that.

`src/files/filetree.ts` exports:

* `pathToFile(path: string): BIDSFile`
* `filesToTree(fileList: BIDSFile[]): FileTree`

And a convenience function that simply combines the two in the obvious way:

```typescript
export function pathsToTree(paths: string[]): FileTree {
  return filesToTree(paths.map(pathToFile))
}
```


bids-validator/src/files/browser.test.ts
```diff
@@ -28,7 +28,7 @@ Deno.test('Browser implementation of FileTree', async (t) => {
       new TestFile(['flat test dataset'], 'README.md', 'ds/README.md'),
     ]
     const tree = await fileListToTree(files)
```
effigies:
```suggestion
    const tree = fileListToTree(files)
```

bids-validator/src/files/browser.test.ts
```diff
@@ -61,9 +61,9 @@ Deno.test('Browser implementation of FileTree', async (t) => {
       ),
     ]
     const tree = await fileListToTree(files)
```
effigies:
```suggestion
    const tree = fileListToTree(files)
```

bids-validator/src/files/browser.test.ts
```diff
@@ -90,7 +90,7 @@ Deno.test('Spread copies of BIDSFileBrowser contain name and path properties', a
     new TestFile(['flat test dataset'], 'README.md', 'ds/README.md'),
   ]
   const tree = await fileListToTree(files)
```
effigies:
```suggestion
  const tree = fileListToTree(files)
```

effigies (commented):
Missed these.

web/src/App.tsx
```diff
@@ -42,7 +42,7 @@ function App() {
     const dirHandle = await directoryOpen({
       recursive: true,
     })
-    const fileTree = await fileListToTree(dirHandle)
```
rwblair, line 45: I had to add this await back in, fileListToTree was returning a promise. Could be a me problem though.

bids-validator/src/files/browser.ts
```diff
@@ -48,40 +49,8 @@ export class BIDSFileBrowser implements BIDSFile {
 /**
  * Convert from FileList (created with webkitDirectory: true) to FileTree for validator use
  */
-export function fileListToTree(files: File[]): Promise<FileTree> {
+export function fileListToTree(files: File[]): FileTree {
```
effigies, line 52: It seems like it shouldn't be?

rwblair: Ah yeah didn't notice it getting sourced from web/dist/main.ts My build.ts isn't working and dist must be outdated.

effigies: Yeah, you'll need #2088 before this will work.

effigies:
Rebased on #2088 for easier testing.

PR #2090, opened 2024-08-16
rf: Use upstream context types

effigies:
It can be done!

~~I'm going to try to implement `Gzip`, `Ome`, and `Tiff`. Will push to this branch if it isn't merged when I finish them, but feel free to merge this when it's as complex as you want to allow one PR to get.~~

Making separate PRs. The OME-TIFF one is a bit of a pain.


bids-validator/src/files/gzip.ts
```diff
@@ -0,0 +1,45 @@
...
+   *
+   * const hasComment = flags & 0x10 // FCOMMENT
+   * let comment: string | undefined
+   * if (hasComment) {
+   *   const start = offset
+   *   offset++ 
+   *   while (offset < buf.length && buf[offset] !== 0) offset++
+   *   comment = new TextDecoder().decode(buf.subarray(start, offset))
+   * }
+   */
```
effigies: Actually, the schema checks that comment == "".  We can either remove that from the schema or try to parse it.

PR #2092, opened 2024-08-17
feat: Parse gzip header

effigies:
Builds on #2090.


bids-validator/src/files/gzip.ts
```diff
@@ -0,0 +1,53 @@
...
+ * @param file - The file to parse
+ * @param maxBytes - The maximum number of bytes to read
+ */
+export async function parseGzip(
+  file: BIDSFile,
+  maxBytes: number = 512,
+): Promise<Gzip | undefined> {
+  const buf = await file.readBytes(maxBytes)
+  const view = new DataView(buf.buffer, buf.byteOffset, buf.byteLength)
+  if (view.getUint16(0, false) !== 0x1f8b) return undefined
```
rwblair: This throws the error `RangeError: offset is outside the bounds of the DataView` for empty files. We should skip empty/undersized files to reduce `console.debug` output.

effigies: What about:

```suggestion
  if (view.byteLength < 2 || view.getUint16(0, false) !== 0x1f8b) return undefined
```

PR #2093, opened 2024-08-17
feat: Parse (OME-)TIFF

effigies:
I initially tried to use exifreader, but that requires some post-install script, which Deno needs coaxing to use. I don't want to push that on users, so I wrote a quick lookup function for the ImageDescription tag. Verified it works on TIFF and BigTIFF.

Could possibly write this to stream instead of reading 4KiB unconditionally. At least in our examples we only need ~1.4KB. Seems fine for now.


effigies:
Looks like Vite does not like the way esbuild bundles `@libs/xml`. Will try to figure this out, but if you already know, @nellh, I wouldn't mind a hint. :-)

effigies:
Gave up on `@libs/xml` for now.

PR #2098, opened 2024-08-19
feat(cli): Add --blacklistModalities and --ignoreWarnings flags

effigies:
A quick test got thrown in here and I don't really care to rebase it out.


rwblair:
Behavioral can be a bit of a sticky wicket, can look at adding it if someone needs it.

PR #2102, opened 2024-08-21
Add flag for validating derivatives inside /derivatives/

rwblair:
Makes it so nested derivatives are *not* validated by default. Adds a recursive flag to re-enable old behavior.


effigies (approved):

rwblair:
@yarikoptic Are you currently using the 'derivativesSummary' returned by the schema validator? Don't want to surprise you with a change in default behavior.

yarikoptic:
@rwblair I do not think I am that advanced of a user yet! I do very appreciate asking though!

PR #2104, opened 2024-08-21
feat(tsv): Warn on undefined columns

effigies:
Partially addresses bids-standard/bids-validator#34. Needs a definition for behavioral data in the schema to fully address it.

### DIFF
--- a/bids-validator/src/schema/tables.ts
+++ b/bids-validator/src/schema/tables.ts
@@ -238,7 +238,6 @@ export function evalAdditionalColumns(
 ): void {
   if (context.extension !== '.tsv') return
   const headers = Object.keys(context?.columns)
-  // hard coding allowed here feels bad
   if (rule.columns) {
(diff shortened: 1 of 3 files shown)


bids-validator/src/schema/tables.ts
```diff
@@ -247,16 +247,19 @@ export function evalAdditionalColumns(
...
     )
-    if (rule.additional_columns === 'allowed_if_defined') {
+
+    if (rule.additional_columns?.startsWith('allowed')) {
       extraCols = extraCols.filter((header) => !(header in context.sidecar))
     }
+    const issue = {
+      code: rule.additional_columns === 'allowed'
+        ? 'TSV_ADDITIONAL_COLUMNS_UNDEFINED'
+        : 'TSV_ADDITIONAL_COLUMNS_NOT_ALLOWED',
```
effigies: Do we want a different error for `allowed_if_defined`? We haven't had one previously, but we could be more helpful if we do.

PR #2108, opened 2024-08-22
change xml2js to import from esm.sh

rwblair:
one more way to skin an import


rwblair:
Ok not as much fun trying to make jsr happy.

rwblair:
files acquired with `pipx run download-esm xml2js`

bids-validator/deno.json
```diff
@@ -34,7 +34,7 @@
     "@hed/validator": "npm:hed-validator@3.15.4",
     "@ignore": "npm:ignore@5.3.2",
     "@mango/nifti": "npm:nifti-reader-js@0.6.8",
-    "@xml2js": "npm:xml2js@0.6.2"
+    "stream": "node:stream"
```
effigies, line 37: Is this used?

effigies:
Want to pull in 76410f7915922613e79fb9ac13e0d0fa253634d4?

bids-validator/deno.json
rwblair: Yeah xml2js dependency sax attempts to import it.

rwblair:
`deno run -A dist/validator/bids-validator.js tests/data/bids-examples/micr_SPIM/`

This command finishes validation and prints the summary for me, but never exits. Issue doesn't exist on ds001. Moving a single ome.tif and its json into ds001 doesn't cause any problems. Seems to be stuck in denos `eventLoopTick` for node compatibility.

rwblair:
esbuild version of validator was expecting a proper exit call, which wasn't happening for valid datasets.

PR #2111, opened 2024-08-26
fix: Import schema correctly from JSR

effigies:
Starting with tests that should catch the error in the future.


bids-validator/src/setup/loadSchema.ts
```diff
@@ -1,6 +1,6 @@
 import type { Schema } from '../types/schema.ts'
 import { objectPathHandler } from '../utils/objectPathHandler.ts'
-import * as schemaDefault from '@bids/schema/schema' with { type: 'json' }
+import default as schemaDefault from '@bids/schema/schema' with { type: 'json' }
```
rwblair:
```suggestion
import { default as schemaDefault } from '@bids/schema/schema' with { type: 'json' }
```

PR #2112, opened 2024-08-27
fix: Load schema with fetch, default to JSR package

effigies:
The current version emits a more-or-less obligatory error due to the use of `import` with an https URL. That seems like a thing of Deno past.

Given that we're hosting the schema at the JSR, let's go ahead and default to that. If someone wants to provide a URL, then a fetch can be done and the JSON can be loaded. This means BIDS_SCHEMA pointing to local files will need to use `file:///...`, but that seems fine.


nellh (approved):

effigies:
@nellh Just pushed a commit to remove a default value from the `--schema` CLI flag, squeezed in after your review.

PR #2114, opened 2024-08-28
build: Include exports for extending validator in JSR packaging

nellh:
These exports are used by OpenNeuro's CLI to extend the Deno validator.

### DIFF
--- a/bids-validator/deno.json
+++ b/bids-validator/deno.json
@@ -1,7 +1,13 @@
 {
   "name": "@bids/validator",
   "version": "1.14.10",
-  "exports": "./src/bids-validator.ts",
+  "exports": {
+    ".": "./src/bids-validator.ts",
+    "./main": "./src/main.ts",
+    "./output": "./src/utils/output.ts",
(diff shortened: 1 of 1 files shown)


bids-validator/deno.json
```diff
@@ -1,7 +1,13 @@
 {
   "name": "@bids/validator",
   "version": "1.14.10",
-  "exports": "./src/bids-validator.ts",
+  "exports": {
+    ".": "./src/bids-validator.ts",
+    "./main": "./src/main.ts",
+    "./output": "./src/utils/output.ts",
+    "./files": "./src/files/deno.ts",
```
effigies, line 8: Is there any chance we'd want to expose the browser files as well? Just thinking about backwards compatibility here.

nellh: It is exported already via main. I think if we wanted to add browser features to the files API later, we would just point it at a new source file and re-export both.

effigies:
Do you have a way to test this kind of thing without pushing to JSR, by the way?

effigies (approved):

PR #2116, opened 2024-08-29
feat: Validate CITATION.cff

### LINKED ISSUE #1959: CITATION.cff support
To support CITATION.cff we need:

* [ ] Load CITATION.cff as YAML, fail if invalid
* [ ] Validate with [JSON schema](https://github.com/citation-file-format/citation-file-format/blob/main/schema.json) ([raw link](https://raw.githubusercontent.com/citation-file-format/citation-file-format/main/schema.json)). We may need/want to maintain a fork of this to support roles until they include them.
* [ ] Populate context (schema validator) with relevant fields for validation checks (needs coordinated PR in spec).

xref https://github.com/bids-standard/bids-specification/issues/901 https://github.com/bids-standard/bids-examples/pull/426

rwblair:
Dataset level check to validate citation.cff. Imports citation.cff json schema from @bids/schema package, convert citation.cff to json, and call ajv on it with stock ajv-formats.

Closes #1959.

### DIFF
--- a/bids-validator/src/validators/citation.ts
+++ b/bids-validator/src/validators/citation.ts
@@ -2,12 +2,10 @@ import type { GenericSchema } from '../types/schema.ts'
 import type { BIDSFile, FileTree } from '../types/filetree.ts'
 import type { BIDSContextDataset } from '../schema/context.ts'
 import { schema as citationSchema } from '@bids/schema/citation'
-import { Ajv, type DefinedError } from '@ajv'
-import _addFormats from '@ajv-formats'
+import { compile } from './json.ts'
+import type { DefinedError } from '@ajv'
 import { parse } from '@std/yaml'
 
-// https://github.com/ajv-validator/ajv-formats/issues/85
-const addFormats = _addFormats as unknown as typeof _addFormats.default
 const citationFilename = 'CITATION.cff'
 
 export async function citationValidate(
@@ -23,9 +21,7 @@ export async function citationValidate(
     throw error
     return
   }
-  const ajv = new Ajv()
-  addFormats(ajv)
-  const citationValidate = ajv.compile(citationSchema)
+  const citationValidate = compile(citationSchema)
   if (!citationValidate(citation)) {
     for (const err of citationValidate.errors as DefinedError[]) {
       dsContext.issues.add({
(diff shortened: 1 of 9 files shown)


bids-validator/src/validators/citation.ts
```diff
@@ -0,0 +1,34 @@
...
+  schema: GenericSchema,
+  dsContext: BIDSContextDataset,
+) {
+  const citationFile = dsContext.tree.get(citationFilename)
+  if (!citationFile || 'directories' in citationFile) return
+  let citation: unknown = {}
+  try {
+    citation = parse(await citationFile.text())
+  } catch (error) {
+    throw error
```
effigies: I assume we want to add our own issue here, instead?

bids-validator/src/validators/citation.ts
```diff
@@ -0,0 +1,34 @@
...
+    return
+  }
+  const citationValidate = compile(citationSchema)
+  if (!citationValidate(citation)) {
+    for (const err of citationValidate.errors as DefinedError[]) {
```
effigies: Having the same name as the containing function makes this feel like it should be recursive, scoping rules be damned.

```suggestion
  const validate = compile(citationSchema)
  if (!validate(citation)) {
    for (const err of validate.errors as DefinedError[]) {
```

bids-validator/src/validators/citation.ts
```diff
@@ -0,0 +1,34 @@
...
+      dsContext.issues.add({
+        code: 'JSON_SCHEMA_VALIDATION_ERROR',
```
effigies: We can give this a subcode so people know what field is failing:

```suggestion
        code: 'JSON_SCHEMA_VALIDATION_ERROR',
        subCode: err['instancePath'],
```

I'm also a little inclined to give this its own code, since it's not a JSON file:

```suggestion
        code: 'CITATION_CFF_VALIDATION_ERROR',
        subCode: err['instancePath'],
```

bids-validator/src/validators/citation.ts
rwblair: Good call. I had named the the outer function something different, then changed it to this collision to match more of the functions in validator.

bids-validator/src/validators/citation.ts
rwblair: Whoops yeah for got to fill that in. Was thinking of creating a generic file read error if we didn't have one already.

rwblair: We do have one, FILE_READ from schema.rules.errors

effigies (approved):
LGTM. CI failures should be resolved in #2117.

PR #2131, opened 2024-09-09
Add option to write validation results to file.

### LINKED ISSUE #829: [wishlist] -o|--output option to specify file to store output in
although it is trivial to implement and workaround on any POSIX system with directing output into a file (i.e. `>| validator.log`), my use case is to use [datalad-containers](https://github.com/datalad/datalad-container/) datalad extension. Since container has the entry point which runs  bids-validator already, it makes it impossible (without making default entry point to be just `bash`) to make invocation to `datalad containers-run` to specify output file.  If there was an `-o` option it would have looked like `datalad containers-run -n containers/bids-validator --input sourcedata --output sourcedata-validator.log '{inputs}' -o '{outputs}'` .

while storing into file there would be no ANSI coloring, it could stay consistent with having --verbose in a regular invocation.

rwblair:
Only works in Deno. Also makes black holing nifit library output less black holey.

Closes https://github.com/bids-standard/bids-validator/issues/829


nellh (approved):

bids-validator/src/main.ts
```diff
@@ -27,26 +27,24 @@ export async function main(): Promise<ValidationResult> {
...
-      }),
-    )
+    output_string = consoleFormat(schemaResult, {
+      verbose: options.verbose ? options.verbose : false,
+    })
+  }
+
+  if (options.outfile) {
+    if (globalThis.Deno) {
+      Deno.writeTextFileSync(options.outfile, output_string)
```
yarikoptic, line 41: I am just curious - is writing to a file is a challenge in original Node???

rwblair: No, but it is a challenge in a browser. This code is currently ever only run by browsers and Deno.

yarikoptic: ah, I see -- it was not "deno vs legacy" (which is on my mind when "not deno") it was "deno vs in-browser"!  Thank you @rwblair for the explanation!

PR #2149, opened 2024-10-03
feat: Pass hedCode as subcode to group errors

effigies:
cc @happy5214 In case you have any other suggestions.


happy5214:
Is `subcode` currently a valid parameter, and will it be added if it isn't?

effigies:
It is:

https://github.com/bids-standard/bids-validator/blob/f85d51f2b710be07137e6147388ed71461c369f5/bids-validator/src/types/issues.ts#L8-L19

Note this is for the schema validator, which has a new issue format.

PR #2152, opened 2024-10-08
Pass value of ignore test to creation of filetree object.

### LINKED ISSUE #2127: code/ and sourcedata/ folders should be ignored and not demanded to follow BIDS naming convention or other requirements
with deno validator on https://openneuro.org/datasets/ds005256 which has git submodules for various components (code, sourcedata) validator tries to impose its will:

```
    [ERROR] EMPTY_FILE Empty files not allowed.
        /sourcedata/d_beh/
```
and we have that empty folder since it is placeholder for submodule, and I do not think anything in BIDS forbid that.  At most it should be a `HINT` of some kind but even that not sure why to hint about.  Note that it is a folder not a "file" so message is a bit off too.

```
        [ERROR] NOT_INCLUDED Files with such naming scheme are not part of BIDS specification. This error is most commonly caused by typos in file names that make them not BIDS compatible. Please consult the specification and make sure your files are named correctly. If this is not a file naming issue (for example when including files not yet covered by the BIDS specification) you should include a ".bidsignore" file in your dataset (see https://github.com/bids-standard/bids-validator#bidsignore for details). Please note that derived (processed) data should be placed in /derivatives folder and source data (such as DICOMS or behavioural logs in proprietary formats) should be placed in the /sourcedata folder.
                /code/spacetop-prep/.github/ISSUE_TEMPLATE/
                /code/spacetop-prep/spacetop_prep/

                1 more files with the same issue
```

again -- `code/spacetop-prep` is submodule and nothing in BIDS regulates content of `code/`, so it is very unclear why validator is trying to enforce bids naming convention there.

rwblair:
Certain directories get contexts during calls to walk, ignore test needs to be run and set before then. Noticed certain directories in sourcedata of ds004720 triggering NOT_INCLUDED errors.

I still need to write tests for this.

Fixes https://github.com/bids-standard/bids-validator/issues/2127.

### DIFF
--- a/bids-validator/src/files/deno.ts
+++ b/bids-validator/src/files/deno.ts
@@ -119,7 +119,7 @@ async function _readFileTree(
 ): Promise<FileTree> {
   await requestReadPermission()
   const name = basename(relativePath)
-  const tree = new FileTree(relativePath, name, parent)
+  const tree = new FileTree(relativePath, name, parent, ignore.test(join(rootPath, relativePath)))
 
   for await (const dirEntry of Deno.readDir(join(rootPath, relativePath))) {
     if (dirEntry.isFile || dirEntry.isSymlink) {
--- a/bids-validator/src/files/deno.ts
+++ b/bids-validator/src/files/deno.ts
@@ -2,7 +2,6 @@
  * Deno specific implementation for reading files
  */
 import { basename, join } from '@std/path'
-import { existsSync } from '@std/fs'
 import * as posix from '@std/path/posix'
 import { type BIDSFile, FileTree } from '../types/filetree.ts'
 import { requestReadPermission } from '../setup/requestPermissions.ts'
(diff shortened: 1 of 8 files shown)


effigies:
Should we just prune `sourcedata/` and `code/` in the same way we do derivatives, except unconditionally?

bids-validator/src/files/deno.ts
```diff
@@ -119,7 +119,7 @@ async function _readFileTree(
 ): Promise<FileTree> {
   await requestReadPermission()
   const name = basename(relativePath)
-  const tree = new FileTree(relativePath, name, parent)
+  const tree = new FileTree(relativePath, name, parent, ignore.test(join(rootPath, relativePath)))
```
effigies: I'm not sure this will be sufficient, since we could recurse into a directory before reading `.bidsignore` from the first directory. I think we might instead need to make `FileTree` do an active lookup into an `ignore` object like `BIDSFileDeno` does above.

effigies:
Another approach could be to adapt:

https://github.com/bids-standard/bids-validator/blob/ae54efa0feb080f7ee42494e12c723cd234acc65/bids-validator/src/schema/walk.ts#L42-L44

To:

```diff
-    if (dsContext.isPseudoFile(dir)) {
+    if (dsContext.isPseudoFile(dir) && !ignore.test(dir.path)) {
       yield new BIDSContext(pseudoFile(dir), dsContext)
     } else {
```

We would need to expose the `ignore` object somewhere outside the `BIDSFile`s to make this work, but maybe that's better than the current approach, where we inject a singleton and then update it before we lose its reference.

rwblair:
Yeah I'm not a fan of how ignore is being managed right now. I'll play with loading bidsignore earlier and storing in the dataset context like other configuration options.

Should a bidsignore file in a root dataset apply to nested datasets?

effigies:
> Should a bidsignore file in a root dataset apply to nested datasets?

I don't think so.

rwblair:
This latest commit fails, cli and web, on ds001583, which has its .bidsignore as a directory instead of a file. Guess I'll make it survive any error but log/warn on non NOENT errors. ds001583 is only dataset I've found with a .bidsignore as a directory.

effigies (approved):

Issue #2130, opened 2024-09-09
deno: error: Uncaught (in promise) Error: key: undefined does not exist in non-schema issues definitions


yarikoptic:
installed/ran as in
- #2129 

resulted in a crash without any report

under `/dartfs-hpc/rc/lab/C/CANlab/labdata/data/spacetop/dartmouth` 
```shell
[d31548v@discovery-01 dartmouth]$ fg
singularity exec -B ~/.var/deno-dir:/deno-dir /tmp/deno_latest.sif deno run --allow-read --allow-env --reload -A https://github.com/bids-standard/bids-validator/raw/deno-build/bids-validator.js $PWD
error: Uncaught (in promise) Error: key: undefined does not exist in non-schema issues definitions
      if (!__hasOwnProp.call(to, key) && key !== except)
              ^
    at _DatasetIssues.add (https://raw.githubusercontent.com/bids-standard/bids-validator/deno-build/bids-validator.js:24:15)
    at https://raw.githubusercontent.com/bids-standard/bids-validator/deno-build/bids-validator.js:59:24
    at eventLoopTick (ext:core/01_core.js:175:7)
    at async validate (https://raw.githubusercontent.com/bids-standard/bids-validator/deno-build/bids-validator.js:58:37)
    at async main (https://raw.githubusercontent.com/bids-standard/bids-validator/deno-build/bids-validator.js:28:24)
    at async https://raw.githubusercontent.com/bids-standard/bids-validator/deno-build/bids-validator.js:3:16

```

mateuszpawlik:
A quick note: I've just encountered this error with the newest deno docker image converted to Apptainer. With deno:1.45.5 the execution works.

My command:
```
apptainer run --bind $PWD:/data:ro --bind ~/Downloads/deno-dir:/deno-dir ../deno_1.45.5.sif deno run -A jsr:@bids/validator -v /data
```

effigies:
Seems like a problem upstream in deno.

effigies:
Nevermind, got basically the same thing in deno 2.0, so this may be a problem in the future, even if it's Deno's fault.

PR #2157, opened 2024-10-10
fix: Check for expected error types, rethrow unknown

effigies:
Partially addresses #2130. @rwblair is working on fixing the real error.


effigies:
@rwblair This should go in before a release.

PR #2161, opened 2024-10-14
Fixes to allow Deno check to pass and build to succeed. 

rwblair:
Add further type checks to catching errors that are thrown issues (then immediately ts-ignore the issues.add).
I think to remove the @ts-expect-error  on the issue.adds we'd need to have issues be a class, instantiate instances in the throw statement then we could leverage instanceof in the catch statements. 

I tried to put the type guard into a function to reduce how often all those && were being repeated but TS wasn't having any of it. Not sure what I was doing wrong. Similarly TS didn't like `hasOwn` so had to use `in` operator. 

Update Deno.run to Deno.Command, run removed in deno2.

Removes stub file validators/isBidsy.ts.


.github/workflows/deno_tests.yml
```diff
@@ -22,10 +22,10 @@ jobs:
       - name: Save describe stamp
         id: describe
         run: echo version=$( git describe ) >> $GITHUB_OUTPUT
-      - uses: denoland/setup-deno@v1
+      - uses: denoland/setup-deno@v2
         with:
-          deno-version: v1.x
-      - run: ./build.ts
+          deno-version: v2.x
+      - run: deno --node-modules-dir=auto --allow-env --allow-read --allow-run --allow-net ./build.ts
```
effigies:
```suggestion
      - run: deno --node-modules-dir=auto -A ./build.ts
```

bids-validator/src/version.ts
```diff
@@ -34,7 +34,7 @@ export async function getVersion(): Promise<string> {
 async function getLocalVersion(path: string): Promise<string> {
   // safe.directory setting so we could still operate from another user
   const command = new Deno.Command("git", {
-    args: ['git', '-C', path, '-c', 'safe.directory=*', 'describe', '--tags', '--always'],
+    args: ['-C', path, '-c', 'safe.directory=*', 'describe', '--tags', '--always'],
   })
```
effigies: We should catch errors here so we don't crash if `git` isn't present.

rwblair: I'll add one. .git is not added to repo so if we catch this we can remove git as a dependency.

Issue #1930, opened 2024-04-13
Make -s|--schema take path or add new --schema-file option to point to schema.json file


yarikoptic:
ATM the option seems to take only some hardcoded set of versions

```
  -s, --schema                    BIDS specification schema version to use for
                                  validation, e.g. "v1.6.0" (beta)
         [choices: "disable", "v1.6.0", "v1.7.0", "master"] [default: "disable"]
```

(note: current release is 1.9.0!).

That precludes to point to a specific `schema.org` e.g. produced while working on a BEP. Ref:

- bids-standard/bids-validator#24 

Quick resolution would be appreciated to proceed with

- https://github.com/bids-standard/bids-specification/pull/1775

effigies:
The schema validator shows:

```
❯ deno run --allow-read --allow-env https://deno.land/x/bids_validator/bids-validator.ts --help

  Usage:   bids-validator <dataset_directory>
  Version: alpha                             

[...]

    -s, --schema          <type>  - Specify a schema version to use for validation                          (Default: "latest")   
```

This appears to be an experimental addition to the legacy validator. Probably best just to remove it.

yarikoptic:
But how to specify not by "version" but by the path to the schema???

effigies:
`deno run --allow-read --allow-env --allow-net https://deno.land/x/bids_validator/bids-validator.ts -s https://bids-specification.readthedocs.io/en/v1.9.0/schema.json /path/to/dataset`

yarikoptic:
oh, will it work with `file://` URLs? (happen you could check quickly... if not -- I will try later).  Just need to wrap it all up into
https://github.com/bids-standard/bids-specification/pull/1775/files#diff-340dd8fa4a5cba6b30520d676d1831fd82628971874ea46396e3ace6c6ef0f03R111 but also then also into bep032 etc ... I feel like we need a github action to reuse here... Filed

- https://github.com/bids-standard/bids-validator/issues/1931

WDYT @effigies ?

nellh:
> oh, will it work with `file://` URLs? (happen you could check quickly... if not -- I will try later).

File URLs or a path to a local file should work.

yarikoptic:
just for clarity -- https://github.com/bids-standard/bids-examples/blob/master/.github/workflows/validate_datasets.yml aren't using new validator yet?

effigies:
Correct.

effigies:
Resolved in bids-standard/legacy-validator#1932.

[...]

PR #2166, opened 2024-10-17
For deno runtime load -s value as local file if non-http

rwblair:
From #1930

### DIFF
--- a/bids-validator/src/setup/loadSchema.ts
+++ b/bids-validator/src/setup/loadSchema.ts
@@ -23,9 +23,17 @@ export async function loadSchema(version?: string): Promise<Schema> {
   ) as Schema
 
   if (schemaUrl !== undefined) {
+    let jsonData = {}
     try {
-      const jsonResponse = await fetch(schemaUrl)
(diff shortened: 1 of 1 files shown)


effigies:
Hmm. I'm actually able to load the schema with `file://$HOME/Projects/bids/specification/src/schema.json` before, but not now. I'm okay with not using `file:///`, and this works with that. Did you test with `file:///`?

effigies:
```
❯ deno
Deno 2.0.3
exit using ctrl+d, ctrl+c, or close()
REPL is running with all permissions allowed.
To specify permissions, run `deno repl` with allow flags.
> fetch('file:///home/chris/Projects/bids/specification/src/schema.json')
Promise {
  Response {
    body: ReadableStream { locked: false },
    bodyUsed: false,
    headers: Headers {},
    ok: true,
    redirected: false,
    status: 200,
    statusText: "OK",
    url: "file:///home/chris/Projects/bids/specification/src/schema.json"
  }
}
```

effigies:
@rwblair Please reopen on bids-validator.

PR #2174, opened 2024-10-25
Fix HED error reporting and tests

### LINKED ISSUE #2170: HED module incorrectly reports HED_VERSION_NOT_DEFINED warning on HED tests
The tests have improperly nested HED data in three sidecars (the `HED` level is the third level, where it should be the second level). This is causing https://github.com/hed-standard/hed-javascript/blob/916236a58b9912cb4fefc0e79aa9c28d0f35cedb/bids/types/json.js#L107 to (correctly) throw an `IssueError` for what amounts to a `SIDECAR_INVALID` HED error. However, the `try` block starting at https://github.com/bids-standard/bids-validator/blob/5646d9dee39eb9228881ec7e57887021cf3072b0/bids-validator/validators/hed.js#L34 incorrectly assumes all caught errors are *missing schema* errors, leading to improper issue reporting.

The tests need to be fixed to correct the level of `HED` in the three bad test sidecars, and `hed.js` needs to be re-coded to find another way to deal with a `null` return from the validation function.

_Originally posted by @happy5214 in https://github.com/bids-standard/bids-validator/issues/2167#issuecomment-2432545093_

happy5214:
This PR fixes the reporting of `IssueError` objects thrown from hed-validator by converting them into proper `Issue` objects. It also cleans up several errors in the HED tests that should have been failing. This should help with the eventual merging of #2167.

Fixes #2170

### DIFF
--- a/bids-validator/tests/hed.spec.js
+++ b/bids-validator/tests/hed.spec.js
@@ -27,7 +27,7 @@ describe('HED', function () {
     ]
     const jsonDictionary = {
       '/sub01/sub01_task-test_events.json': {
-        myCodes: {
+        test: {
           HED: {
             one: 'Duration/5 s',
           },
@@ -55,12 +55,8 @@ describe('HED', function () {
     ]
     const jsonDictionary = {
       '/sub01/sub01_task-test_events.json': {
-        myCodes: {
-          test: {
-            HED: {
-              one: 'Label/#',
-            },
-          },
+        test: {
+          HED: 'Label/#',
         },
       },
       '/dataset_description.json': { HEDVersion: '8.0.0' },
@@ -86,11 +82,9 @@ describe('HED', function () {
 
     const jsonDictionary = {
       '/sub01/sub01_task-test_events.json': {
-        myCodes: {
-          test: {
-            HED: {
-              one: 'ts:Sensory-presentation, Label/#',
-            },
+        test: {
+          HED: {
+            one: 'ts:Sensory-presentation, Train',
           },
         },
       },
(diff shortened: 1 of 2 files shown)


happy5214:
Let me know if the additional warnings for missing `HEDVersion` fields are a problem.

effigies:
This is now reliably timing out tests at 10s. Looking at it, I would guess that there's an efficiency problem with the use of `issues.push`. #2162 removed some uses to avoid call stack limits.

happy5214:
> This is now reliably timing out tests at 10s. Looking at it, I would guess that there's an efficiency problem with the use of `issues.push`. #2162 removed some uses to avoid call stack limits.

I didn't really add any new uses of `push` other than `catch` handlers (which shouldn't be executing that much anyway). I also eliminated a function call. I tried a `flatMap`, which didn't help. Any other ideas?

effigies:
Nope. @mguaypaq Any advice? Or @rwblair?

mguaypaq:
Would it be possible to run the tests [with profiling enabled](https://nodejs.org/en/learn/getting-started/profiling)? Even better would be having both "before this PR" and "after this PR" profiles, to see any significant changes. Without something like this, it's really hard to diagnose performance issues.

effigies:
It does not look trivial, but we can look into it. If there's an example somewhere of someone setting up profiling for jest tests, that would be a helpful model.

happy5214:
The timeouts are because the tests are not structured properly. Any time a testcase assertion fails, the `isdone` callback is never called, which leads to the timeout (as Jest never figures out the block is finished). Just fix the assertions in `bids.spec.js` and you're good to go.

effigies:
Okay, so it looks like datasets without HED in are now expected to raise HED warnings?

happy5214:
I rewrote the branch so that it no longer raises warnings for missing `HEDVersion` fields except when there's HED data. In doing so, I also (accidentally) clobbered the `isdone` changes (which really should be implemented instead by making the anonymous function an `async`).

effigies:
Thanks!

PR #2184, opened 2024-10-30
fix: Use our fork of nifti-reader-js to fix console output bugs

nellh:
Fork is available at https://github.com/bids-standard/NIFTI-Reader-JS and I released 0.6.9 to include the console.log fixes that haven't landed in upstream.


effigies (approved):
LGTM.

